Issue #336, opened 2021-07-21
Rest API: Most important API for digital shadow, sends asset specific data


larswitte:
This issue was split out of the #313 issues and is for the digital shadow issue in general #297.

**Functionality**
This function returns all uniqueProducts for that specific asset in the given timerange. One Datapoint contains one childUID, all parentUID's and all available alternativeUniqueProductID's. All unqiueProductTags and uniqueProductTagStrings (value and timestamp for each uniqueProduct) for the childUID are returned in the same Datapoint.

`get /{customer}/{location}/{asset}/uniqueProductsWithTags`
from `<timestamp1>` to `<timestamp2>` (in RFC 3999 Format)

**Open Questions**
- What happens if there is no productTag for that uniqueProduct? (probably null). 
- What happens if there are multiple values for one uniqueProductTag (error?)

**Example**
(Two data points. The data for one single part at one asset is combined together to one data point.)
```.js
{
	"columnNames": [
		"ValveHeadAlternativeUniqueID",
		"TorqueScrew1",
		"TorqueScrew2",
		"TorqueScrew3",
		"TorqueScrew4",
		"gasketUID",
		"childUID"
		"Timestamp_ValveHeadID"
		"Timestamp_GasketImageID"
	]
	"datapoints": [
		[
			"124000003941",
			5.2,
			5.4,
			4.9,
			5.4,
			"345g48t34ru2,
			"2502934850239587"
			1239481702
			2314092348
		]
			"124000003942",
			5.1,
			5.1,
			4.8,
			5.9,
			"2493trf8434r",
			"32ß4529ß8359"
			13204982743
			12359873404
		]
	]
}
```

Gornoka:
Format:
``` json
[
   row

]

row:
{
"ValveHeadAlternativeUniqueID": str<123>
"TorqueScrew1": number<5.0>
"TorqueScrew2": number<5.0>
"TorqueScrew3": number<5.0>
"TorqueScrew4": number<5.0>
"gasketUID": number<5.0>
"childUID": number<5.0>
"Timestamp_ValveHeadID": number<5.0>
"Timestamp_GasketImageID": number<5.0>
}
	
´´´

example
´´´json
[
{
"ValveHeadAlternativeUniqueID": str<123>
"TorqueScrew1": number<5.0>
"TorqueScrew2": number<5.0>
"TorqueScrew3": number<5.0>
"TorqueScrew4": number<5.0>
[...]

Issue #313, opened 2021-07-14
Factoryinsight: Add REST API's for productTag data from TimescaleDB


larswitte:
**Description**
To access the productTagTable Data (used for digital shadow, track and trace: #297) we need to provide new REST API's.

**New REST API's**
- get /{customer}/{location}/{asset}/productTags
from ```<timestamp1>``` to ```<timestamp2>``` (in RFC 3999 Format)
Example: (Two data points. The data for one single part at one asset is combined together to one data point.)
```
{
	"columnNames": [
		"ValveHeadID",
		"TorqueScrew1",
		"TorqueScrew2",
		"TorqueScrew3",
		"TorqueScrew4",
		"GasketImageID",
		"UID"
		"Timestamp_ValveHeadID"
		"Timestamp_GasketImageID"
	]
	"datapoints": [
		[
			"124000003941",
			5.2,
			5.4,
			4.9,
			5.4,
			"345g48t34ru2,
			"2502934850239587"
			1239481702
			2314092348
		]
			"124000003942",
			5.1,
			5.1,
			4.8,
			5.9,
			"2493trf8434r",
			"32ß4529ß8359"
			13204982743
			12359873404
		]
	]
}
```

- get /{customer}/{location}/digitalProductShadow
from ```<timestamp>``` to ```<timestamp>``` (in RFC 3999 Format)
for specific ```<ValveHeadID>``` or ```<HeatExchangerID>``` or ```<CompressorID>``` or ```<EBoxID>```.
Provide the constant values (ValveHeadID etc.) without and asset specific values with AssetID prefix to have identifiable values.
Example (with one datapoint):
```
{
	"columnNames": [
		"ValveHeadID",
		"HeatExchangerID",
		"EBoxID"
		"CompressorID",
		"ProductID"
		"Washing_Timestamp",
		"Preassembly_Timestamp",
		"Preassembly_ScrewBatch",
		"Assembly_1_TorqueScrew1",
		"Assembly_1_TorqueScrew2",
		"Assembly_1_TorqueScrew3",
		"Assembly_1_TorqueScrew4",
		"Assembly_1_GasketImageID",
		"Assembly_1_Timestamp_Scan",
		"Assembly_1_Timestamp_Image",
		"Assembly_2_Timestamp"
		"Assembly_3_Timestamp"
	]
	"datapoints": [
		[
			"124000003941",
			"848000001248",
			"256000002348",
			"167000002348",
			"TestProduct123"
[...]

PR #359, opened 2021-07-22
313 add productshadow rest api

JeremyTheocharis:
Implements the REST API specified in #336 

Originally started from #313 , but due to change in scope I only worked on #336


Gornoka (changes requested):
recommending the validation against a set of defined generators since they are usually "harder" to pass then the validator.
changing inline datatypes to schemas as noted in line 115 of factroyinsights

Gornoka:
@JeremyTheocharis  please see proposed changes at #362

Gornoka (approved):

PR #366, opened 2021-07-23
365 assign docker container to camera

### LINKED ISSUE #365: Assign camera to docker via serial number
So far, the cameraconnect module will randomly pick a camera which is available in the network. If more than one camera is available the user has no option to specify which one should be used. 

Solution: The user should be able to specify the camera via its serial number within the env file of the container

JonaVo:
Fixes #365 
Multi camera support now available
User specifies camera via mac address, which is written on the back of each camera

Also adjusted the harvester version used

### DIFF
--- a/cameraconnect/src/cameras.py
+++ b/cameraconnect/src/cameras.py
@@ -50,7 +50,7 @@ class CamGeneral(ABC):
         See inheritors
     """
 
-    def __init__(self, mqtt_host, mqtt_port, mqtt_topic, serial_number, image_storage_path=None) -> None:
+    def __init__(self, mqtt_host, mqtt_port, mqtt_topic, serial_number, mac_address, image_storage_path=None) -> None:
         """
         Base class constructor configures the object with the 
         MQTT host settings.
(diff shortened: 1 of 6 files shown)


JonaVo:
Please check again, except the 16 parameters for the init function, smelly code has been removed

cameraconnect/src/cameras.py
```diff
@@ -137,7 +140,7 @@ def _publish_mqtt(self, image) -> None:
         message = json.dumps(prepared_message)
         # Publish the message
         ret = self.client.publish(self.mqtt_topic, message, qos=0)
-        print(ret)
+        print("Image No.: " + str(ret[1]))
```
JeremyTheocharis, line 143: Please use logging module for each use of print

JeremyTheocharis (changes requested):

cameraconnect/src/cameras.py
JeremyTheocharis: @JonaVo Please take a look at the logging (using the logging instead of print)

JonaVo: done 1bc5eed

Issue #396, opened 2021-07-29
timescaleDB: implement digital shadow tables + add changes


larswitte:
****
Four tables are especially relevant:
- uniqueProductTable contains entries with a pair of one UID and one AID and other data.
- productTagTable and productTagStringTable store information referenced to the UID's in the uniqueProductTable. Stored is everything from individual measurements to quality classes.
- productInheritanceTable contains pairs of child and parent UID's. The table as a whole thereby contains the complete inheritance information of each individual part. One entry describes one edge of the inheritance graph.

The new relevant tables are dotted, the uniqueProductTable changes are bold in the timescaleDB structure visualization.
![timescaleDB_image](https://user-images.githubusercontent.com/87082063/127489013-1072c945-1ea3-4ca0-8919-c57ebb01ba2b.png)

Issue #297, opened 2021-07-08
Digital Shadow - Track and Trace


larswitte:
*This is an overview site about the digital shadow - track and trace feature development. Please link relevant issues to this main issue.*

**Goal:**
In order to gain detailed insight in the production process and into the produced products we need a system of features to acquire, and access information gained by scanners, sensors etc. This would enable better quality assurance and production improvement.

**Possible Solution:**
We could send small mqtt messages containing a timestamp with a single value like a scanned ID, a measured value etc. and contextualize them with microservices. The gained data would be pushed into a Database by mqtt-to-postgresql. After that factoryinsight might provide a interface for BI-Tools like Tableau.

larswitte:
**Overview: current timescaleDB concept (new tables with dotted line)**
![postgreSQL_DatabaseVis_02](https://user-images.githubusercontent.com/87082063/125815493-4526ee03-a10d-4cca-9a12-35fdace05baf.png)

larswitte:
Possible **general way to solve the contextualization** with microprocesses before mqtt-to-postgres (processValues + child AID -> productTag, combination child AID with parent AID's, /uniqueProducts for new UID's with connected AID)

1. Make predicted empty containers for necessary Messages to mqtt-to-postgres when new "step beginning trigger<2 (first production step) is recieved
2. fill cointainers if relevant messages come in
3. if full send
4. if trigger before container full send container and set missing filds to null

larswitte:
**Updated version of digital shadow concept:**
![DigitalShadow_18](https://user-images.githubusercontent.com/87082063/126606415-1cb8be4f-bfa8-40d6-8f07-0331eabc03c2.png)

larswitte:
**Updated version 22 of digital shadow concept:**
![DigitalShadow_22](https://user-images.githubusercontent.com/87082063/126634030-efc802a9-b64e-434e-b3b5-ed05f606e3b0.png)

larswitte:
The database model refactoring for digital shadow is at the issue #367. Thank you @Gornoka!

PR #399, opened 2021-07-29
timescaleDB: implement digital shadow tables + changes

larswitte:
Fixes issue #396 
*Not ready yet!*

General issue: digital shadow: #297.

**Open questions:**
- `#??` -> what number should be directly over the `ALTER TABLE` statements?
- `uid` in `uniqueProductTable` is `TEXT` type. Is that correct (not `SERIAL`)?
- For the productTag tables I only used the `SELECT create_hypertable` function. Should I also index them?

timescaleDB_image


larswitte:
Merged into #440 to enable testing.

PR #413, opened 2021-07-30
Fixed #410

### LINKED ISSUE #410: mqtt-to-blob needs to adhere to the MQTT message standard
Also related to #409 

Currently, it needs a image_uid. However, it should be called image_id

JeremyTheocharis:
Additionally, added safe JSON parsing


JonaVo (approved):
looks correct

PR #415, opened 2021-07-30
adjusted error message to display correct error description

### LINKED ISSUE #406: correct error displayed by cameraconnect
The error show claims that it is caused by wrong network settings. This has been the cause when camera connect was solely used with docker-compose. However in Kubernetes this error is thrown if two containers occupy the same camera. This occurs e.g. if one container is manually restartet. Since the old container needs time to terminate the camera will not be immidiately available for the new container.
The given error description needs to be adjusted.

image

JonaVo:
Fixed #406

adjusted error description since it was outdated

This was caused by the switch from docker-compose to kubernetes

### DIFF
--- a/cameraconnect/src/cameras.py
+++ b/cameraconnect/src/cameras.py
@@ -412,7 +412,7 @@ class GenICam(CamGeneral):
                 try:
                     self.ia = self.h.create_image_acquirer(id_=camera.id_)
                 except:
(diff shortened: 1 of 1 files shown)


JonaVo:
Fixed. Pls review

JeremyTheocharis (approved):

PR #436, opened 2021-08-05
Fixes #435

### LINKED ISSUE #435: factorycube-server is pulling incorrect image
factorycube-server is pulling testing for mqtt-to-blob & grafana-proxy instead of latest.

## Expected Behavior

This should be latest

## Current Behavior

Pulls testing

## Possible Solution

Replace ```tag: testing``` with ```tag: latest```

## Detailed Description

See:
https://github.com/united-manufacturing-hub/united-manufacturing-hub/blob/main/deployment/factorycube-server/values.yaml#L54
https://github.com/united-manufacturing-hub/united-manufacturing-hub/blob/main/deployment/factorycube-server/values.yaml#L66

Scarjit:
Resolves #435

### DIFF
--- a/deployment/factorycube-server/values.yaml
+++ b/deployment/factorycube-server/values.yaml
@@ -51,7 +51,6 @@ mqtttoblob:
   enabled: true
   image:
     repository: unitedmanufacturinghub/mqtt-to-blob
-    tag: latest
     pullPolicy: Always
   replicas: 1
   pdb:
@@ -63,7 +62,6 @@ grafanaproxy:
   enabled: true  
   image:
     repository: unitedmanufacturinghub/grafana-proxy
-    tag: latest
     pullPolicy: Always
   replicas: 1
   service:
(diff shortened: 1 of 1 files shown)


deployment/factorycube-server/values.yaml
```diff
@@ -51,7 +51,7 @@ mqtttoblob:
   enabled: true
   image:
     repository: unitedmanufacturinghub/mqtt-to-blob
-    tag: testing
```
JeremyTheocharis, line 54: Let's just remove the tags entirely as then we have the default tag handling kicking in 

See also:  https://github.com/united-manufacturing-hub/united-manufacturing-hub/blob/5177804744f0ff0294b822e85b4967655c2bc9fb/deployment/factorycube-server/templates/mqtttoblob-deployment.yaml#L36-L40 

The default behavior means that it will use the latest release tag (currently v0.5.0). With this we prevent the issue with latest, which we had e.g. with VerneMQ.

JeremyTheocharis (changes requested):

JeremyTheocharis (approved):

PR #464, opened 2021-08-24
fixed type conversion product_id in postgres

### LINKED ISSUE #463: postgres conversion of string to int in uniqueProductTable.product_id doesn't work
**Problem:**
Following conversion doesn't work.
```sql
  ALTER TABLE uniqueProductTable ALTER COLUMN product_id SET DATA TYPE INTEGER;
```

**Solution:**
Specify explicit cast to convert the product_id.
```sql
  ALTER TABLE uniqueProductTable ALTER COLUMN product_id SET DATA TYPE INTEGER USING (product_id::integer);
```

larswitte:
Fixes #463


larswitte:
After working on the digital shadow documentation I have one more questions:
Should we make `value` in `productTagTable` and `productTagStringTable` NOT NULL instead of NULL? A producttag with a null value is in the end not very useful anyways.

JeremyTheocharis (approved):

Issue #500, opened 2021-09-02
Problem with queue: locking


larswitte:
**Problem Description**
The specific problem occured after a product wasn't addet to the productTable. Follwing uniqueProduct MQTT messages couldn't be processed. Because of that also addParentToChild and productTag messages were not processed.

Deleting the mqtt-to-postgres pod and adding the missing product to the productTable didn't solve the issue: still no new uniqueProducts, productTags and addParentToChild messages were processed.

**Suspected underlying issues:**
1. If one queue item has an issue writing into sql, everything is put back into the queue and retried from the beginning (even queue items that are already in the timescaleDB)
2. After one issue, mqtt-to-postgres doesn't try to write the following queue items into timescaleDB but breaks. Because of that, certain queue items (previous items might depend on) are never even tried to write in queue.
3. After a while there are no error messages any more witch might be, because `stmt.Close()` isn't executed in case of errors.

*Example*
Message 1: -> wrote into DB
Message 2: depends on Message 3 -> issue, process stopped
Message 3: -> never tried to write in db

[repeat]

**Possible solution**
1. Try to write every message into db, only put not working messages back into queue
2. Close with `stmt.Close()`, even in case of errors
3. to prevent the queue from gradually filling with problematic messages, add timeout for queue items (maybe long term store somewere else)

*Example*
Message 1: -> wrote into DB
Message 2: depends on Message 3 -> issue, returned and put back into queue the next time
Message 3: -> wrote into DB

Message 2: retried -> wrote into DB (because now Message 3 is already in DB)

Scarjit:
Fixed in #506

PR #504, opened 2021-09-02
503 get unique product id latest UID

### LINKED ISSUE #503: DigitalShadow: GetUniqueProductID gets wrong ID
*Related to Digital Shadow: #297*

**Context:**
One item (with one AID) can be recorded multiple times at one asset, meaning that the combination of assetID and AID isn't unique. This is correctly represented by the timescaleDB ddl. If a product is passing multiple times at one asset (maybe because of rework) the GetUniqueProductID function should always return the latest UID.

**Problem:**
The current func `GetUniqueProductID` always returns the oldest and not the youngest UID. This is because the SQL query doesn't accout for the possibility of non-unique assetID and AID combinations and returns multiple UID (default in ascending order).

**Solution:**
Order the UID's by timestamp and select only the latest one.

**Disclaimer**
An AID is likely to be recorded multiple times (e.g. in case of scanning twice). Users have to make shure to only send the intended uniqueProduct MQTT messages.

larswitte:
Fixes #503


golang/internal/cache.go
```diff
@@ -561,62 +561,6 @@ func StoreAssetIDToCache(customerID string, location string, assetID string, DBa
 
 	b := strconv.Itoa(DBassetID)
 
-	err := rdb.Set(ctx, key, b, dataExpiration).Err()
-	if err != nil {
-		zap.S().Errorf("redis failed")
-		return
-	}
-}
-
```
larswitte, line 570: I didn't delete this part, I just deleted the caching parts. Is it just a displaying wrongly or some underlying issue?
Please check @JeremyTheocharis.

golang/internal/cache.go
```diff
@@ -561,62 +561,6 @@ func StoreAssetIDToCache(customerID string, location string, assetID string, DBa
...
-// StoreUniqueProductIDToCache stores uniqueProductID to cache
-func StoreUniqueProductIDToCache(aid string, DBassetID int, uid int) {
-	if rdb == nil { // only the case during tests
-		//zap.S().Errorf("rdb == nil")
-		return
-	}
-
-	key := fmt.Sprintf("getUniqueProductID-%s-%d", aid, DBassetID)
```
larswitte, line 611: I also didn't edit this part...

golang/internal/cache.go
```diff
@@ -561,62 +561,6 @@ func StoreAssetIDToCache(customerID string, location string, assetID string, DBa
...
-
-	if uid == 0 {
-		// zap.S().Debugf("input is empty. aborting storing into database.", key)
-		return
-	}
-
-	b := strconv.Itoa(uid)
-
 	err := rdb.Set(ctx, key, b, dataExpiration).Err()
 	if err != nil {
```
larswitte, line 565: I deleted this part.

golang/internal/cache.go
larswitte: The red marking seems to be 6 lines to high.

larswitte:
I checked it with my node red test flow: worked!

JeremyTheocharis (commented):
@Scarjit 

See also other PR

Scarjit:
Integrated in #500 (https://github.com/Scarjit/united-manufacturing-hub/commit/4c2f9854c9ea8bf4eb1b1dc1fdf8f260547e4ae8)

PR #557, opened 2021-09-23
fix: MQTT port collisions in development setup between factorycube-se…

### LINKED ISSUE #547: factory cube edge and server can not run on the same device reliably since they both occupy the same port for mqtt
when running factory cube edge and server on the same device both mqtt brokers try to use the same port, this yields unpredictable behaviour since one broker may be faster than the other
## Expected Behavior

Tell us what should happen 
both brokers should use different porst so that factorycube edge and server can run on the same device/ network adapter
## Current Behavior

they both bind on the same port and one randomly crashes 
## Possible Solution

change one to use a different port (probably server since edge has more connections to external software)

## Steps to Reproduce
1. install factorycube edge and server using the install guide ( development config)
2. observe unpredictable behaviour

JeremyTheocharis:
…rver and factorycube edge

Fixes #547

### DIFF
--- a/deployment/factorycube-server/values.yaml
+++ b/deployment/factorycube-server/values.yaml
@@ -2247,11 +2247,13 @@ vernemq:
     labels: {}
     mqtt:
       enabled: true
-      nodePort: 1883
+      # using a different port than 1883 to prevent collisions with factorycube-edge in development setups
+      nodePort: 1884
       port: 1883
     mqtts:
       enabled: true
-      nodePort: 8883
+      # using a different port than 8883 to prevent collisions with factorycube-edge in development setups
+      nodePort: 8884
       port: 8883
     type: LoadBalancer
     ws:


Scarjit (commented):
This change should be documented, since it is a breaking change for anything, that would interact directly with factorycube-server (ex: mqtt-to-postgres)

Scarjit:
Changes are needed here:
 - https://github.com/united-manufacturing-hub/united-manufacturing-hub/blob/main/deployment/mqtt-to-blob/docker-compose.yml#L19
 - https://github.com/united-manufacturing-hub/united-manufacturing-hub/blob/main/deployment/factorycube-server/templates/mqtttopostgresql-statefulset.yaml#L80
 - https://github.com/united-manufacturing-hub/united-manufacturing-hub/blob/main/deployment/factorycube-server/templates/mqtttoblob-deployment.yaml#L50

Scarjit (approved):

PR #621, opened 2021-10-29
604 Install script should not continue if encountering an error

### LINKED ISSUE #604: Install script should not continue if encountering an error
## Expected Behavior

The initiail_setup.sh should stop or even revert, when encountering errors

## Current Behavior

The script continues, even if there are major errors

## Steps to Reproduce

1. Delete the DO_NOT_DELETE folder
2. Run initial_setup.sh

Scarjit:
Fixes #604

### DIFF
--- a/docs/static/examples/development.yaml
+++ b/docs/static/examples/development.yaml
@@ -94,41 +94,83 @@ write_files:
   path: /home/rancher/configs/factorycube-edge-development.yaml
 
 - content: |
-    #!/bin/bash
-    LOCKDIR=/home/rancher/DO_NOT_DELETE
+      #!/bin/bash
+      LOCKDIR=/home/rancher/DO_NOT_DELETE
 
-    if [ $UID -eq 0 ]; then
-        exec su "rancher" "$0" -- "$@"
-        exit
-    fi
+      if [ $UID -eq 0 ]; then
+          exec su "rancher" "$0" -- "$@"
+          exit
+      fi
 
-    if mkdir $LOCKDIR; then
-        echo "Initial setup! Waiting 20 seconds for full startup..."
+      if mkdir $LOCKDIR; then
+          echo "Initial setup! Waiting 20 seconds for full startup..."
 
-        sleep 20
-        echo "Continuing with installation. Installing helm..."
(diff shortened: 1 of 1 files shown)


JeremyTheocharis:
Idea: let's check NTP, DNS connections and check whether the current time is higher than 01-01-2021. With this we prevent a lot of typical failures during installation.

Could we additionally highlight "our" messages and include a short "next steps can be found in our documentation"?

docs/static/examples/development.yaml
```diff
@@ -94,41 +94,165 @@ write_files:
...
+      reachable_ntp_server_found=1
+      fi
+      if [ $reachable_ntp_server_found -eq 1 ]; then
+      break
+      fi
+      i=$[$i+1]
+      done
+
+      if [ $reachable_ntp_server_found -eq 0 ]; then
+      echo "No reachable NTP server found, aborting setup"
```
JeremyTheocharis, line 152: Please add a note for someone who does not know what a NTP server is, e.g., "No reachable NTP server found, aborting setup. This can happen when you either have no connection to the internet or when you are in a limited network that blocks NTP requests (some universities do that). The solution is to connect ot the internet and / or to specify the allowed NTP servers manually in the development.yaml. Fore more information visit docs.umh.app"

docs/static/examples/development.yaml
```diff
@@ -94,41 +94,165 @@ write_files:
...
+      else
+      echo "DNS lookup successfull"
+      dns_query_successfull=1
+      break
+      fi
+      i=$[$i+1]
+      done
+
+      if [ $dns_query_successfull -eq 0 ]; then
+      echo "Failed to resolve umh.app, aborting setup"
```
JeremyTheocharis, line 174: See above.

docs/static/examples/development.yaml
JeremyTheocharis: also the script should not abort when the time is configured properly (e.g., in BIOS). For this you can place the additional time check as a "else if" to this "if" clause

docs/static/examples/development.yaml
```diff
@@ -94,41 +94,165 @@ write_files:
...
+      exit
+      fi
+
+      if [ $(date +%s) -lt 1609455600 ]; then
+      echo "Current Date is < 01.01.2021, aborting setup"
```
JeremyTheocharis, line 179: See above.

JeremyTheocharis (changes requested):

JeremyTheocharis (approved):

PR #623, opened 2021-11-02
fix: misconfigured node red takes down the entire stack

### LINKED ISSUE #614: Misconfigured Node-RED takes down the entire device
Misconfigured Node-RED takes down the entire device

## Expected Behavior

Node-RED should crash when misconfigured, stuck in a recursion loop, whatever

## Current Behavior

Node-RED just takes all CPU and memory and continues with its work. This can take down the entire device.

## Possible Solution

Add resource limits to all components of factorycube-edge

## Steps to Reproduce

1. Let Node-RED send an image 100 times per second
2. Profit

JeremyTheocharis:
Fixes #614


Scarjit (approved):

JeremyTheocharis:
@Scarjit You did not squash merge this one :/ we should avoid having commits like "TEMP" in the commit history :)

Issue #660, opened 2021-11-11
Long time between MQTT message and readable message in dashboard (>30 seconds)


kruegeralex:
For an impressive impression, the time between the arrival of a message and its display in the dashboard should be <<5 sec

Please contact @kruegeralex  to get access to the test environment.

JeremyTheocharis:
Could you describe further what is happening? Used versions? CPU load? (Previous release had high CPU usage, probably system was not upgraded yet)

kruegeralex:
<img width="770" alt="image" src="https://user-images.githubusercontent.com/77407527/142256368-97cc6b93-8728-4753-9d05-c0393045a9b6.png">

Helm chart version 0.6.0 with mqtt-to-postgres upgraded to 0.7.0 @Scarjit correct me if i am wrong

Scarjit:
I assume the messages get stuck due to a large amount of invalid messages.
The solution will be to tweak the sleep times, but this will also increase CPU load, due to the increased amount of times it tries & failes to insert incorrect data.

Scarjit:
See #671

PR #667, opened 2021-11-18
fix: This aims to fix the long wait time observed in #660 by reducing the queue timeout on error.

Scarjit:
Reduces the impact of #660 
#660 should remain open until confirmation of the fix in production


JeremyTheocharis (approved):
See also #671

PR #682, opened 2021-11-23
feat: remove landingpage from documentation

### LINKED ISSUE #680: Remove landingpage from documentation
When you currently go to docs.umh.app, you are greeted by a "start page", which currently adds no value, so I suggest removing it.

Scarjit:
Closes #680

### DIFF
--- a/docs/config.toml
+++ b/docs/config.toml
@@ -10,7 +10,7 @@ theme = ["docsy"]
 enableGitInfo = true
 
 # Language settings
-contentDir = "content/en/"
+contentDir = "content/en"
 defaultContentLanguage = "en"
 defaultContentLanguageInSubdir = false
 # Useful when translating.
--- a/docs/config.toml
+++ b/docs/config.toml
@@ -10,7 +10,7 @@ theme = ["docsy"]
 enableGitInfo = true
 
 # Language settings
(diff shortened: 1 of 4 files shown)


kruegeralex:
Hi @Scarjit for the quick implementation! When I click on the preview link, only this empty page appears (see attachment). If I then click on "Documentation" (top right), the normal documentation appears.
image

Scarjit:
@kruegeralex Thats why the pull request is still a draft ;)

Scarjit:
After trying out multiple methods, that the internet says would redirect the page, i think it would be easier to make the redirect at webserver level.
This would also reduce the problems with old browsers who don't understand the refresh meta tag.

@JeremyTheocharis Can you make an 301 redirect from https://docs.umh.app to https://docs.umh.app/docs in our webserver ?

JeremyTheocharis:
> After trying out multiple methods, that the internet says would redirect the page, i think it would be easier to make the redirect at webserver level. This would also reduce the problems with old browsers who don't understand the refresh meta tag.
> 
> @JeremyTheocharis Can you make an 301 redirect from https://docs.umh.app to https://docs.umh.app/docs in our webserver ?

Done :)

PR #704, opened 2021-11-24
fix: manually overwrite ALLOW_PLAINTEXT_LISTENER

Scarjit:
Hopefully finally fixes the kafka startup problem


JeremyTheocharis (changes requested):
specify yes or no instead in the Helm chart. See definition here: https://github.com/bitnami/charts/search?q=allowPlaintextListener

JeremyTheocharis (approved):

PR #759, opened 2021-12-13
fix: moved pull request template in accordance with https://stackoverflow.com/a/62613198

Scarjit:
According to [stackoverflow](https://stackoverflow.com/a/62613198) the pull request template has to be in the root folder of .github.


JeremyTheocharis (approved):
Let's hope it works then :)

Scarjit:
Seems to be stuck at dockercloud *again*

PR #988, opened 2022-04-21
Added microservices to calculate state from activity

### LINKED ISSUE #987: Add microservice calculating state from activity and detectedAnomaly
**Is your feature request related to a problem? Please describe.**
We need to calculate the machine state from activity and detectedAnomaly

**Describe the solution you'd like**
So we need a microservice enabled by default subscribing to all machines, watching for activity and detectedAnomaly and then applying the logic.

Logic (proposal, suggestions welcome):

Foreach activity message: if activity is false, then set state to UnspecifiedStopState. If activity is true, then set state to ProducingAtFullSpeedState (use timestamp of incoming activity message as state timestamp)

Foreach detectedAnomaly: 
1. get current activity by fetching last x (e.g., 100) activity messages from kafka, ordering them by timestamp (not timestamp of kafka, but of the message) and then jus taking the last one (reason: it could happen that the messages are unordered). 
2. If activity is true, then ignore message. 
3. If activity is false, then set state to detectedAnomaly (assuming detectedAnomaly is a string containing a valid state number in here: https://docs.umh.app/docs/concepts/state/, need to append documentation to suggest using these state numbers instead of a free-written string)

it should be enabled by default in the helm chart and should be able to disabled in 000_commonConfig

**Describe alternatives you've considered**
Doing this logic in Node-RED and MQTT is impossible as it requires persisting state (where Node-RED is not good in). 

**Additional context**
The logic above is a proposal and can be modified for a more reliable / performant approach. As a second step you could store the activity state also in-memory and only fetch it from kafka during startup. For performance: it should be able to handle 10 incoming activity and detectedAnomaly messages per second.

Scarjit:
# Description

This pr adds a microservices, calculating state from activity and detectAnomaly

Fixes #987 

## Type of change

- [X] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [ ] Send activity messages and check if state messages are send
- [ ] Send activity messages and check if detectAnomaly messages are send

# Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I have checked that my issue complies with the [Contributing guidelines](https://github.com/united-manufacturing-hub/united-manufacturing-hub/blob/main/CONTRIBUTING.md)


Scarjit:
TODO: Add Helm chart related files and GH workflow

golang/cmd/kafka-state-detector/activityKafka.go
```diff
@@ -0,0 +1,123 @@
...
+		if err != nil {
+			zap.S().Errorf("[AC] Failed to parse activity message: %s", err)
+			continue
+		}
+
+		var stateMessage datamodel.State
+		if activityMessage.Activity {
+			stateMessage.State = datamodel.ProducingAtFullSpeedState
+		} else {
+			stateMessage.State = datamodel.UnspecifiedStopState
```
JeremyTheocharis, line 97: ensure that this is only applied when activity changes and not when someone sends repeatedly the same activity=false message.

Reason: we do not want to overwrite a "detectedAnomaly state" with a "duplicated activity=false and therefore state UnspecifiedStopState message"

JeremyTheocharis: deduplication with checking timestamp and activity from latest activity (you already have that function somewhere)

JeremyTheocharis (changes requested):

golang/cmd/kafka-state-detector/activityKafka.go
JeremyTheocharis: did you worked on this? i am not seeing any commits

JeremyTheocharis (approved):

JeremyTheocharis:
@Scarjit: best is if you merge this PR, i am too unsure about messageprocessing.go (whether to keep it or remove it)

Scarjit:
This should be merged after #1014

PR #1014, opened 2022-04-27
Refactor kafka-bridge/kafka-to-postgres

Scarjit:
# Description

This pr reduces duplicate code in kafka-bridge and kafka-to-postgres.
It also moves putback messages to header from key, for better partition handling

## Type of change


golang/cmd/kafka-to-postgresql/main.go
```diff
@@ -353,10 +316,10 @@ func PerformanceReport() {
 			"\n[HT] ProcessValueString queue lenght: %d"+
```
JeremyTheocharis, line 316: Maybe use a different delimiter, `\n` will be shown and will not do a new line. Try something like |

JeremyTheocharis (approved):

PR #1112, opened 2022-05-18
fix: incorrect regex match groups

Scarjit:
# Description

Fixed an issue in J14, where the regex match group was incorrect

## Type of change

# How Has This Been Tested?

- [X] Local tests work


Scarjit:
Superseeded by #1113

PR #1151, opened 2022-06-03
Additional protection against deep and long topic names

### LINKED ISSUE #1150: Bug Report: mqttkafkabridge crashing when using `.` in topic
### Expected Behavior

mqttkafkabridge should not send messages to Kafka that have a `.` in the topic. It should throw an error instead.

### Current Behavior

Currently message land up in Kafka and then mqttkafkabridge gets thrown out of VerneMQ because on the way back Kafka --> MQTT all the `.` are switched to a `/` character (which might result in a `topic_max_depth_exceeded`)

### Relevant log output

_No response_

### Steps to Reproduce

Use an IP-address with a lot of `.` in a MQTT topic

### Context (Environment)

-

### Possible Solution

_No response_

### Version

v0.9.x

### Anything else?

_No response_

Scarjit:
# Description

This pr adds additional protection against deep topic names (>= 10 layers (vernemq limitation)) and topic names >= 200 chars (kafka limitation).
The depth issue could only occur on raw before, since all topic are already matched against an regex.

Fixes #1150 

## Type of change

# How Has This Been Tested?

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have checked that my issue complies with the [Contributing guidelines](https://github.com/united-manufacturing-hub/united-manufacturing-hub/blob/main/CONTRIBUTING.md)


JeremyTheocharis (changes requested):
Add check that no topics containing a `.` character from MQTT to Kafka are send

JeremyTheocharis (approved):

Issue #1223, opened 2022-08-18
A new processValues takes minutes to arrive in Grafana, subsequent messages are fine


JeremyTheocharis:
### Expected Behavior

When sending new processValues, I want them to appear immediately in Grafana

### Current Behavior

It takes a couple of minutes until they appear in Grafana

### Steps to Reproduce

1. Create a new processValue and send a message
2. It will take a couple of minutes until it gets selectable in Grafana

### Context (Environment)

-

### Possible Solution

This is likely happening beause the topic creation process in Kafka takes too much time. Maybe we can speed that step up? If not possible, we should add a general warning in the documentation stating that the first message to any MQTT topic might take a couple of minutes to go initially though (but all subsequent messages will be sub-second). @Sphingobium 

### Version

v0.9.x

Scarjit:
This should be fixable by lowering [metadata.max.age.ms](https://docs.confluent.io/platform/current/installation/configuration/consumer-configs.html#consumerconfigs_metadata.max.age.ms)

Scarjit:
@JeremyTheocharis Can you give me a good time, for metadata.max.age.ms ?
I don't want to go to low.

Azure uses [3 minutes](https://docs.microsoft.com/en-us/azure/event-hubs/apache-kafka-configurations), Confluent uses 5 minutes by default.

Going to low, could block other kafka operations and will increase load on the broker.

Scarjit:
Let's use 3m

PR #1257, opened 2022-09-16
feat: add push based subscription for new topics

### LINKED ISSUE #1253: Replace metadata pull by push based option
**Is your feature request related to a problem? Please describe.**
Currently kafka consumers will pull every X minutes, making it non-live

**Describe the solution you'd like**
Instead of reducing the metadata.max.age.ms further, we could have a topic, where new metadata update are published (for now just new topics)
Every consumer could consume that one, and the do a force refresh, once a new topic is created and broadcasted.
This would make it nearly live

**Additional context**
Make sure to include some small random pause, before force requesting the metadata update, preventing a DOS when running lots of kafka services.
This should also be in some common code part, so it's easy to integrate.

DaruZero:
# Description

Adds a specific topic to which all consumer services subscribe. Messages in this topic contain the names of any new topics created by publishers, and consumers will automatically add the new topics to their subscription list.
In this way we can reduce the time needed to add a new topic to a couple of seconds, instead of relying on the `metadata.max.age.ms` parameter (#1223)

Fixes #1253 

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Full deployment of the stack on my local minikube cluster

- [x] Create new topics with MQTTExplorer. The new topics are received by `kafka-to-postgresql` in the matter of seconds
- [ ] Injecting new topics via NodeRed does not work, but that is probably NodeRed's problem. In a production environment all the new topics will eventually pass through MQTT.

# Checklist:

- [x] I have made corresponding changes to the changelog (if needed).
- [ ] I have made corresponding notices to the upgrade instructions (if needed)

# Changelog changes

Added push-based detection of new topics

# Upgrade instructions

*not needed*

### DIFF
--- a/golang/internal/kafkaMessageProcessing.go
+++ b/golang/internal/kafkaMessageProcessing.go
@@ -6,6 +6,7 @@ package internal
 import (
 	"bytes"
 	"encoding/gob"
+	"encoding/json"
 	"github.com/confluentinc/confluent-kafka-go/kafka"
 	"github.com/coocood/freecache"
 	"go.uber.org/zap"
@@ -19,6 +20,10 @@ type ParsedMessage struct {
 	CustomerId  string
 	PayloadType string
 	Payload     []byte
+}
+
+type TopicProbeMessage struct {
+	Topic string
 }
 
 // ParseMessage parses a kafka message and returns a ParsedMessage struct or false if the message is not a valid message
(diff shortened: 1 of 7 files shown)


golang/internal/kafkaMessageProcessing.go
```diff
@@ -110,3 +115,27 @@ func GetCacheParsedMessage(msg *kafka.Message) (valid bool, found bool, message
...
+func StartTopicProbeQueueProcessor(topicProbeProcessorChannel chan *kafka.Message) {
+	zap.S().Debugf("[TP] Starting queue processor")
+	for !ShuttingDownKafka {
+		msg := <-topicProbeProcessorChannel
+		if msg == nil {
+			continue
+		}
+
+		var topicProbeMessage TopicProbeMessage
+		err := json.Unmarshal(msg.Value, &topicProbeMessage)
```
Scarjit, line 130: Please use jsoniter here

Scarjit (approved):

PR #1373, opened 2022-12-06
Feat/softing

Sphingobium:
# Description

do not merge  

adds softing to the opcua sim branch


deployment/united-manufacturing-hub/templates/softing/deployment.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: apps/v1
```
JeremyTheocharis, line 1: do not put this into the helm chart. the goal is to have a tutorial containing a kubernetes deployment file like this and to add it then after UMH installation of the stack.

For our customer, I can do this in the setup scripts

JeremyTheocharis (changes requested):

PR #1409, opened 2023-01-06
Fix/1404/factoryinsight returns thousands of duplicate values

### LINKED ISSUE #1404: factoryinsight returns thousands of duplicate values
### Expected Behavior

factoryinsight should only send values, if the value came from the DB or was produced by LOCF/Interpolate

### Current Behavior

factoryinsight returns thousand of datapoints, even if they contain the same data, and have no db backing.

### Relevant log output

_No response_

### Steps to Reproduce

1) Install latest step
2) Use v2 datasource to query any datapoint

### Context (Environment)

Customer

### Possible Solution

_No response_

### Version

v0.9.x

### Anything else?

_No response_

Scarjit:
# Description

This pr checks if a datapoint only contains null values, and if so, skips it.

Fixes #1404 

## Type of change

Please delete options that are not relevant.

### DIFF
--- a/golang/cmd/factoryinsight/v2/services/service-tree.go
+++ b/golang/cmd/factoryinsight/v2/services/service-tree.go
@@ -288,7 +288,7 @@ func GetTagsTreeStructure(customer string, site string, area string, line string
 		return nil, err
 	}
 	for _, tagGroup := range tags {
-		var structure = make(map[string]*models.TreeEntryFormat)
+		var structure map[string]*models.TreeEntryFormat
 		if tagGroup == models.StandardTagGroup {
 			structure, err = GetStandardTagsTree(customer, site, area, line, cell, tagGroup)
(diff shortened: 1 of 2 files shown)


golang/cmd/factoryinsight/v2/services/service-tree.go
```diff
@@ -345,10 +345,10 @@ func GetCustomTagsTree(
 		// ignore underscores at the beginning or end of the tag or if there are multiple underscores in a row
 		sanitizedSplitTag := removeEmpty(splitTag)
 		// remove unused capacity to save memory
-		sanitizedSplitTag = slices.Clip(sanitizedSplitTag)
+		sanitizedSplitTag = slices.Clip(sanitizedSplitTag) // WTF is this?
```
DaruZero, line 348: Removes unused capacity from the slice, to save a little bit of memory. I thought it wouldn't hurt

PR #1494, opened 2023-03-22
Added cache size overwrite and reduced memory usage for mqtt-kafka-bridge

Scarjit:
# Description

This pr allows the use of the old LRU keys to set the ARC size.
It is also now allocating less memory per default.


DaruZero:
Update the documentation and link the PR here

JeremyTheocharis (approved):

PR #1523, opened 2023-03-30
Metrics generator

Scarjit:
# Description

This pr adds an microservice, collecting non-personal information.

It executes on cluster start and once every 24 hours (at midnight)


deployment/united-manufacturing-hub/templates/metrics/deployment.yaml
```diff
@@ -0,0 +1,44 @@
+# Copyright 2023 UMH Systems GmbH
```
JeremyTheocharis, line 1: deployment AND cronjob?

deployment/united-manufacturing-hub/values.yaml
```diff
@@ -1307,3 +1311,11 @@ mqtt_broker:
         - TLS_RSA_WITH_AES_128_GCM_SHA256
         - TLS_RSA_WITH_AES_128_CBC_SHA
         - TLS_RSA_WITH_AES_256_CBC_SHA
+
+metrics:
+  image:
+    repository: ghcr.io/united-manufacturing-hub/metrics
+    pullPolicy: IfNotPresent
+    tag: latest
```
JeremyTheocharis, line 1319: latest is even fine here I think because it is so unimportant

JeremyTheocharis (changes requested):

DaruZero:
consider merging onto `0.9.12-prerelease1` instead of `main`?

PR #1537, opened 2023-04-13
Replace kafka with redpanda

DaruZero:
# Description

This PR substitutes Apache Kafka with Redpanda, granting backwards compatibility by adding a service with the old name.

Fixes #1436

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] Deployment
- [x] Upgrade
- [x] Data flow

# Checklist:

- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding changes to the changelog (if needed).

# Changelog changes

Enter here your text for the changelog (or fill it out directly in Notion if you are internal)

PR #1547, opened 2023-04-19
Replace kafka with redpanda

DaruZero:
Continuation of #1537


test/test-values-tiny.yaml
```diff
@@ -18,20 +18,37 @@ _000_commonConfig:
...
+redpanda:
+  resources:
+    cpu:
+      cores: 1
+    memory:
+      enable_memory_locking: true
+      container:
+        max: 3Gi
+  console:
+    enabled: false
```
Scarjit, line 54: Can't we enable it here, and remove our dependency on standalone console ?

DaruZero: It's already like that in the default values. The `test/test-values-tiny.yaml` values are just for the CI, to test that everything works while having the minimum services possible. Since no one would access the console from the GH runner, there is no point to enable it. 
In the `test/test-values-full.yaml` values is enabled, so it's also tested in the CI

Scarjit (approved):

DaruZero:
Even with 5Gi of memory, RedPanda is able to handle at least 500 messages per second on a single topic. Tested using the sensorconnect mock and sensorconnect set to a lower polling time of 2 ms. Data processed using the following benthos config

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: benthos-1-config
  namespace: united-manufacturing-hub
  labels:
    app: benthos-1
data:
  benthos.yaml: |-
input:
  kafka:
    addresses:
    - united-manufacturing-hub-kafka:9092
    topics:
    - ia.raw.default.000201610192.X01
    consumer_group: ""
pipeline:
  processors:
    - bloblang: |
        root.button = this."Input 0".number()
        root.timestamp_ms = this.timestamp_ms.number() + random_int(31536000)
output:
  kafka:
    addresses:
    - united-manufacturing-hub-kafka:9092
    topic: "ia.factoryinsight.testLocation.benthos.processValue.ButtonPress"
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: benthos-1-deployment
  namespace: united-manufacturing-hub
  labels:
    app: benthos-1
spec:
  replicas: 1
  selector:
    matchLabels:
      app: benthos-1
  template:
    metadata:
      labels:
        app: benthos-1
    spec:
      containers:
        - name: benthos-1
          image: "ghcr.io/united-manufacturing-hub/benthos-umh:latest"
          imagePullPolicy: IfNotPresent
          ports:
            - name: http
              containerPort: 4195
              protocol: TCP
          livenessProbe:
            httpGet:
              path: /ping
              port: http
          readinessProbe:
            httpGet:
              path: /ready
              port: http
          volumeMounts:
            - name: config
              mountPath: "/benthos.yaml"
              subPath: "benthos.yaml"
              readOnly: true
      volumes:
        - name: config
          configMap:
            name: benthos-1-config
```

The biggest bottleneck was the database itself

deployment/united-manufacturing-hub/templates/kafkatopostgresql/deployment.yaml
```diff
@@ -55,11 +55,13 @@ spec:
           image: {{.Values.kafkatopostgresql.initContainer.repository}}:{{.Values.kafkatopostgresql.initContainer.tag}}
           {{- else}}
           image: {{.Values.kafkatopostgresql.initContainer.repository}}:{{.Chart.AppVersion}}
+          {{ if and (hasKey .Values._000_commonConfig.infrastructure.kafka "useSSL") .Values._000_commonConfig.infrastructure.kafka.useSSL }}
```
JeremyTheocharis, line 58: let's test whether all cases work here as expected. I remember we tried something like this before, but were never able to enable or disable it.

deployment/united-manufacturing-hub/values.yaml
```diff
@@ -1128,30 +1016,24 @@ kafkatoblob:
...
   storage:
     persistentVolume:
-      size: 8Gi
+      size: 10Gi
+  listeners:
+    kafka:
+      port: 9092
   resources:
     cpu:
       cores: 2
```
JeremyTheocharis, line 1032: lets remember to move it down to 1 core :)

JeremyTheocharis (approved):

deployment/united-manufacturing-hub/values.yaml
DaruZero: https://github.com/united-manufacturing-hub/united-manufacturing-hub/issues/1554

deployment/united-manufacturing-hub/templates/kafkatopostgresql/deployment.yaml
DaruZero: After solving this https://github.com/united-manufacturing-hub/united-manufacturing-hub/issues/1549 it will be easier to adjust this

PR #1548, opened 2023-04-19
Add Makefile

DaruZero:
# Description

This PR adds a Makefile in order to simplify frequently run commands 

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)

PR #1556, opened 2023-04-25
Add unit tests for factoryinsight v2 service-tag

Ayato77:
### Description
Test for the function converting string of the timebucket to duration


golang/cmd/factoryinsight/v2/services/service-tags_test.go
```diff
@@ -0,0 +1,98 @@
+package services
+
+import (
+	"fmt"
+	"testing"
+	"time"
+)
+
+func TestTimeBucketToDuration(t *testing.T) {
+	testMinuteString := "3m"
```
Scarjit, line 10: Please convert these to a map of input & expected output, then use a loop to check all of them, to keep the code more concise.

Makefile
```diff
@@ -0,0 +1,204 @@
+.PHONY: all
```
Scarjit, line 1: FYI for @JeremyTheocharis 
This change comes from the #1548 pr, this is based on.

Scarjit (changes requested):

golang/cmd/factoryinsight/v2/services/service-tags_test.go
Ayato77: The test uses map and list for inputs/expected results now.

JeremyTheocharis (approved):

Issue #1467, opened 2023-02-23
Modify statetable, disallowing unchanged states


Scarjit:
**Is your feature request related to a problem? Please describe.**
At one of our customers, an incorrect Node-Red led to millions of duplicate rows.
This resulted in instable Grafana

**Describe the solution you'd like**
Modify the statetable, so that a new entries state for a given asset_id must be different from the previosly entered state.

Add an guide, on how to remove duplicate values and how to apply this fix to old tables.

```
ALTER TABLE statetable ADD CONSTRAINT state_change_constraint CHECK (
  NOT EXISTS (
    SELECT 1
    FROM (
      SELECT 
        state, 
        LAG(state) OVER (PARTITION BY asset_id ORDER BY timestamp) AS prev_state
      FROM statetable
    ) subq
    WHERE prev_state IS NOT NULL AND prev_state = state
  )
);
```

```
BEGIN TRANSACTION;

WITH cte AS (
  SELECT *
  FROM (
    SELECT
      *,
      LAG(state) OVER (PARTITION BY asset_id ORDER BY timestamp ASC) AS prev_state
    FROM statetable
    WHERE asset_id = <ASSET_ID>
  ) subq
  WHERE prev_state IS NOT NULL AND prev_state = state
)
DELETE FROM statetable
WHERE (timestamp, asset_id) IN (SELECT timestamp, asset_id FROM cte);


SELECT * FROM statetable WHERE asset_id=<ASSET_ID> ORDER BY timestamp ASC LIMIT 100;
SELECT count(1) FROM statetable WHERE asset_id=<ASSET_ID>;

ROLLBACK;
```

PR #1577, opened 2023-05-02
Add a trigger which removes duplicated rows after insertions

Ayato77:
### Description
A trigger removing duplicated rows after insertions is added to resolve the problem #1467 

### Note
I used **CHECK** constraint mentioned in the issue, but it does not work because SQL checks only status of the statetable before insertion, i.e., all insertion after inserting a duplicated row will be refused. So, I decided to use **trigger** feature which triggered after every insertion.


deployment/united-manufacturing-hub/templates/timescaledb/configmap.yaml
```diff
@@ -196,6 +196,35 @@ data:
     -- #2
     ALTER TABLE configurationTable ADD IF NOT EXISTS AutomaticallyIdentifyChangeovers BOOLEAN DEFAULT true;
 
+    -- #51 state table
```
Scarjit, line 199: Instead of using a trigger here, can we make it use a constraing ?
So that we don't insert then remove, but check before inserting.

Scarjit (changes requested):

deployment/united-manufacturing-hub/templates/timescaledb/configmap.yaml
Ayato77: Updated so that it checks new inputs before inserting. Also, a transaction that repairs statetable is added.

PR #1617, opened 2023-07-06
Improve conn stability

Scarjit:


DaruZero (approved):
Make depguard happy. Other than that, LGTM

PR #1635, opened 2023-09-03
refactor grafana proxy and add helper for grateful shutdowns

ricci2511:
First of all, I added a new helper to our `internal` pkg for easily setting up graceful shutdowns. This should be more robust and will reduce code duplication by quite a bit, once implemented into all other services. Check out `golang/internal/gracefulShutdown_test.go` for an example.

Aside from that, I refactored some of the code related to the grafana proxy and fixed some issues that were present, especially in the `main.go` file. So let's begin with the issue, before this PR right here, the main function ended with:
```go
select {} // block forever
```
As the comment implies, this blocks until the program is explicitly terminated with an `os.Exit(0)`, which should be normally avoided if possible, reason being it immediately exits the program, so any other tasks will abruptly end and `defer` calls will not run. This last point takes me to the block of code located at the top of the main function (current code in main branch):
```go
logLevel, _ := env.GetAsString("LOGGING_LEVEL", false, "PRODUCTION") //nolint:errcheck
log := logger.New(logLevel)
defer func(logger *zap.SugaredLogger) {
	err := logger.Sync()
	if err != nil {
		panic(err)
	}
}(log)
```
So because of the select trick above to block forever, and the shutting down of the service with `os.Exit(0)`, this defer call never runs, but with the new graceful shutdown helper implementation it works as it should. Once I tested my changes and triggered a shutdown, the actual deferred function ran, great! But... the `logger.Sync()` call returned an error and therefore panicked, and this last thing is the reason for this [PR](https://github.com/united-manufacturing-hub/umh-utils/pull/4).

The rest of the changes are of refactoring nature to clean up the code. For example I turned the CORS handling of the main REST API to a middleware and introduced some dependency injection where it made sense to do so.

**NOTE:** CI will complain about the `logger.Sync` call because it's part of [this](https://github.com/united-manufacturing-hub/umh-utils/pull/4). And it also seems to complain about [`errors.Join`](https://pkg.go.dev/errors#Join) being undefined, the explanation for that is simple, the setup-go action seems to use Go version 1.19 and the `Join` method was added to Go 1.20. So I guess we also should update that.


DaruZero (commented):
Looks great! I have updated the go version in the CI and bumped umh-utils to 0.2.2 in [this](https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/1636) PR. You can wait for it to be merged (might take a while) and rebase, or you can merge that branch here to see if the checks are passing

ricci2511:
> Looks great! I have updated the go version in the CI and bumped umh-utils to 0.2.2 in [this](https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/1636) PR. You can wait for it to be merged (might take a while) and rebase, or you can merge that branch here to see if the checks are passing

Awesome. Yeah, whenever your PR is merged, I'll just rebase and push again, so no worries.

golang/cmd/grafana-proxy/main.go
```diff
@@ -15,110 +15,130 @@
...
+		case "FACTORYINPUT_KEY":
+			f.APIKey = val
+		case "FACTORYINPUT_USER":
+			f.User = val
+		case "FACTORYINPUT_BASE_URL":
+			f.BaseURL = formatUrl(val)
+		case "FACTORYINSIGHT_BASE_URL":
+			f.InsightBaseURL = formatUrl(val)
+		}
+	}
```
Scarjit, line 142: Please add a check if all 4 env variables have been set

Scarjit (commented):
Looks good

golang/cmd/grafana-proxy/main.go
ricci2511: Is it necessary, though? On line 127, I try to get the value associated with the env variable of the current iteration, and if not available, I throw an error, which is then caught in the main function. So if the loop succeeds, the `Factory` struct should be already properly Initialized.

Scarjit: true, have overlooked that

Scarjit (approved):

ricci2511:
> Looks good

Squash and merge?

PR #1643, opened 2023-09-12
Add base functionality for data bridge

DaruZero:
This PR adds the base functionality for the new data bridge.

The `main()` functions checks for the env variables and validates them, starts the health check, and defines the clients.
I decided to use a generic `client` interface in order to have a cleaner code. Both kafka and MQTT will implement those methods, each in their own way

Close https://github.com/united-manufacturing-hub/MgmtIssues/issues/344

PS: it'ok that the codecov checks fails, as some of the functions are not yet implemented. Notice that this PR wants to merge into `feat/338/create-data-bridge`, so that we can test in that branch once that all the functionalities have been added

### DIFF
--- a/golang/cmd/data-bridge/main.go
+++ b/golang/cmd/data-bridge/main.go
@@ -1,0 +1,99 @@
+package main
+
+import (
+	"net/http"
+
+	"github.com/heptiolabs/healthcheck"
+	"github.com/united-manufacturing-hub/umh-utils/env"
+	"github.com/united-manufacturing-hub/umh-utils/logger"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/internal"
+	"go.uber.org/zap"
+)
+
+func main() {
+	var err error
+	logLevel, _ := env.GetAsString("LOGGING_LEVEL", false, "PRODUCTION") //nolint:errcheck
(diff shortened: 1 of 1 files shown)


golang/cmd/data-bridge/main.go
```diff
@@ -138,6 +139,16 @@ func main() {
...
+	case 2: // mqtt to mqtt
+		zap.S().Infof("Starting mqtt to mqtt bridge")
+		clientA, err = newMqttClient(brokerA, topic, false, "")
+		if err != nil {
+			zap.S().Errorf("failed to create mqtt client: %s", err)
+		}
+		clientB, err = newMqttClient(brokerB, topic, false, "")
+		if err != nil {
+			zap.S().Errorf("failed to create mqtt client: %s", err)
+		}
```
Scarjit, line 151: Case 3: mqtt to kafka ?

Scarjit (changes requested):

golang/cmd/data-bridge/main.go
DaruZero: can be done in case 2, setting the correct bridge-mode

Scarjit (approved):

PR #1649, opened 2023-09-13
0.9.15 prerelease

Scarjit:
This early prerelease is mostly for testing of the new redpanda chart for the HA setup


Scarjit:
GitGuardian complains due to new default private key for databridge

DaruZero:
Is this prerelease 1 or 2?

DaruZero:
We should wait for this #1651 and then rebase

PR #1651, opened 2023-09-15
Create data bridge

DaruZero:
This PR add the logic for the new data-bridge microservice, which purpose is to bridge messages between Kafka and MQTT brokers, using the new data model and splitting the topic in case the producer sends data to Kafka.

Close https://github.com/united-manufacturing-hub/MgmtIssues/issues/338


JeremyTheocharis (changes requested):
See also slack

> I looked at the Kafka bridge, and I cannot find the commit logic. So that incoming messages from Kafka are only committed when have been send to MQTT successfully.

> am I missing it somewhere or is it somewhere in a library?

Scarjit:
Please decide on `data-bridge` or `databridge` as name :)

Scarjit:
Everything in here is now included in #1649

PR #1652, opened 2023-09-15
Add MQTT handler

DaruZero:
This PR adds the MQTT client that implements the `client` interface.
The MQTT client uses MQTT v3.1, since the v5 library is still in early stages of development

When using the `startConsumer` method, it starts reading the messages from the broker and sending them to the `msgChan`.
When using the `startProducer` method, it reads the messages from `msgChan`, transforms them from kafka-style to mqtt-style (if necessary) and does the splitting

Close https://github.com/united-manufacturing-hub/MgmtIssues/issues/346

### DIFF
--- a/golang/cmd/data-bridge/mqtt.go
+++ b/golang/cmd/data-bridge/mqtt.go
@@ -1,0 +1,231 @@
+package main
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/hex"
+	"fmt"
+	"os"
+	"regexp"
+	"strings"
+	"sync/atomic"
+
+	MQTT "github.com/eclipse/paho.mqtt.golang"
+	"github.com/goccy/go-json"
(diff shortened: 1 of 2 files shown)


golang/cmd/data-bridge/mqtt.go
```diff
@@ -0,0 +1,231 @@
...
+		zap.S().Fatalf("Connection lost to MQTT broker %s: %v", broker, err)
+	})
+
+	mc.client = MQTT.NewClient(opts)
+	if token := mc.client.Connect(); token.Wait() && token.Error() != nil {
+		return nil, token.Error()
+	}
+
+	var arcSize int
+	arcSize, err = env.GetAsInt("MESSSAGE_LRU_SIZE", false, 1_000_000)
```
Scarjit, line 74: Can you change this to be an ENV defaulting to 1 million ?

Scarjit (approved):

PR #1666, opened 2023-10-26
Chore/xxx/cleanup

JeremyTheocharis:
- threw out old tests that were flaky or otherwise not stable working
- removed Makefile and all documentation of it, as it was yet another way of using the UMH (we need to limit the options)
- only build normal containers now (no need for building them with race with every commit, it fills the amount of required checks unnecessary)


.github/workflows/autoupdate.yaml
```diff
@@ -23,7 +23,8 @@ on:
 jobs:
   autoupdate:
     name: autoupdate
-    runs-on: ubuntu-20.04
+    runs-on: 
+      group: large-runners
```
Scarjit, line 27: This really doesn't have to run on large-runners

go.work.sum
```diff
@@ -0,0 +1 @@
+github.com/ugorji/go v1.2.7 h1:qYhyWUUd6WbiM+C6JZAUkIJt/1WrjzNHY9+KCIjVqTo=
```
Scarjit, line 1: Why is this new dep here ?

Scarjit (approved):

PR #1688, opened 2023-11-18
Main

JeremyTheocharis:


.github/workflows/helm-release.yaml
```diff
@@ -14,43 +14,20 @@ jobs:
         with:
           fetch-depth: 0 # Important for semantic-release to analyze all commits
 
-      - name: Configure Git
+      - name: Install Helm
+        if: steps.semantic.outputs.new_release_published == 'true'
```
Scarjit, line 18: Can this even work ?
Semantic release get's called after this step ?

JeremyTheocharis: lol thx

PR #1707, opened 2023-12-07
Higher insertion speed for k2pg-v2

Scarjit:
This is based on https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/1705, which should be merged first.

TS tuning for lower memory usage
 - set wal pv size to `20Gi`
 - add patroni postgresql parameter `max_wal_size: 15GB`


Scarjit:
Reached the thermal limits of that nvme.

image

If we want to continue testing on that machine, we need to improve the airflow !

deployment/united-manufacturing-hub/templates/kafkatopostgresqlv2/deployment.yaml
```diff
@@ -86,6 +86,10 @@ spec:
               value: {{.Values.kafkatopostgresqlv2.asset.cache.lru.size | quote}}
             - name: VALUE_CHANNEL_SIZE
               value: {{.Values.kafkatopostgresqlv2.workers.channel.size | quote}}
+            - name: VALUE_TABLE_SIZE
```
JeremyTheocharis, line 89: Can we use a default value if this is not set yet? simplifies the upgrade process :)

- name: FLUSH_INTERVAL_SECONDS
  value: {{ .Values.kafkatopostgresqlv2.postgresql.flush.interval | default "10" | quote }}

JeremyTheocharis (changes requested):

deployment/united-manufacturing-hub/templates/kafkatopostgresqlv2/deployment.yaml
Scarjit: yes, can do

JeremyTheocharis (changes requested):

PR #1723, opened 2024-01-22
Remove unused dependencies

Ayato77:
- Unused dependencies in go removed by go mod tidy are removed
- Dependency of deprecated microservices in dependabot is already removed

close united-manufacturing-hub/MgmtIssues/issues/1483


Makefile
```diff
@@ -0,0 +1,402 @@
+# Copyright 2023 UMH Systems GmbH
```
JeremyTheocharis, line 1: Where is this file coming from?

JeremyTheocharis (changes requested):

JeremyTheocharis (approved):

Scarjit (approved):

PR #1736, opened 2024-01-30
Upgraded grafana to 9.5.5

Scarjit:
Chart version 6.57.4 installs 9.5.5 while the next higher 6.58.0 installs 10.X, therefore i cannot upgrade to 9.5.15, without major upgrade.

Closes https://github.com/united-manufacturing-hub/MgmtIssues/issues/1550


Scarjit:
Given that 9.5.5 was released in 22. Jun 2023 and https://github.com/grafana/grafana/pull/64149#issuecomment-1471808501 was merged onto main in March it is safe to assume that the fix is included.

ricci2511:
@Scarjit It might be worth it to trigger the full e2e test so that the grafana test runs, or should we just merge to staging, see what happens and then fix if needed?

Scarjit:
@ricci2511 Sadly we have to push this to main and release it to be able to do a e2e on it

DaruZero (approved):

PR #1753, opened 2024-03-01
work order in kafka-to-postgresql-v2

Scarjit:
Closes https://github.com/united-manufacturing-hub/MgmtIssues/issues/1658


JeremyTheocharis (approved):

golang/cmd/kafka-to-postgresql-v2/postgresql/analytics.go
```diff
@@ -0,0 +1,112 @@
...
+	defer cncl()
+	tx, err := c.db.Begin(ctx)
+	if err != nil {
+		return err
+	}
+
+	var cmdTag pgconn.CommandTag
+	cmdTag, err = tx.Exec(ctx, `
+		UPDATE work_orders
+		SET status = 1, startTime = to_timestamp($2 / 1000)
```
JeremyTheocharis, line 60: can we do an additional SQL check here ensuring that this only works if the status is 0 before and where there is no startTime set before?

then we can make this message idempotent (so it can execute 100 times and the result would still be the same). currently it could happen that a duplicated startWorkOrder message gets stuck somewhere and overwrites the stop work order message.

golang/cmd/kafka-to-postgresql-v2/postgresql/analytics.go
```diff
@@ -0,0 +1,112 @@
...
+	defer cncl()
+	tx, err := c.db.Begin(ctx)
+	if err != nil {
+		return err
+	}
+
+	var cmdTag pgconn.CommandTag
+	cmdTag, err = tx.Exec(ctx, `
+		UPDATE work_orders
+		SET status = 2, endTime = to_timestamp($2 / 1000)
```
JeremyTheocharis, line 94: same as before

JeremyTheocharis (changes requested):

JeremyTheocharis (approved):

PR #1757, opened 2024-03-04
[#1662 (0)] state in kafka-to-postgresql-v2

Scarjit:
Closes https://github.com/united-manufacturing-hub/MgmtIssues/issues/1662

### DIFF
--- a/.github/workflows/run_go_tests_k2pg-v2.yaml
+++ b/.github/workflows/run_go_tests_k2pg-v2.yaml
@@ -1,0 +1,21 @@
+name: kafka-to-postgresql-v2-go-tests
+on:
+  pull_request:
+    branches:
+      - main
+      - staging
+
+jobs:
+  run-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Set up Go 1.22
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.22
+      - name: Run tests (inside golang/cmd/kafka-to-postgresql-v2)
+        run: |
+          cd golang/cmd/kafka-to-postgresql-v2
+          go test -v ./...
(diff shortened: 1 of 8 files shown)


.github/workflows/run_go_tests_k2pg-v2.yaml
```diff
@@ -0,0 +1,21 @@
...
+  pull_request:
+    branches:
+      - main
+      - staging
+
+jobs:
+  run-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
```
JeremyTheocharis, line 12: can we add here the logic that it only runs if somthing was changed e.g., in the umh version? you could check for the installer script if something was changed there

golang/cmd/kafka-to-postgresql-v2/worker/analytics_test.go
```diff
@@ -235,3 +235,54 @@ func TestShiftDelete(t *testing.T) {
 		assert.Error(t, err)
 	})
 }
+
+func TestStateAdd(t *testing.T) {
```
JeremyTheocharis, line 239: in the future, you could create new tests using ginkgo directly, so then we would not need to convert them on  
a later stage. nothing to do here, just a note for future tests

JeremyTheocharis (changes requested):

.github/workflows/run_go_tests_k2pg-v2.yaml
Scarjit: It would be *way* easier here to check for changes inside the `deployment/united-manufacturing-hub` folder.

JeremyTheocharis: ahhh messed up the repos. thought we were in ManagementConsole yes, check for changes in your suggested folder, that is better

JeremyTheocharis (approved):

PR #1762, opened 2024-03-06
[#1662 (3)] datamodel v1 bug fixes

Scarjit:
Fixed various bugs that happened when executing the kafka-to-postgressqlv2 integration tests. The base code that is changed here is unreleased, so we can happily change it all.

As a result, the integration test passes.

I hope that nothing broke during branch split.
Based on https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/1761

PR #1763, opened 2024-03-06
[#1662 (4)] Run docker build in all pr's, conditionally

Scarjit:
Runs docker builds on every commit, but only if there are changes

Based on https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/1762


Scarjit:
hm, seems like we need to adjust the docker builders

JeremyTheocharis (approved):

JeremyTheocharis (approved):

PR #1769, opened 2024-03-08
fix: v2 tables & pg2k-v2 changes

Scarjit:


Scarjit:
@JeremyTheocharis All of these have been independently reviewed on the merges into staging.
Ill therefore just merge now

PR #1787, opened 2024-03-14
Fix: revert keys for hivemq

Ayato77:
Regarding reverting [this commit](https://github.com/united-manufacturing-hub/united-manufacturing-hub/commit/ee5fbd505d21feca1fc5701acfcb86860382729c), the key should be also moved to values.yaml.


Scarjit:
This already seems to be on main, but not on staging:
https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/1788

Ayato77:
> This already seems to be on main, but not on staging: #1788

Ok, then I set to Draft and close this PR

PR #1792, opened 2024-03-25
Upgrade grafana

Ayato77:
### Description
This change upgrades Grafana from v9.5.5 to v10.4.0

- New installation and data flow is tested
- Upgrading from the current UMH version is tested and was successful. Data flow was tested and also, visualization on Grafana using [our data visualization guide](https://umh.docs.umh.app/docs/getstarted/datavisualization/) was successful. Our guide should be updated accordingly. 

#### Breaking changes (they should be added into our change log)
- Angular plugin turned off by default, but one can still enable it
- When you upgrade to Grafana v10, Grafana will automatically migrate all API keys to service accounts and hide the API keys screen that was under the Admin section
- Some data source permission endpoints are removed, so one should use another endpoint if needed
- Some transformation by table visualization and transformation with time fields might break, so we should give information how to resolve it (navigate to the Grafana's change log https://grafana.com/docs/grafana/latest/breaking-changes/breaking-changes-v10-3/).


JeremyTheocharis:
Can you write the notes that one needs to do when upgrading from an old UMH version? Can you also check if the old datasource plugins work?

Ayato77:
> Can you write the notes that one needs to do when upgrading from an old UMH version? Can you also check if the old datasource plugins work?

- `umh-datasource` works also after upgrading. `umh-datasource-v2` has a bug on the current version (0.13.6) and reported in [this issue](https://github.com/united-manufacturing-hub/MgmtIssues/issues/1820). 
-  I created another issue for writing notes task: https://github.com/united-manufacturing-hub/MgmtIssues/issues/1816

JeremyTheocharis (approved):

PR #1793, opened 2024-04-03
Upgrade timescaledb chart

Ayato77:
### Description
This change upgrades timescadb 2.8.0 to 2.9.1 and postgresql 13.8 to 13.9. Since databases created by postgresql 13 are not compatible with 14, use of 14 is prevented (Otherwise, user has to backup and delete all persistent volumes). Due to changes in timescaledb's values.yaml, some values had to be adjusted.

- A new installation and upgrade from UMH v0.13.6 work successfully: installation/upgrade and  data flow to umh_v2, factoryinsight, and grafana database are checked.

- Upgrade guide should be created by https://github.com/united-manufacturing-hub/MgmtIssues/issues/1835

close united-manufacturing-hub/MgmtIssues/issues/1649


JeremyTheocharis:
I think we can upgrade this only when we have the Upgrade UMH functionality in the Management Console. But once we have it, we can easily do this one here.

PR #1862, opened 2025-03-25
ENG-2729 Resolve coderabbit issues.

Scarjit:


umh-core/pkg/fsm/benthos/actions.go
```diff
@@ -245,7 +245,13 @@ func (b *BenthosInstance) AnyRestartsSinceCreation() bool {
 		return false
 	}
 
-	return true
+	for _, exit := range b.ObservedState.ServiceInfo.S6ObservedState.ServiceInfo.ExitHistory {
+		if exit.ExitCode == 0 && exit.Timestamp.After(time.Now().Add(-10*time.Second)) {
```
JeremyTheocharis, line 249: i dont think this is what we want here:
1. if it gets OOM killed, it would now be not detected anymore (but it makes sense to detect those). the code comment is misleading there.
2. my recommendation for easier testability and maintainability here would be to pass the time.now to the main reconcile function (i did this similarly with the logs), and to put the 10 seconds as a constant in the const package. Reason: the entire code base works only with "ticks", but for some things we just have to use time-based approaches. for this i think it could be worthwhile to highlight it up to the main reconcile function, that something requires the current time. otherwise we have these hidden non-deterministc behaviours (e.g., if a single reconcile loop has two time.now() calls, it will vehave unexpected.

JeremyTheocharis (changes requested):

umh-core/pkg/fsm/benthos/actions.go
Scarjit: Now using tickStartTime

Scarjit: Also that one is now passed basically everywhere, due to our architecture requiring the interfaces to be the same.

JeremyTheocharis (approved):

PR #1863, opened 2025-03-25
ENG-2730 Implement sentry reporting.

Scarjit:
Note: These are only the stubs to be called later

### DIFF
--- a/umh-core/cmd/main.go
+++ b/umh-core/cmd/main.go
@@ -23,12 +23,18 @@ import (
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/control"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/logger"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/metrics"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/sentry"
 	"go.uber.org/zap"
 )
+
+var appVersion string // set by the build system
 
 func main() {
 	// Initialize the global logger first thing
 	logger.Initialize()
+
+	// Initialize Sentry
+	sentry.InitSentry(appVersion)
 
 	// Get a logger for the main component
 	log := logger.For(logger.ComponentCore)
--- a/umh-core/pkg/sentry/sentry.go
+++ b/umh-core/pkg/sentry/sentry.go
@@ -1,0 +1,56 @@
+package sentry
+
+import (
+	"os"
+	"strconv"
+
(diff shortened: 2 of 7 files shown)


umh-core/pkg/sentry/sentry.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	if rate := os.Getenv("SENTRY_SAMPLE_RATE"); rate != "" {
+		if parsedRate, err := strconv.ParseFloat(rate, 64); err == nil {
+			sampleRate = parsedRate
+		}
+	}
+
+	err = sentry.Init(sentry.ClientOptions{
+		Dsn:              "https://abc@staging.management.umh.app/sentry",
```
JeremyTheocharis, line 35: staging?

JeremyTheocharis (changes requested):

umh-core/cmd/main.go
```diff
@@ -35,6 +35,7 @@ func main() {
 
 	// Initialize Sentry
 	sentry.InitSentry(appVersion)
+	sentry.ReportIssuef(sentry.IssueTypeFatal, logger.For(logger.ComponentCore), "TEST")
```
JeremyTheocharis, line 38: hmm, i think this should be likely be removed :)

umh-core/Makefile
```diff
@@ -16,14 +16,16 @@
 
 IMAGE_NAME = umh-core
 TAG = latest
+# Dynamically select podman if available, otherwise use docker
+OCI_RUNTIME := $(shell command -v podman >/dev/null 2>&1 && echo podman || echo docker)
```
JeremyTheocharis, line 20: are we sure we want this? the advantage of a devcontainer is, that we standardize the development environment. i can already see how this could cause to future issues, because some people work in docker, and some work in podman, so maybe some make commands will only run in one of those.

if everyone introduces his own slight adjustments to his local environment, then it will not become clear for a new person what he needs to do to set it up.

JeremyTheocharis: we can also use podman, but then we need to adjust the devcontainer to work with podman as well

JeremyTheocharis (changes requested):

umh-core/Makefile
JeremyTheocharis: can we take this out for now and discuss this evening?

JeremyTheocharis (changes requested):

umh-core/pkg/sentry/sentry.go
Scarjit: Yes, our apps always seem to use staging.management.umh.app

JeremyTheocharis (approved):

PR #1865, opened 2025-03-25
Eng 2735 Setup Status Message Model

ricci2511:
Configure the model of an instance's Status Message as per our discussions.
Whether types/structs should be separated into different files is up to debate.

## Summary by CodeRabbit

- **New Features**
	- Introduced enhanced system status monitoring that provides comprehensive insights into overall system health and performance across core functionalities and plugins, enabling improved diagnostics and reporting.
	- Added detailed structures for monitoring various components such as agents, containers, and data flow components, enhancing the clarity and usability of system status reporting.


umh-core/pkg/generated/status-message.go
```diff
@@ -0,0 +1,317 @@
...
+	Avg float64 `json:"avg"`
+	Max float64 `json:"max"`
+	Min float64 `json:"min"`
+	P95 float64 `json:"p95"`
+	P99 float64 `json:"p99"`
+}
+
+type Redpanda struct {
+	Health     Health  `json:"health"`
+	Throughput float64 `json:"throughput"`
```
JeremyTheocharis, line 155: we need throughput for incoming and outgoing.

 I recommend here stating the unit in the name, like ThroughputIncomingInMsgSec. this will prevent a lot of issues about what unit its is...

umh-core/pkg/generated/status-message.go
```diff
@@ -0,0 +1,317 @@
...
+
+type CPU struct {
+	Health Health  `json:"health"`
+	Limit  float64 `json:"limit"`
+	Usage  float64 `json:"usage"`
+}
+
+type Disk struct {
+	Health Health  `json:"health"`
+	Limit  float64 `json:"limit"`
```
JeremyTheocharis, line 83: also here again with the units. Is it bytes? is it percent?

umh-core/pkg/generated/status-message.go
```diff
@@ -0,0 +1,317 @@
...
+type Container struct {
+	Architecture string `json:"architecture"`
+	CPU          CPU    `json:"cpu"`
+	Disk         Disk   `json:"disk"`
+	Hwid         string `json:"hwid"`
+	Memory       Memory `json:"memory"`
+}
+
+type CPU struct {
+	Health Health  `json:"health"`
```
JeremyTheocharis, line 76: what would be the potential states for CPU?

umh-core/pkg/generated/status-message.go
```diff
@@ -0,0 +1,317 @@
...
+	Health Health `json:"health"`
+}
+
+type Health struct {
+	Message string `json:"message"`
+	State   string `json:"state"`
+}
+
+type Container struct {
+	Architecture string `json:"architecture"`
```
JeremyTheocharis, line 68: Can we add codecomments here, maybe trogu hthe generator? It does really help when you can see directly in the code what the format for architecture here is. is it arm/v7, armv7, armhf, etc.

JeremyTheocharis: the JSON is unfortuantely only very ahrd to read, i tend to only look at the go structs

umh-core/pkg/generated/status-message.go
```diff
@@ -0,0 +1,317 @@
...
+type Memory struct {
+	Health Health  `json:"health"`
+	Limit  float64 `json:"limit"`
+	Usage  float64 `json:"usage"`
+}
+
+type Dfc struct {
+	Connections        []Connection `json:"connections,omitempty"`
+	CurrentVersionUUID *string      `json:"currentVersionUUID"`
+	DeploySuccess      bool         `json:"deploySuccess"`
```
JeremyTheocharis, line 96: does that make sense here?

umh-core/pkg/generated/status-message.go
```diff
@@ -0,0 +1,317 @@
...
+}
+
+type Memory struct {
+	Health Health  `json:"health"`
+	Limit  float64 `json:"limit"`
+	Usage  float64 `json:"usage"`
+}
+
+type Dfc struct {
+	Connections        []Connection `json:"connections,omitempty"`
```
JeremyTheocharis, line 94: why a connection array?

umh-core/pkg/generated/status-message.go
```diff
@@ -0,0 +1,317 @@
...
+	CurrentVersionUUID *string      `json:"currentVersionUUID"`
```
JeremyTheocharis, line 95: do we want to stick with the current approach of versioning, or can we maybe find something more idiomatic?

umh-core/pkg/generated/status-message.go
```diff
@@ -0,0 +1,317 @@
...
+	DataContract       *string      `json:"dataContract,omitempty"`
+	InputType          *string      `json:"inputType,omitempty"`
+	IsReadOnly         *bool        `json:"isReadOnly,omitempty"`
+	OutputType         *string      `json:"outputType,omitempty"`
+}
+
+type Connection struct {
+	Health  Health  `json:"health"`
+	Host    string  `json:"host"`
+	Latency float64 `json:"latency"`
```
JeremyTheocharis, line 111: what does this field mean? we also have a struct with latency. what type of latency is it?

umh-core/pkg/generated/status-message.go
```diff
@@ -0,0 +1,317 @@
...
+	UUID               string       `json:"uuid"`
+	DataContract       *string      `json:"dataContract,omitempty"`
+	InputType          *string      `json:"inputType,omitempty"`
+	IsReadOnly         *bool        `json:"isReadOnly,omitempty"`
+	OutputType         *string      `json:"outputType,omitempty"`
+}
+
+type Connection struct {
+	Health  Health  `json:"health"`
+	Host    string  `json:"host"`
```
JeremyTheocharis, line 110: what is the format for host? is it opc.tcp://123123:23123/whatever, or is it just the IP?

umh-core/pkg/generated/status-message.go
```diff
@@ -0,0 +1,317 @@
...
+	General General                `json:"general"`
+	Plugins map[string]interface{} `json:"plugins"`
+}
+
+type Core struct {
+	Agent             Agent                  `json:"agent"`
+	Container         Container              `json:"container"`
+	Dfcs              []Dfc                  `json:"dfcs"`
+	EventsTable       map[string]EventsTable `json:"eventsTable,omitempty"`
+	Latency           Latency                `json:"latency"`
```
JeremyTheocharis, line 50: what latency is this? isnt this the latency of the agent?

umh-core/pkg/generated/status-message.go
```diff
@@ -0,0 +1,317 @@
...
+	Redpanda          Redpanda               `json:"redpanda"`
+	ReleaseChannel    string                 `json:"releaseChannel"`
+	SupportedFeatures []string               `json:"supportedFeatures"`
```
JeremyTheocharis, line 53: arent these the supported features of the agent?

umh-core/pkg/generated/status-message.go
JeremyTheocharis, line 94: maybe it would make sense to directly include something like "desiredState"? so that you can directly see when an instance is "paused", and maybe it is then quite easy to implement a pause action

JeremyTheocharis (changes requested):

umh-core/pkg/generated/status-message.go
ricci2511: Yes. Tbh json schema is very annoying to work with and i'm considering that maybe we can go back to just maintaining the go structs and afterwards generate TS types from it for the client with something like https://github.com/gzuidhof/tygo

umh-core/pkg/generated/status-message.go
ricci2511: This is up to debate. I know that depending on the entity, there are different kinds of states, but if these states are given to the client, we should have a few defined states (enum like) so that the client knows how to behave accordingly.

umh-core/pkg/generated/status-message.go
ricci2511: I'll make it a full `uri `instead. Splitting host and port if needed is very easy to do anyways.

Scarjit (approved):

PR #1873, opened 2025-03-28
Fixed integration test by reducing p99 of HTTP requests

JeremyTheocharis:
## Key Changes:

Timeout Enforcement:
The HTTP client is now configured with connection and response timeouts based on the context deadline. This ensures each HTTP request (e.g., /ping, /ready, /version, /metrics) adheres to the expected time constraints.

Early Cancellation with errgroup:
Instead of directly calling g.Wait(), the wait is executed in a separate goroutine. A select statement is then used to either capture the error from the errgroup or immediately return if the context is canceled (or deadline exceeded). This prevents the function from waiting on slow or unresponsive goroutines, ensuring that the reconcile loop is not blocked.

Improved Reliability:
All HTTP requests now use a derived context (gctx) from errgroup.WithContext(ctx), so that if one request fails or if the overall deadline is exceeded, all in-flight requests are canceled promptly.

## Rationale:

Preventing Overruns:
The previous implementation could occasionally exceed the target duration **(observed up to 70ms in some cases)** because g.Wait() blocks until all goroutines complete. By running g.Wait() in its own goroutine and using a select to monitor the context, we ensure an early exit when deadlines are exceeded.

Maintaining System Health:
Given that the health check function is part of a critical reconcile loop, it is imperative that a single misbehaving HTTP call does not delay the entire loop. This change helps keep the p99 execution times within acceptable limits, as demonstrated in our performance tests.

## Summary by CodeRabbit

- **New Features**
  - Introduced a context-aware HTTP client for standardized request handling, including timeout management and error reporting.
  - Added a new method to the mock HTTP client for performing GET requests with body retrieval.

- **Refactor**
  - Improved service health monitoring by enabling concurrent status checks, leading to faster, more reliable responses and clear error reporting during system health assessments.

- **Chores**
  - Updated workflow configurations to trigger builds and tests on pull requests, enhancing the CI/CD process.
  - Upgraded underlying dependencies to better support concurrency and overall system stability.

- **Documentation**
  - Added comprehensive documentation for the new HTTP client service, outlining its features and usage.


JeremyTheocharis:
will not fix the sentry issues as they have been fixed already in multiple PRs

umh-core/pkg/service/benthos/benthos.go
```diff
@@ -727,111 +730,187 @@ func (s *BenthosService) GetHealthCheckAndMetrics(ctx context.Context, s6Service
 
 	baseURL := fmt.Sprintf("http://localhost:%d", metricsPort)
 
-	// Helper function to make HTTP requests with context
-	doRequest := func(endpoint string) (*http.Response, error) {
-		start := time.Now()
+	// Create a client to use for our requests
+	// If it's a mock client (used in tests), use it directly
```
Scarjit, line 734: Let's extract this logic so we can re-use it.

Scarjit (approved):

PR #1881, opened 2025-03-29
Even more timeouts

JeremyTheocharis:
I added even more timeout protections to prevent a single component taking down the entire reconcile loop.

Likely I am doing a little bit overkill here, but I just want to ensure that the strict integration / chaos test, that will fail when the timeouts are not properly working, is not flaky.

the general idea is that the loop will only call the next "main manager" if it has at least 50ms left to run. Benthos instance will only run, if it has at least 40ms, and S6 (which is included in benthos) only 30ms. in most cases, this will all take 1-5ms, but just for the edge cases where it might maybe not

## Summary by CodeRabbit

- **New Features**
  - Enhanced configuration retrieval with improved context cancellation and timeout handling for increased reliability.
  - Introduced new constants for better control over timing and execution parameters in the control loop.

- **Bug Fixes**
  - Improved logging for error reporting during the reconciliation process.

- **Refactor**
  - Improved error handling and logging for context deadlines and time sufficiency checks during reconciliation processes.
  - Enhanced HTTP client creation process to handle contexts without deadlines more robustly.

- **Tests**
  - Added new test files and expanded coverage for context management and timeout functionalities to ensure robustness under various scenarios.


JeremyTheocharis:
test runs fine in CI/CD, but it fails on my PC because of some weird filesystem calls suddenly taking 120ms. seems likely that it is my system as the integration test keeps passing without any issues in the CI/CD

JeremyTheocharis:
Ok, apparently it also fails in CI/CD, then let me take a look at it. Maybe it is a little bit too much timeouts here

Scarjit (approved):

PR #1887, opened 2025-04-02
ENG-2754 DFC & redpanda optimizations

Scarjit:


umh-core/pkg/fsm/dataflowcomponent/benthos_config_manager.go
```diff
@@ -0,0 +1,267 @@
...
+	logger.Debugf("Creating new BenthosConfigManager with config file path: %s", configFilePath)
+	return &DefaultBenthosConfigManager{
+		configFilePath: configFilePath,
+		logger:         logger,
+		fs:             filesystem.NewDefaultService(),
+	}
+}
+
+// readFullConfig reads the full config file
+func (m *DefaultBenthosConfigManager) readFullConfig(ctx context.Context) (*config.FullConfig, error) {
```
FreiPaul, line 59: We already have read/write config functions in the config manager

FreiPaul (changes requested):

umh-core/examples/example-config-dataflow.yaml
```diff
@@ -0,0 +1,68 @@
...
+  metricsPort: 8080
+
+redpanda:
+  desiredState: stopped
+
+# Data Flow Components
+dataFlowComponents:
+  - name: data-flow-hello-world
+    desiredState: active
+    versionUUID: 123e4567-e89b-12d3-a456-426614174000
```
JeremyTheocharis, line 25: does versionUUID make sense? why not simply use UUID?

Scarjit: versionUUID is in here, to already provide information for the upcoming versioning feature.

umh-core/pkg/fsm/dataflowcomponent/benthos_config_manager.go
Scarjit: This will be removed entirely

FreiPaul (approved):

umh-core/examples/example-config-dataflow.yaml
Scarjit: removed, due to our communication today.

PR #1888, opened 2025-04-02
Added performance instructions to mdc

Scarjit:
Adds some instructions from https://goperf.dev/ to mdc.

## Summary by CodeRabbit

- **Documentation**
  - Expanded the Go development guidelines with a new section titled "Efficient Go" focused on performance optimization techniques.
  - Added detailed recommendations on various strategies such as Object Pooling, Memory Preallocation, and Goroutine Worker Pools to enhance application performance.


.cursor/rules/golang.mdc
```diff
@@ -3,9 +3,110 @@ description: Golang
...
 
 # Golang Rules
 
 - We use Ginkgo v2 & Gomega for testing
 - Do not remove focused tests
-- Do not remove existing documentation, unless they are being replaced with a better or more up to date one.
\ No newline at end of file
+- Do not remove existing documentation, unless they are being replaced with a better or more up to date one.
+
+## Efficient Go
```
JeremyTheocharis, line 18: Can we add that to only "trigger" when performance is really needed? I want to prevent writing overly complicated code, that is unnecessarily performant. Something like "when the user explicitly requests this". Furthermore, I think applying this on our filesystem logic would probably give us a huge performance boost, so to fetch the entire S6 directory at the beginning of the loop, and then use it throughout the same tick.

JeremyTheocharis (changes requested):

.cursor/rules/golang.mdc
Scarjit: This already contains sections to advice the AI on when to use this

kanapuli (approved):

PR #1891, opened 2025-04-03
ENG-2769: Proper Structure for Actions

FreiPaul:
## Summary by CodeRabbit

- **New Features**
  - Introduced detailed action system documentation alongside a new action type for more flexible user operations.

- **Enhancements**
  - Upgraded configuration management with improved backoff logic and atomic updates for increased reliability.
  - Enhanced system state handling and routing for smoother performance and consistent instance updates.
  - Updated latency metrics to provide values in milliseconds for clearer performance insights.
  - Added new fields to the `CommunicationState` and `Router` for improved context management during operations.
  - Introduced a new type for instance location management to streamline configuration handling.
  - Improved test structures for better integration with configuration management.
  - Implemented context-aware mutexes for enhanced concurrency control in configuration management.

- **Chores**
  - Streamlined dependencies and reorganized code structure to boost overall maintainability and efficiency.


go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/united-manufacturing-hub/united-manufacturing-hub
```
Scarjit, line 1: This was probably an accident ?

umh-core/pkg/communicator/actions/edit_instance.go
```diff
@@ -147,85 +121,37 @@ func (a *EditInstanceAction) Execute() (interface{}, map[string]interface{}, err
 	SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionExecuting, "Updating instance location", a.outboundChannel, models.EditInstance)
 
 	// Update the location in the configuration
-	err := a.updateLocation()
+	err := a.configManager.AtomicSetLocation(context.Background(), *a.location)
```
Scarjit, line 124: This should probably use a smaller ctx then background

umh-core/pkg/config/mock.go
```diff
@@ -16,17 +16,21 @@ package config
...
-	GetConfigCalled bool
-	Config          FullConfig
-	ConfigError     error
-	ConfigDelay     time.Duration
-	mutex           sync.Mutex
+	GetConfigCalled   bool
+	Config            FullConfig
+	ConfigError       error
+	ConfigDelay       time.Duration
+	mutexReadOrWrite  sync.Mutex
```
JeremyTheocharis, line 32: please explain the mutexes, use RWMutex, etc. as discussed in our meeting jsut now

JeremyTheocharis (changes requested):

umh-core/pkg/communicator/actions/edit_instance_test.go
```diff
@@ -327,7 +329,7 @@ func (w *writeFailingMockConfigManager) GetConfig(ctx context.Context, tick uint
 	return w.mockConfigManager.GetConfig(ctx, tick)
 }
 
-func (w *writeFailingMockConfigManager) WriteConfig(ctx context.Context, config config.FullConfig) error {
+func (w *writeFailingMockConfigManager) writeConfig(ctx context.Context, config config.FullConfig) error {
```
JeremyTheocharis, line 332: maybe move the mock to its own file

umh-core/pkg/config/manager.go
```diff
@@ -38,16 +38,41 @@ const (
...
-	// WriteConfig writes the current config to the file
-	WriteConfig(ctx context.Context, config FullConfig) error
 	// AtomicSetLocation sets the location in the config atomically
 	AtomicSetLocation(ctx context.Context, location models.EditInstanceLocationModel) error
 }
 
+// GetConfigManager returns the singleton instance of ConfigManager
+func GetConfigManager() ConfigManager {
+	once.Do(func() {
+		// Default to FileConfigManagerWithBackoff as it's the more robust implementation
```
JeremyTheocharis, line 62: produce error message if the instance already exists

umh-core/pkg/config/manager.go
```diff
@@ -220,7 +248,12 @@ func NewFileConfigManagerWithBackoff() *FileConfigManagerWithBackoff {
 	}
 }
 
+func (m *FileConfigManagerWithBackoff) GetConfigWithOverwritesOrCreateNew(ctx context.Context, config FullConfig) (FullConfig, error) {
+	return m.configManager.GetConfigWithOverwritesOrCreateNew(ctx, config)
+}
+
 func (m *FileConfigManager) WriteConfig(ctx context.Context, config FullConfig) error {
+	// we use a write lock here, because we write the config file
```
JeremyTheocharis, line 256: do not expose writeConfig

umh-core/pkg/config/manager.go
```diff
@@ -59,14 +84,16 @@ type FileConfigManager struct {
 	// logger is the logger for the config manager
 	logger *zap.SugaredLogger
 
-	// mutexReadAndWrite for full cycle read and write access to the config file
-	mutexReadAndWrite sync.Mutex
+	// mutexAtomicUpdate for full cycle read and write access (atomic update) to the config file
```
JeremyTheocharis, line 87: when to use mutexAtomicUpdate: 
1. all writes need to go through an atomic update function, this requires that writeConfig is not exposed externally
2. therefore we can assume that all atomic update functions use this mutex

JeremyTheocharis: the goal is:
1. to prevent two read/write cycles "atomic updates" happening at the same time

umh-core/pkg/config/manager.go
```diff
@@ -59,14 +84,16 @@ type FileConfigManager struct {
...
+	mutexAtomicUpdate sync.Mutex
 
 	// simple mutex for read access or write access to the config file
-	mutexReadOrWrite sync.Mutex
+	mutexReadOrWrite sync.RWMutex
```
JeremyTheocharis, line 91: when to use mutexReadOrWrite:
1. it will be used by GetConfig or writeConfig
2. writeConfig is only to be used in this package through atomic functions

the goal is to:
1. allow multiple GetConfigs / read operations
2. but prevent reading and writing at the same time

umh-core/pkg/config/manager.go
```diff
@@ -142,8 +169,9 @@ func (m *FileConfigManager) GetConfigWithOverwritesOrCreateNew(ctx context.Conte
 
 // GetConfig returns the current config, always reading fresh from disk
 func (m *FileConfigManager) GetConfig(ctx context.Context, tick uint64) (FullConfig, error) {
-	m.mutexReadOrWrite.Lock()
-	defer m.mutexReadOrWrite.Unlock()
+	// we use a read lock here, because we only read the config file
```
JeremyTheocharis, line 172: this will block indefnitely ignoring any cancels of the context --> need to fix that

maybe there is a GetMutexOrCancelContext function somewher

umh-core/pkg/config/manager.go
```diff
@@ -220,7 +248,12 @@ func NewFileConfigManagerWithBackoff() *FileConfigManagerWithBackoff {
 	}
 }
 
+func (m *FileConfigManagerWithBackoff) GetConfigWithOverwritesOrCreateNew(ctx context.Context, config FullConfig) (FullConfig, error) {
```
JeremyTheocharis, line 251: function comment

umh-core/pkg/config/manager.go
```diff
@@ -220,7 +248,12 @@ func NewFileConfigManagerWithBackoff() *FileConfigManagerWithBackoff {
...
+	return m.configManager.GetConfigWithOverwritesOrCreateNew(ctx, config)
+}
+
 func (m *FileConfigManager) WriteConfig(ctx context.Context, config FullConfig) error {
```
JeremyTheocharis, line 255: function comment, incl. why this should not be exposed

umh-core/pkg/config/manager.go
```diff
@@ -330,8 +363,27 @@ func (m *FileConfigManagerWithBackoff) WriteConfig(ctx context.Context, config F
 
 // AtomicSetLocation sets the location in the config atomically
 func (m *FileConfigManager) AtomicSetLocation(ctx context.Context, location models.EditInstanceLocationModel) error {
-	m.mutexReadAndWrite.Lock()
-	defer m.mutexReadAndWrite.Unlock()
+	// Check if context is already cancelled
```
JeremyTheocharis, line 366: https://chatgpt.com/share/67effa0b-19c4-8010-a22d-09097ba9a1dc

JeremyTheocharis (changes requested):

umh-core/pkg/communicator/actions/edit_instance_test.go
FreiPaul: It is only three small functions that are tailor suited for this action test so I would keep it here

umh-core/pkg/communicator/actions/edit_instance.go
```diff
@@ -21,54 +21,31 @@ import (
...
-func NewEditInstanceAction(userEmail string, actionUUID, instanceUUID uuid.UUID, outboundChannel chan *models.UMHMessage) *EditInstanceAction {
+// exposed for testing purposes
+func NewEditInstanceAction(userEmail string, actionUUID uuid.UUID, instanceUUID uuid.UUID, outboundChannel chan *models.UMHMessage, configManager config.ConfigManager) *EditInstanceAction {
 	return &EditInstanceAction{
 		userEmail:       userEmail,
 		actionUUID:      actionUUID,
 		instanceUUID:    instanceUUID,
 		outboundChannel: outboundChannel,
-		configManager:   config.NewFileConfigManager(),
+		configManager:   configManager,
```
JeremyTheocharis, line 45: should it be a pointer or not? just wanting to make sure that our mutexes on the config file work as expected

umh-core/pkg/communicator/actions/edit_instance.go
```diff
@@ -147,85 +122,39 @@ func (a *EditInstanceAction) Execute() (interface{}, map[string]interface{}, err
 	SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionExecuting, "Updating instance location", a.outboundChannel, models.EditInstance)
 
 	// Update the location in the configuration
-	err := a.updateLocation()
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
JeremyTheocharis, line 125: maybe move the 5 seconds into the constant package?

umh-core/pkg/config/ctxrwmutex/ctxrwmutex.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// RUnlock unlocks the mutex for reading
+func (m *CtxRWMutex) RUnlock() {
+	m.sem.Release(1)
+}
+
+// Lock locks the mutex for writing
+func (m *CtxRWMutex) Lock(ctx context.Context) error {
+	return m.sem.Acquire(ctx, 100)
```
JeremyTheocharis, line 50: maybe make the 100 a constant and move to constants package.

umh-core/pkg/config/manager.go
JeremyTheocharis: what about this here? this doesn't produce an error yet, so it can lead to incorrect implemenetations. This here should fail if GetConfigManager is called a second time.

umh-core/pkg/config/manager.go
```diff
@@ -36,10 +38,47 @@
...
+		// Instance already exists, log an error
+		loggerInstance := logger.For(logger.ComponentConfigManager)
+		loggerInstance.Error("Attempted to set ConfigManager instance that already exists. This is likely a programming error.")
+		return
+	}
+
+	once.Do(func() {
+		instance = manager
+	})
 }
```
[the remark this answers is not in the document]
JeremyTheocharis: makes sense

umh-core/pkg/config/ctxrwmutex/ctxrwmutex.go
```diff
@@ -0,0 +1,42 @@
+package ctxrwmutex
+
+import (
+	"context"
+
+	"golang.org/x/sync/semaphore"
+)
+
+// CtxRWMutex is a context aware RWMutex
```
JeremyTheocharis, line 23: maybe add a comment here that the only reason we are using a weighted semaphore is because of the context.

umh-core/pkg/communicator/actions/edit_instance.go
FreiPaul: Since NewFileConfigManagerWithBackoff returns a pointer, this is a pointer to the interface. It's never dereferenced explicitly with * because it's used through interface methods.

JeremyTheocharis (approved):

PR #1897, opened 2025-04-04
nmap_fsm

led0nk:
## Summary by CodeRabbit

- **New Features**
  - Introduced Nmap monitoring with a finite state machine managing lifecycle and port state transitions.
  - Added YAML configuration example and integration builder for Nmap services.
  - Integrated Nmap management into control loops and metrics tracking.
  - Enhanced mock services with detailed scan logs, error simulation, and forced removal capability.
  - Added forced removal capability for Nmap instances in failure states.
  - Incorporated configurable backoff mechanisms across FSM instances for improved retry handling.

- **Bug Fixes**
  - Improved error detection in Nmap scan outputs and refined forced removal handling.

- **Tests**
  - Included extensive unit and integration tests validating Nmap FSM transitions, error handling, and lifecycle.

- **Chores**
  - Updated logging and metrics components to include Nmap.
  - Refined backoff configuration usage across FSM instances for better reliability.

### DIFF
--- a/umh-core/examples/example-config-nmap.yaml
+++ b/umh-core/examples/example-config-nmap.yaml
@@ -21,7 +21,7 @@ internal:
     desiredState: stopped
   nmap:
     - name: hello-world
-      desiredState: active
+      desiredState: open
       nmapServiceConfig:
         target: "127.0.0.1"
         port: 443
--- a/umh-core/examples/example-config-nmap.yaml
+++ b/umh-core/examples/example-config-nmap.yaml
@@ -1,0 +1,27 @@
+# Copyright 2025 UMH Systems GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+agent:
+  metricsPort: 8080
+
+internal:
+  redpanda:
+    desiredState: stopped
+  nmap:
+    - name: hello-world
+      desiredState: active
+      nmapServiceConfig:
+        target: "127.0.0.1"
+        port: 443
--- a/umh-core/pkg/control/loop_test.go
+++ b/umh-core/pkg/control/loop_test.go
@@ -128,7 +128,7 @@ var _ = Describe("ControlLoop", func() {
 			loop := NewControlLoop(mockConfig)
 			Expect(loop).NotTo(BeNil())
 			Expect(loop.tickerTime).To(Equal(constants.DefaultTickerTime))
-			Expect(loop.managers).To(HaveLen(4))
+			Expect(loop.managers).To(HaveLen(5))
 			Expect(loop.configManager).NotTo(BeNil())
 		})
 	})
--- a/umh-core/pkg/fsm/nmap/actions.go
+++ b/umh-core/pkg/fsm/nmap/actions.go
@@ -26,9 +26,16 @@ import (
 	nmap_service "github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/service/nmap"
 )
 
-// In Benthos, actions.go contained idempotent operations (like starting/stopping a service).
-// For the nmap monitor, we technically don't "start/stop" nmap itself—we're only
-// enabling or disabling the monitoring. We'll keep placeholder actions for consistency.
+// The functions in this file define heavier, possibly fail-prone operations
+// (for example, network or file I/O) that the Benthos FSM might need to perform.
+// They are intended to be called from Reconcile.
+//
+// IMPORTANT:
+//   - Each action is expected to be idempotent, since it may be retried
+//     multiple times due to transient failures.
+//   - Each action takes a context.Context and can return an error if the operation fails.
+//   - If an error occurs, the Reconcile function must handle
+//     setting error state and scheduling a retry/backoff.
 
 // CreateInstance is called when the FSM transitions from to_be_created -> creating.
 func (n *NmapInstance) CreateInstance(ctx context.Context, filesystemService filesystem.Service) error {
@@ -62,7 +69,7 @@ func (n *NmapInstance) RemoveInstance(ctx context.Context, filesystemService fil
 		metrics.ObserveReconcileTime(metrics.ComponentNmapInstance, n.baseFSMInstance.GetID()+".initiateS6Remove", time.Since(start))
 	}()
 
-	n.baseFSMInstance.GetLogger().Debugf("Starting Action: Removing S6 service %s ...", n.baseFSMInstance.GetID())
+	n.baseFSMInstance.GetLogger().Debugf("Starting Action: Removing Nmap service %s from S6 manager ...", n.baseFSMInstance.GetID())
 
 	// Initiate the removal cycle
 	err := n.monitorService.RemoveNmapFromS6Manager(ctx, n.baseFSMInstance.GetID())
@@ -80,7 +87,7 @@ func (n *NmapInstance) RemoveInstance(ctx context.Context, filesystemService fil
 
 // StartInstance attempts to start the Nmap by setting the desired state to running for the given instance
 func (n *NmapInstance) StartInstance(ctx context.Context, filesystemService filesystem.Service) error {
-	n.baseFSMInstance.GetLogger().Infof("Enabling monitoring for Nmap: %s (no-op)", n.baseFSMInstance.GetID())
+	n.baseFSMInstance.GetLogger().Infof("Starting Action: Starting Nmap service %s ...", n.baseFSMInstance.GetID())
 
 	err := n.monitorService.StartNmap(ctx, n.baseFSMInstance.GetID())
 	if err != nil {
@@ -94,7 +101,7 @@ func (n *NmapInstance) StartInstance(ctx context.Context, filesystemService file
 
 // StopInstance attempts to stop the Nmap by setting the desired state to stopped for the given instance
 func (n *NmapInstance) StopInstance(ctx context.Context, filesystemService filesystem.Service) error {
-	n.baseFSMInstance.GetLogger().Infof("Disabling monitoring for %s (no-op)", n.baseFSMInstance.GetID())
+	n.baseFSMInstance.GetLogger().Infof("Starting Action: Stopping Nmap service %s ...", n.baseFSMInstance.GetID())
 	// Set the desired state to stopped for the given instance
 	err := n.monitorService.StopNmap(ctx, n.baseFSMInstance.GetID())
 	if err != nil {
@@ -121,6 +128,10 @@ func (n *NmapInstance) UpdateObservedStateOfInstance(ctx context.Context, filesy
 
 	// Stores the raw service info
 	n.ObservedState.ServiceInfo = svcInfo
+
+	if n.ObservedState.LastStateChange == 0 {
+		n.ObservedState.LastStateChange = time.Now().Unix()
+	}
 
 	// Fetch the actual Nmap config from the service
 	start = time.Now()
--- a/umh-core/pkg/fsm/nmap/actions.go
+++ b/umh-core/pkg/fsm/nmap/actions.go
@@ -50,7 +50,7 @@ func (n *NmapInstance) CreateInstance(ctx context.Context, filesystemService fil
 
 	}
 
-	n.baseFSMInstance.GetLogger().Debugf("S6 service %s directory structure created", n.baseFSMInstance.GetID())
+	n.baseFSMInstance.GetLogger().Debugf("Nmap service %s added to S6 manager", n.baseFSMInstance.GetID())
 	return nil
 }
 
@@ -71,16 +71,16 @@ func (n *NmapInstance) RemoveInstance(ctx context.Context, filesystemService fil
 			n.baseFSMInstance.GetLogger().Debugf("Nmap service %s not found in S6 manager", n.baseFSMInstance.GetID())
 			return nil // do not throw an error, as each action is expected to be idempotent
 		}
-		return fmt.Errorf("failed to remove service for %s: %w", n.baseFSMInstance.GetID(), err)
+		return fmt.Errorf("failed to remove Nmap service %s from S6 manager: %w", n.baseFSMInstance.GetID(), err)
 	}
 
-	n.baseFSMInstance.GetLogger().Debugf("S6 service %s removed", n.baseFSMInstance.GetID())
+	n.baseFSMInstance.GetLogger().Debugf("Nmap service %s removed from S6 manager", n.baseFSMInstance.GetID())
 	return nil
 }
 
 // StartInstance attempts to start the Nmap by setting the desired state to running for the given instance
 func (n *NmapInstance) StartInstance(ctx context.Context, filesystemService filesystem.Service) error {
-	n.baseFSMInstance.GetLogger().Infof("Enabling monitoring for %s (no-op)", n.baseFSMInstance.GetID())
+	n.baseFSMInstance.GetLogger().Infof("Enabling monitoring for Nmap: %s (no-op)", n.baseFSMInstance.GetID())
 
 	err := n.monitorService.StartNmap(ctx, n.baseFSMInstance.GetID())
 	if err != nil {
--- a/umh-core/pkg/fsm/nmap/actions.go
+++ b/umh-core/pkg/fsm/nmap/actions.go
@@ -120,9 +120,21 @@ func (n *NmapInstance) UpdateObservedStateOfInstance(ctx context.Context, filesy
 		return ctx.Err()
 	}
 
+	currentState := n.baseFSMInstance.GetCurrentFSMState()
+	desiredState := n.baseFSMInstance.GetDesiredFSMState()
+	// If both desired and current state are stopped, we can return immediately
+	// There wont be any logs, metrics, etc. to check
+	if desiredState == OperationalStateStopped && currentState == OperationalStateStopped {
+		return nil
+	}
+
 	start := time.Now()
 	svcInfo, err := n.monitorService.Status(ctx, filesystemService, n.config.Name, tick)
 	if err != nil {
+		// We want to return this specific error here, because we need to check
+		// whether the execution of Nmap failed multiple times in a row and set this
+		// as a permanent Error. This also might happen, because we read the nmap
+		// metrics from logs, which can be inconsistent in some ticks.
 		if errors.Is(err, nmap_service.ErrScanFailed) {
 			return err
 		}
--- a/umh-core/pkg/fsm/nmap/actions.go
+++ b/umh-core/pkg/fsm/nmap/actions.go
@@ -37,21 +37,16 @@ func (n *NmapInstance) CreateInstance(ctx context.Context, filesystemService fil
 		metrics.ObserveReconcileTime(metrics.ComponentNmapInstance, n.baseFSMInstance.GetID()+".initiateS6Create", time.Since(start))
 	}()
 
-	n.baseFSMInstance.GetLogger().Debugf("Starting Action: Adding Benthos service %s to S6 manager ...", n.baseFSMInstance.GetID())
+	n.baseFSMInstance.GetLogger().Debugf("Starting Action: Adding Nmap service %s to S6 manager ...", n.baseFSMInstance.GetID())
 
-	// Check if we have a config with command or other settings
-	configEmpty := n.config.NmapServiceConfig.Port == 0 && n.config.NmapServiceConfig.Target == ""
-
-	if !configEmpty {
-		// Create service with custom configuration
-		err := n.monitorService.AddNmapToS6Manager(ctx, &n.config.NmapServiceConfig, n.baseFSMInstance.GetID())
-		if err != nil {
-			if err == nmap_service.ErrServiceAlreadyExists {
-				n.baseFSMInstance.GetLogger().Debugf("Nmap service %s already exists in S6 manager", n.baseFSMInstance.GetID())
-				return nil // do not throw an error, as each action is expected to be idempotent
-			}
-			return fmt.Errorf("failed to create service with config for %s: %w", n.baseFSMInstance.GetID(), err)
+	err := n.monitorService.AddNmapToS6Manager(ctx, &n.config.NmapServiceConfig, n.baseFSMInstance.GetID())
+	// Create service with custom configuration
+	if err != nil {
+		if err == nmap_service.ErrServiceAlreadyExists {
+			n.baseFSMInstance.GetLogger().Debugf("Nmap service %s already exists in S6 manager", n.baseFSMInstance.GetID())
+			return nil // do not throw an error, as each action is expected to be idempotent
 		}
+		return fmt.Errorf("failed to create service with config for %s: %w", n.baseFSMInstance.GetID(), err)
 
 	}
 
--- a/umh-core/pkg/fsm/nmap/actions.go
+++ b/umh-core/pkg/fsm/nmap/actions.go
@@ -16,6 +16,7 @@ package nmap
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"time"
@@ -122,6 +123,9 @@ func (n *NmapInstance) UpdateObservedStateOfInstance(ctx context.Context, filesy
 	start := time.Now()
 	svcInfo, err := n.monitorService.Status(ctx, filesystemService, n.config.Name, tick)
 	if err != nil {
+		if errors.Is(err, nmap_service.ErrScanFailed) {
+			return err
+		}
 		return fmt.Errorf("failed to get nmap metrics: %w", err)
 	}
 	metrics.ObserveReconcileTime(logger.ComponentNmapInstance, n.baseFSMInstance.GetID()+".getServiceStatus", time.Since(start))
--- a/umh-core/pkg/fsm/nmap/actions.go
+++ b/umh-core/pkg/fsm/nmap/actions.go
@@ -1,0 +1,170 @@
+// Copyright 2025 UMH Systems GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package nmap
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/logger"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/metrics"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/service/filesystem"
+	nmap_service "github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/service/nmap"
+)
+
+// In Benthos, actions.go contained idempotent operations (like starting/stopping a service).
+// For the nmap monitor, we technically don't "start/stop" nmap itself—we're only
+// enabling or disabling the monitoring. We'll keep placeholder actions for consistency.
+
+// CreateInstance is called when the FSM transitions from to_be_created -> creating.
+func (n *NmapInstance) CreateInstance(ctx context.Context, filesystemService filesystem.Service) error {
+	start := time.Now()
+	defer func() {
+		metrics.ObserveReconcileTime(metrics.ComponentNmapInstance, n.baseFSMInstance.GetID()+".initiateS6Create", time.Since(start))
+	}()
+
+	n.baseFSMInstance.GetLogger().Debugf("Starting Action: Adding Benthos service %s to S6 manager ...", n.baseFSMInstance.GetID())
+
+	// Check if we have a config with command or other settings
+	configEmpty := n.config.NmapServiceConfig.Port == 0 && n.config.NmapServiceConfig.Target == ""
+
+	if !configEmpty {
+		// Create service with custom configuration
+		err := n.monitorService.AddNmapToS6Manager(ctx, &n.config.NmapServiceConfig, n.baseFSMInstance.GetID())
+		if err != nil {
+			if err == nmap_service.ErrServiceAlreadyExists {
+				n.baseFSMInstance.GetLogger().Debugf("Nmap service %s already exists in S6 manager", n.baseFSMInstance.GetID())
+				return nil // do not throw an error, as each action is expected to be idempotent
+			}
+			return fmt.Errorf("failed to create service with config for %s: %w", n.baseFSMInstance.GetID(), err)
+		}
+
+	}
+
+	n.baseFSMInstance.GetLogger().Debugf("S6 service %s directory structure created", n.baseFSMInstance.GetID())
+	return nil
+}
+
+// RemoveInstance attempts to remove the Nmap service.
+// It requires the service to be stopped before removal.
+func (n *NmapInstance) RemoveInstance(ctx context.Context, filesystemService filesystem.Service) error {
+	start := time.Now()
+	defer func() {
+		metrics.ObserveReconcileTime(metrics.ComponentNmapInstance, n.baseFSMInstance.GetID()+".initiateS6Remove", time.Since(start))
+	}()
+
+	n.baseFSMInstance.GetLogger().Debugf("Starting Action: Removing S6 service %s ...", n.baseFSMInstance.GetID())
+
+	// Initiate the removal cycle
+	err := n.monitorService.RemoveNmapFromS6Manager(ctx, n.baseFSMInstance.GetID())
+	if err != nil {
+		if err == nmap_service.ErrServiceNotExist {
+			n.baseFSMInstance.GetLogger().Debugf("Nmap service %s not found in S6 manager", n.baseFSMInstance.GetID())
+			return nil // do not throw an error, as each action is expected to be idempotent
+		}
+		return fmt.Errorf("failed to remove service for %s: %w", n.baseFSMInstance.GetID(), err)
+	}
+
+	n.baseFSMInstance.GetLogger().Debugf("S6 service %s removed", n.baseFSMInstance.GetID())
+	return nil
+}
+
+// StartInstance attempts to start the Nmap by setting the desired state to running for the given instance
+func (n *NmapInstance) StartInstance(ctx context.Context, filesystemService filesystem.Service) error {
+	n.baseFSMInstance.GetLogger().Infof("Enabling monitoring for %s (no-op)", n.baseFSMInstance.GetID())
+
+	err := n.monitorService.StartNmap(ctx, n.baseFSMInstance.GetID())
+	if err != nil {
+		// if the service is not there yet but we attempt to start it, we need to throw an error
+		return fmt.Errorf("failed to start Nmap service %s: %w", n.baseFSMInstance.GetID(), err)
+	}
+
+	n.baseFSMInstance.GetLogger().Debugf("Nmap service %s start command executed", n.baseFSMInstance.GetID())
+	return nil
+}
+
+// StopInstance attempts to stop the Nmap by setting the desired state to stopped for the given instance
+func (n *NmapInstance) StopInstance(ctx context.Context, filesystemService filesystem.Service) error {
+	n.baseFSMInstance.GetLogger().Infof("Disabling monitoring for %s (no-op)", n.baseFSMInstance.GetID())
+	// Set the desired state to stopped for the given instance
+	err := n.monitorService.StopNmap(ctx, n.baseFSMInstance.GetID())
+	if err != nil {
+		// if the service is not there yet but we attempt to stop it, we need to throw an error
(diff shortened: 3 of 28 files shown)


umh-core/pkg/fsm/nmap/models.go
```diff
@@ -0,0 +1,86 @@
...
+		MonitoringStateOpen:
+		return true
+	}
+	return false
+}
+
+// NmapObservedState holds the last known nmap metrics and health status
+type NmapObservedState struct {
+	// We store the nmap data from nmap_monitor.GetStatus
+	NmapStatus *nmap_service.NmapStatus
```
JeremyTheocharis, line 70: rename NmapStatus to ServiceInfo

umh-core/pkg/fsm/nmap/models.go
```diff
@@ -0,0 +1,86 @@
...
+}
```
JeremyTheocharis, line 71: ObservedNmapServiceConfig

umh-core/pkg/fsm/nmap/models.go
```diff
@@ -0,0 +1,86 @@
...
+
+// Ensure it implements the ObservedState interface
+func (n NmapObservedState) IsObservedState() {}
+
+// NmapMonitorInstance implements fsm.FSMInstance
+// (See machine.go for the struct definition.)
+type NmapMonitorInstance interface {
```
JeremyTheocharis, line 78: adjust with var _ publicfsm.FSMInstance = (*ContainerInstance)(nil)

umh-core/pkg/fsm/nmap/machine.go
```diff
@@ -0,0 +1,205 @@
...
+	"github.com/looplab/fsm"
+	internal_fsm "github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/internal/fsm"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/constants"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/logger"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/metrics"
+	nmap_service "github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/service/nmap"
+)
+
+// NmapConfig is the portion of your FullConfig for container monitoring.
+type NmapConfig struct {
```
JeremyTheocharis, line 31: this belongs into config.go in the config package

umh-core/pkg/fsm/nmap/observed_state_snapshot.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"github.com/tiendc/go-deepcopy"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsm"
+	nmap_service "github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/service/nmap"
+)
+
+// NmapObservedStateSnapshot is a copy of the current metrics so the manager can store snapshots.
+type NmapObservedStateSnapshot struct {
+	// For example, copy the entire *models.Nmap if that's your data
+	NmapStatusSnapshot nmap_service.NmapStatus
```
JeremyTheocharis, line 26: add config

umh-core/pkg/fsm/nmap/observed_state_snapshot.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// Ensure it satisfies fsm.ObservedStateSnapshot
+func (n *NmapObservedStateSnapshot) IsObservedStateSnapshot() {}
+
+// CreateObservedStateSnapshot is called by the manager to record the state
+func (n *NmapInstance) CreateObservedStateSnapshot() fsm.ObservedStateSnapshot {
+	snapshot := &NmapObservedStateSnapshot{}
+	if n.ObservedState.NmapStatus != nil {
```
JeremyTheocharis, line 35: include also a deep copy of the config here

umh-core/pkg/fsm/nmap/reconcile.go
```diff
@@ -0,0 +1,297 @@
...
+			metrics.IncErrorCount(metrics.ComponentNmapMonitor, instanceName)
+		}
+	}()
+
+	// Check if context is already cancelled
+	if ctx.Err() != nil {
+		return ctx.Err(), false
+	}
+
+	// 2) Check if we should skip due to a recent error with backoff
```
JeremyTheocharis, line 50: // Step 1: If there's a lastError, see if we've waited enough.

umh-core/pkg/fsm/nmap/observed_state_snapshot.go
led0nk: image

from what i've found we should use the `ServiceInfo` instead of the NmapStatus here, because it contains `NmapStatus` and that's also used in s6

image

umh-core/pkg/fsm/nmap/actions.go
```diff
@@ -0,0 +1,115 @@
...
+	start := time.Now()
+	defer func() {
+		metrics.ObserveReconcileTime(metrics.ComponentS6Instance, n.baseFSMInstance.GetID()+".initiateS6Remove", time.Since(start))
+	}()
+
+	n.baseFSMInstance.GetLogger().Debugf("Starting Action: Removing S6 service %s ...", n.baseFSMInstance.GetID())
+
+	// First ensure the service is stopped
+	if n.ObservedState.ServiceInfo.NmapStatus.IsRunning {
+		return fmt.Errorf("service %s cannot be removed while running", n.baseFSMInstance.GetID())
```
JeremyTheocharis, line 69: it can right? what the S6 manager will do is it will set the desired state of nmap s6 service to stopped, and then remove it (or force remove it if it gets stuck somewhere)

umh-core/pkg/fsm/nmap/actions.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// monitoringStart is called when the nmap monitoring should be enabled.
+// Currently this is a no-op as the monitoring service runs independently.
+func (n *NmapInstance) monitoringStart(ctx context.Context) error {
+	n.baseFSMInstance.GetLogger().Infof("Enabling monitoring for %s (no-op)", n.baseFSMInstance.GetID())
+
+	err := n.monitorService.StartNmap(ctx, n.baseFSMInstance.GetID())
+	if err != nil {
+		// if the service is not there yet but we attempt to start it, we need to throw an error
```
JeremyTheocharis, line 89: you could catch this error by checking for the error type. then you could check for errors that are expected (and might be e.g., ignored) and unknown exceptiopns

umh-core/pkg/fsm/nmap/actions.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// monitoringStop is called when the nmap monitoring should be disabled.
+// Currently this is a no-op as the monitoring service runs independently.
+func (n *NmapInstance) monitoringStop(ctx context.Context) error {
+	n.baseFSMInstance.GetLogger().Infof("Disabling monitoring for %s (no-op)", n.baseFSMInstance.GetID())
+	// Set the desired state to stopped for the given instance
+	err := n.monitorService.StopNmap(ctx, n.baseFSMInstance.GetID())
+	if err != nil {
+		// if the service is not there yet but we attempt to stop it, we need to throw an error
```
JeremyTheocharis, line 104: see above

umh-core/pkg/fsm/nmap/manager.go
```diff
@@ -0,0 +1,127 @@
...
+		func(nn config.NmapConfig) (string, error) {
+			return nn.Name, nil
+		},
+		// Desired state from config
+		func(nn config.NmapConfig) (string, error) {
+			return nn.DesiredFSMState, nil
+		},
+		// Create instance
+		func(nn config.NmapConfig) (public_fsm.FSMInstance, error) {
+			// Typically create a new nmap_monitor.Service or reuse shared
```
JeremyTheocharis, line 75: this comment is probably AI generated?

umh-core/pkg/fsm/nmap/manager.go
```diff
@@ -0,0 +1,127 @@
...
+			return ni.config.NmapServiceConfig.Equal(nn.NmapServiceConfig), nil
+		},
+		// Set config if only small changes
+		func(instance public_fsm.FSMInstance, nn config.NmapConfig) error {
+			ni, ok := instance.(*NmapInstance)
+			if !ok {
+				return fmt.Errorf("instance is not a NmapInstance")
+			}
+			ni.config = nn
+			// Also update desired state so the FSM can adapt
```
JeremyTheocharis, line 96: this comment is not needed anymore right?

umh-core/pkg/fsm/nmap/nmap_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+		It("transitions from degraded to open when port state is open", func() {
+			// Simulate that the last scan reports the port state "open".
+			mockSvc.SetServicePortState("test-nmap", "open", 10.0)
+
+			// A reconcile call should detect the port state and trigger EventPortOpen.
+			err, did := inst.Reconcile(ctx, mockFS, 24)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(did).To(BeTrue())
+			Expect(inst.GetCurrentFSMState()).To(Equal(nmap.OperationalStateOpen))
```
JeremyTheocharis, line 143: add a check that it now remains in open

umh-core/pkg/fsm/nmap/nmap_test.go
```diff
@@ -0,0 +1,173 @@
...
+		It("transitions from degraded to closed when port state is closed", func() {
+			mockSvc.SetServicePortState("test-nmap", "closed", 30.0)
+			err, did := inst.Reconcile(ctx, mockFS, 26)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(did).To(BeTrue())
+			Expect(inst.GetCurrentFSMState()).To(Equal(nmap.OperationalStateClosed))
+		})
+
+		It("remains degraded if metrics are not healthy", func() {
+			// Force the mock service to simulate unhealthy metrics.
```
JeremyTheocharis, line 163: this comment is slightly confusing

umh-core/pkg/fsm/nmap/reconcile.go
```diff
@@ -0,0 +1,444 @@
...
+
+	// Step 1: If there's a lastError, see if we've waited enough.
+	if n.baseFSMInstance.ShouldSkipReconcileBecauseOfError(tick) {
+		backErr := n.baseFSMInstance.GetBackoffError(tick)
+		if backoff.IsPermanentFailureError(backErr) {
+			// If permanent, we want to remove the instance or at least stop it
+			// For now, let's just remove it from the manager:
+			if n.IsRemoved() || n.IsRemoving() || n.IsStopped() || n.IsStopping() {
+				n.baseFSMInstance.GetLogger().Errorf("Permanent error on nmap monitor %s but it is already in a terminal/removing state", instanceName)
+				err := n.monitorService.RemoveNmapFromS6Manager(ctx, n.config.Name)
```
JeremyTheocharis, line 67: this now should do a force remove

umh-core/pkg/fsm/nmap/reconcile.go
```diff
@@ -0,0 +1,444 @@
...
+			// already consumed significant time. We return reconciled=true to prevent
+			// further reconciliation attempts in the current tick.
+			return nil, true // We don't want to return an error here, as this can happen in normal operations
+		}
+
+		n.baseFSMInstance.SetError(err, tick)
+		n.baseFSMInstance.GetLogger().Errorf("error reconciling state: %s", err)
+		return nil, false // We don't want to return an error here, because we want to continue reconciling
+	}
+
```
JeremyTheocharis, line 118: don't we need to reconcile the underlying S6manager here?

umh-core/pkg/fsm/nmap/reconcile.go
```diff
@@ -0,0 +1,444 @@
...
+}
+
+// updateObservedState queries nmap_monitor.Service for new metrics
+func (n *NmapInstance) updateObservedState(ctx context.Context, filesystemService filesystem.Service, tick uint64) error {
+	svcInfo, err := n.monitorService.Status(ctx, filesystemService, n.config.Name, tick)
+	if err != nil {
+		return fmt.Errorf("failed to get nmap metrics: %w", err)
+	}
+	// Save to observed state
+	// TODO: remove debug-statement
```
JeremyTheocharis, line 197: :)

umh-core/pkg/fsm/nmap/reconcile.go
```diff
@@ -0,0 +1,444 @@
...
+		if err := n.monitoringStart(ctx); err != nil {
+			return err, false
+		}
+		// Send event to transition from Stopped to Starting
+		return n.baseFSMInstance.SendEvent(ctx, EventStart), true
+	case IsStartingState(currentState):
+		return n.reconcileStartingState(ctx, filesystemService, currentState, currentTime)
+	case IsRunningState(currentState):
+		return n.reconcileRunningState(ctx, filesystemService, currentState, currentTime)
+	}
```
JeremyTheocharis, line 277: maybe add here a default case that returns an error? jsut to double check if we have logic errors

umh-core/pkg/fsm/nmap/reconcile.go
```diff
@@ -0,0 +1,444 @@
...
+	}()
+
+	switch currentState {
+	case OperationalStateOpen:
+		// If we're in open, we need to check whether it is degraded
+		if !n.isNmapHealthy() {
+			return n.baseFSMInstance.SendEvent(ctx, EventNmapExecutionFailed), true
+		}
+
+		changedPortState, event := n.checkPortState(currentState)
```
JeremyTheocharis, line 341: add comment:

// if not degraded, check if the last reported state was open/filtered/closed and then set its state accordingly

umh-core/pkg/fsm/nmap/reconcile.go
```diff
@@ -0,0 +1,444 @@
...
+		if changedPortState {
+			return n.baseFSMInstance.SendEvent(ctx, event), true
+		}
+
+		return nil, false
+	case OperationalStateDegraded:
+		// If we're in Degraded, we need to recover with Nmap port state
+		// TODO: remove debugstatement
```
JeremyTheocharis, line 373: :)

umh-core/pkg/fsm/nmap/reconcile.go
```diff
@@ -0,0 +1,444 @@
...
+		return false
+	}
+
+	for _, log := range status.Logs {
+		if strings.Contains(log.Content, "error") {
+			return false
+		}
+	}
+	// Only consider nmap healthy if the overall health status is running
+	return status.IsRunning == true
```
JeremyTheocharis, line 407: isnt this double with 

```
if !n.IsNmapRunning() {
		return false
	}
```

umh-core/pkg/fsm/nmap/reconcile.go
```diff
@@ -0,0 +1,444 @@
...
+}
+
+// checkPortState checks on the port state and if it differs to the currentState
+// it returns true and the corresponding event
+func (n *NmapInstance) checkPortState(currentState string) (bool, string) {
+	if n.ObservedState.ServiceInfo.NmapStatus.LastScan == nil {
+
+		// TODO: remove debugstatement
+		n.baseFSMInstance.GetLogger().Warnf("before return ")
```
JeremyTheocharis, line 416: :)

umh-core/pkg/fsm/nmap/reconcile.go
```diff
@@ -0,0 +1,444 @@
...
+		return false, ""
+	}
+
+	portState := strings.ToLower(n.ObservedState.ServiceInfo.NmapStatus.LastScan.PortResult.State)
+
+	// TODO: remove debugstatement
+	n.baseFSMInstance.GetLogger().Warnf("before portState = %s", portState)
```
JeremyTheocharis, line 423: :)

umh-core/pkg/fsm/nmap/reconcile.go
```diff
@@ -0,0 +1,444 @@
...
+		return nil, false
+	case OperationalStateClosed:
+		// If we're in closed, we need to check whether it is degraded
+		if !n.isNmapHealthy() {
+			return n.baseFSMInstance.SendEvent(ctx, EventNmapExecutionFailed), true
+		}
+
+		changedPortState, event := n.checkPortState(currentState)
+		if changedPortState {
+			return n.baseFSMInstance.SendEvent(ctx, event), true
```
JeremyTheocharis, line 367: maybe as an idea: set reconciled to false if it switches between up/closed/filtered as there is not much happening here in the first place.

background: reconciled tells the control loop to stop executing at this point and wait for the next tick. it makes sense when we do "large actions" such as the creation of a new service etc. to give the system enough time to react, and it also makes sense for the starting, stopping, stopped, etc. phases for nmap, but maybe for up/filtered/closed it could make sense to keep reconciling (otherwise a single connection that went bad will now cause the entire loop to be skipped). could also do this for degraded state as well.

if we do it, we should definitely document that.

umh-core/pkg/fsm/nmap/actions.go
led0nk: yes make sense, since this also wasn't the correct `IsRunning` int the first place

umh-core/pkg/fsm/nmap/actions.go
led0nk: this might not make sense here since the only errors we're possibly receiving here are `context.DeadlineExceeded`, `context.Canceled` `ErrServiceNotExist` and `"s6 manager not initialized`, which are all relevant from my perspective and shouldn't get reported here different

umh-core/pkg/fsm/nmap/reconcile.go
led0nk: yup - just for my brain, so when we are not in a `currentState` that is included for the transition to running we always return an error

led0nk:
@JeremyTheocharis should we also move unittests to `umh-core/test/fsm/nmap` ?
I just copy/pasted from container-fsm but seemingly benthos and s6 are in those directories and we should keep it the same for all.

Scarjit (changes requested):

umh-core/pkg/fsm/nmap/actions.go
```diff
@@ -0,0 +1,110 @@
...
+// For the nmap monitor, we technically don't "start/stop" nmap itself—we're only
+// enabling or disabling the monitoring. We'll keep placeholder actions for consistency.
+
+// initiateNmapCreate is called when the FSM transitions from to_be_created -> creating.
+// For nmap monitoring, this is a no-op as there's no actual nmap to create.
+// This function is present for structural consistency with other FSM packages.
+func (n *NmapInstance) initiateNmapCreate(ctx context.Context, filesystemService filesystem.Service) error {
+	start := time.Now()
+	defer func() {
+		metrics.ObserveReconcileTime(metrics.ComponentS6Instance, n.baseFSMInstance.GetID()+".initiateS6Create", time.Since(start))
```
kanapuli, line 36: Shouldn't the first parameter be `metrics.ComponentNmapInstace` ? If yes , then it should be changed throughout this file

umh-core/pkg/fsm/nmap/actions.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to remove service for %s: %w", n.baseFSMInstance.GetID(), err)
+	}
+
+	n.baseFSMInstance.GetLogger().Debugf("S6 service %s removed", n.baseFSMInstance.GetID())
+	return nil
+}
+
+// monitoringStart is called when the nmap monitoring should be enabled.
+// Currently this is a no-op as the monitoring service runs independently.
```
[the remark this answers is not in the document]
led0nk: nope, comment removed

umh-core/pkg/fsm/nmap/actions.go
```diff
@@ -0,0 +1,110 @@
...
+		// if the service is not there yet but we attempt to start it, we need to throw an error
+		return fmt.Errorf("failed to start Nmap service %s: %w", n.baseFSMInstance.GetID(), err)
+	}
+
+	n.baseFSMInstance.GetLogger().Debugf("Nmap service %s start command executed", n.baseFSMInstance.GetID())
+	return nil
+}
+
+// monitoringStop is called when the nmap monitoring should be disabled.
+// Currently this is a no-op as the monitoring service runs independently.
```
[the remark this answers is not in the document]
led0nk: nope, comment removed

umh-core/pkg/fsm/nmap/actions.go
led0nk: ofc, good catch

umh-core/test/fsm/nmap/nmap_test.go
```diff
@@ -0,0 +1,209 @@
...
+		mockFS filesystem.Service
+	)
+
+	BeforeEach(func() {
+		ctx, cancel = context.WithCancel(context.Background())
+
+		// Create a mock Nmap service.
+		mockSvc = nmap_service.NewMockNmapService()
+		// Optionally, configure the mock service for healthy state:
+		// mockSvc.SetupMockForHealthyState()
```
JeremyTheocharis, line 50: why is this commented out? this is a little bit confusing

umh-core/pkg/control/loop_test.go
```diff
@@ -128,8 +128,8 @@ var _ = Describe("ControlLoop", func() {
 			loop := NewControlLoop(mockConfig)
 			Expect(loop).NotTo(BeNil())
 			Expect(loop.tickerTime).To(Equal(constants.DefaultTickerTime))
-			//Expect(loop.managers).To(HaveLen(4))
-			Expect(loop.managers).To(HaveLen(3))
+			//Expect(loop.managers).To(HaveLen(5))
```
JeremyTheocharis, line 131: this comment can be entirely removed

umh-core/pkg/fsm/nmap/reconcile.go
```diff
@@ -0,0 +1,502 @@
...
+
+// checkPortState checks on the port state and if it differs to the currentState
+// it returns true and the corresponding event
+func (n *NmapInstance) checkPortState(currentState string) (bool, string) {
+	if n.ObservedState.ServiceInfo.NmapStatus.LastScan == nil {
+
+		return false, ""
+	}
+
+	portState := strings.ToLower(n.ObservedState.ServiceInfo.NmapStatus.LastScan.PortResult.State)
```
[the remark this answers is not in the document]
led0nk: added

umh-core/pkg/fsm/nmap/reconcile.go
```diff
@@ -0,0 +1,502 @@
...
+		}
+		err = nil // The service does not exist, which is fine as this happens in the reconcileStateTransition}
+	}
+
+	// Step 3: Attempt to reconcile the state.
+	currentTime := time.Now()
+	err, reconciled = n.reconcileStateTransition(ctx, filesystemService, currentTime)
+	if err != nil {
+		// If the instance is removed, we don't want to return an error here, because we want to continue reconciling
+		// Also this should not
```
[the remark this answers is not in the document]
led0nk: nothing, comment is removed

umh-core/pkg/fsm/nmap/actions.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+// In Benthos, actions.go contained idempotent operations (like starting/stopping a service).
+// For the nmap monitor, we technically don't "start/stop" nmap itself—we're only
+// enabling or disabling the monitoring. We'll keep placeholder actions for consistency.
+
+// initiateNmapCreate is called when the FSM transitions from to_be_created -> creating.
+// For nmap monitoring, this is a no-op as there's no actual nmap to create.
+// This function is present for structural consistency with other FSM packages.
+func (n *NmapInstance) initiateNmapCreate(ctx context.Context, filesystemService filesystem.Service) error {
```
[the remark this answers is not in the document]
led0nk: adjusted the functions

umh-core/test/fsm/nmap/nmap_test.go
```diff
@@ -0,0 +1,209 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/config"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsm/nmap"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/service/filesystem"
+	nmap_service "github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/service/nmap"
+)
+
+var _ = Describe("Nmap FSM", func() {
```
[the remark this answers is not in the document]
led0nk: I added the missing nmap states and created a table-driven test for all state-transitions

umh-core/examples/example-config-nmap.yaml
```diff
@@ -0,0 +1,28 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+agent:
+  metricsPort: 8080
+
+redpanda:
```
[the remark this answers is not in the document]
led0nk: moved

umh-core/pkg/fsm/nmap/actions.go
```diff
@@ -0,0 +1,165 @@
...
+	"time"
+
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/logger"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/metrics"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/service/filesystem"
+	nmap_service "github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/service/nmap"
+)
+
+// In Benthos, actions.go contained idempotent operations (like starting/stopping a service).
+// For the nmap monitor, we technically don't "start/stop" nmap itself—we're only
```
JeremyTheocharis, line 30: i think this comment is valid for container, but not for nmap

JeremyTheocharis: maybe instead use thiscoment e..g, from benthos

```
// The functions in this file define heavier, possibly fail-prone operations
// (for example, network or file I/O) that the Benthos FSM might need to perform.
// They are intended to be called from Reconcile.
//
// IMPORTANT:
//   - Each action is expected to be idempotent, since it may be retried
//     multiple times due to transient failures.
//   - Each action takes a context.Context and can return an error if the operation fails.
//   - If an error occurs, the Reconcile function must handle
//     setting error state and scheduling a retry/backoff.
```

umh-core/pkg/fsm/nmap/manager.go
```diff
@@ -0,0 +1,120 @@
...
+func (n *NmapManagerSnapshot) IsObservedStateSnapshot() {}
+
+// NewNmapManager constructs a manager.
+// You might keep `sharedMonitorService` if you want one global service instance.
+func NewNmapManager(name string) *NmapManager {
+	managerName := fmt.Sprintf("%s_%s", logger.ComponentNmapManager, name)
+
+	baseMgr := public_fsm.NewBaseFSMManager[config.NmapConfig](
+		managerName,
+		"/dev/null", // no actual S6 base dir needed for a pure monitor
```
JeremyTheocharis, line 48: are we sure here?

umh-core/pkg/fsm/nmap/models.go
```diff
@@ -0,0 +1,164 @@
...
+	OperationalStateStopped = "stopped"
+
+	// Starting phase states
+	// starting is the operational state when nmap is starting
+	OperationalStateStarting = "starting"
+	// degraded means nmap is running, but execution is not OK
+	OperationalStateDegraded = "degraded"
+
+	// Running phase states
+	// active means nmap is running and it shows port filtered
```
JeremyTheocharis, line 37: maybe leave a comment to a link explaining all of these states

umh-core/pkg/fsm/nmap/models.go
```diff
@@ -0,0 +1,164 @@
...
+	// desired state, removal, etc.
+	baseFSMInstance *internal_fsm.BaseFSMInstance
+
+	// ObservedState: last known container metrics, updated in reconcile
+	ObservedState NmapObservedState
+
+	// The nmap monitor service used to gather metrics
+	monitorService nmap_service.INmapService
+
+	// Possibly store config needed for the container monitor
```
JeremyTheocharis, line 146: adjust comment

umh-core/pkg/fsm/nmap/models.go
```diff
@@ -0,0 +1,164 @@
...
+// Verify at compile time that we implement fsm.FSMInstance
+var _ publicfsm.FSMInstance = (*NmapInstance)(nil)
+
+// NmapInstance holds the FSM instance and references to the container monitor service.
+type NmapInstance struct {
+	// This embeds the "BaseFSMInstance" which handles lifecycle states,
+	// desired state, removal, etc.
+	baseFSMInstance *internal_fsm.BaseFSMInstance
+
+	// ObservedState: last known container metrics, updated in reconcile
```
JeremyTheocharis, line 140: last known nmap metrics

umh-core/pkg/fsm/nmap/reconcile.go
```diff
@@ -0,0 +1,503 @@
...
+		}
+		err = nil // The service does not exist, which is fine as this happens in the reconcileStateTransition}
+	}
+
+	// Step 3: Attempt to reconcile the state.
+	currentTime := time.Now()
+	err, reconciled = n.reconcileStateTransition(ctx, filesystemService, currentTime)
+	if err != nil {
+		// If the instance is removed, we don't want to return an error here, because we want to continue reconciling
+		if errors.Is(err, fsm.ErrInstanceRemoved) {
```
JeremyTheocharis, line 100: the logic here is different from benthos,

in benthos we also check for ErrServiceNotExist and don't check for ErrInstanceRemoved

maybe it is correct like this, i did not thought about it yet

umh-core/pkg/fsm/nmap/reconcile.go
```diff
@@ -0,0 +1,503 @@
...
+
+// isNmapHealthy decides if the nmap health is Active
+func (n *NmapInstance) isNmapHealthy() bool {
+	status := n.ObservedState.ServiceInfo.NmapStatus
+
+	if status.LastScan == nil && len(status.Logs) == 0 {
+		return false
+	}
+
+	for _, log := range status.Logs {
```
JeremyTheocharis, line 450: i just realized that the service with parsing the logs to get the the results here, probably suffers from the same logrotation issues that also @Scarjit had in his redpanda monitor PR. maybe need to add this exception as well to the nmap service code

JeremyTheocharis (changes requested):

umh-core/pkg/fsm/nmap/reconcile.go
led0nk: https://github.com/united-manufacturing-hub/united-manufacturing-hub/blob/5f3b198171dc6823df388fdd30a370c9e2977f26/umh-core/pkg/fsm/benthos/reconcile.go#L105-L125

No it's basically the same, or am i missing something here?

umh-core/pkg/fsm/nmap/manager.go
led0nk: no thanks for pointing this out, i think it makes sense to set this accordingly

umh-core/pkg/fsm/nmap/reconcile.go
JeremyTheocharis: i think the issue is more on the nmap service side:
1. S6 logs are always stored ina file current
2. they are logrotated once they reach 1 MB
3. now it can happen that half of them are in the new current file, and the other half in a logrotated archive

--> in this case, we should skip the reocncile and wait for the full next config again, except when we did not get anything reported for const time, then it should go into nmap failure state

umh-core/pkg/fsm/nmap/reconcile.go
JeremyTheocharis: you are right, somehow i messes up the lines

umh-core/pkg/fsm/nmap/actions.go
```diff
@@ -0,0 +1,180 @@
...
+	metrics.ObserveReconcileTime(logger.ComponentNmapInstance, n.baseFSMInstance.GetID()+".getConfig", time.Since(start))
+	if err == nil {
+		// Only update if we successfully got the config
+		n.ObservedState.ObservedNmapServiceConfig = observedConfig
+	} else {
+		if strings.Contains(err.Error(), nmap_service.ErrServiceNotExist.Error()) {
+			// Log the error but don't fail - this might happen during creation when the config file doesn't exist yet
+			n.baseFSMInstance.GetLogger().Debugf("Service not found, will be created during reconciliation: %v", err)
+			return nil
+		} else {
```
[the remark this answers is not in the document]
led0nk: no this makes sense here, since the error might be wrapped

umh-core/pkg/fsm/nmap/reconcile.go
led0nk: as we discussed today i therefore implemented some kind of backoff-functionality when the `ServiceInfo` from the current Status of the `monitorService` throws an error, which is ErrScanFailed we throw this error in the reconcile loop. Therefore we needed to adjust the `BackoffConfig` So i created an additional `func NewBackoffConfig` which we can then put into the `NewBaseFSMInstance` so we can handle our:
- `InitialInterval`
- `MaxInterval`
- `MaxRetries`
according to our needs. The default is set to 5, which would result in a permanentFailure for Nmap, if it will work in the 6th tick.

In this example it would result in adjusting the MaxRetries to e.g. 20 => max amount of 20 ticks (each 100ms) and Nmap executes every second. So in one each 10 Ticks there should exist full Nmap logs -> and if there is some coincidence happening, we get a second retry in the next 10 ticks.

led0nk: changed that to
- `initialInterval` = 10
- `MaxInterval` = 20
- `MaxRetries` = 2

therefore it should have the following pattern:
-> error occurs
-> wait 10 ticks for next registration of error
-> error occurs
-> wait 20 ticks for next registration of error
-> error occurs -> permanent failure

but instead of waiting 20 ticks, which would make sense due to exponential backoff, it just waits 15 ticks.

JeremyTheocharis (changes requested):

umh-core/pkg/fsm/nmap/reconcile.go
```diff
@@ -0,0 +1,522 @@
...
+	currentTime := time.Now()
+	err, reconciled = n.reconcileStateTransition(ctx, filesystemService, currentTime)
+	if err != nil {
+		// If the instance is removed, we don't want to return an error here, because we want to continue reconciling
+		if errors.Is(err, fsm.ErrInstanceRemoved) {
+			return nil, false
+		}
+
+		if errors.Is(err, context.DeadlineExceeded) {
+			// Updating the observed state can sometimes take longer,
```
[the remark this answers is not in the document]
led0nk: wrongly copy/pasted

umh-core/pkg/fsm/nmap/reconcile.go
```diff
@@ -0,0 +1,522 @@
...
+				n.baseFSMInstance.ResetState() // clear the error
+				err = n.Remove(ctx)            // attempt removal
+				if err != nil {
+					// If removing doesn't work because the fsm is not in the OperationalStateBeforeRemove
+					// we will force it to remove.
+					n.baseFSMInstance.GetLogger().Errorf("error removing Nmap instance %s: %v", instanceName, err)
+					forceErr := n.monitorService.ForceRemoveNmap(ctx, filesystemService, instanceName)
+					if forceErr != nil {
+						n.baseFSMInstance.GetLogger().Errorf("error force removing Nmap instance %s: %v", instanceName, forceErr)
+						return forceErr, false
```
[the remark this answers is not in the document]
led0nk: according to your comment from slack i adjusted this and wrapped the PermanentFailureError and forceErr -> then it will get detected by the baseFSM_manager

umh-core/pkg/fsm/nmap/actions.go
```diff
@@ -0,0 +1,188 @@
...
+	// If both desired and current state are stopped, we can return immediately
+	// There wont be any logs, metrics, etc. to check
+	if desiredState == OperationalStateStopped && currentState == OperationalStateStopped {
+		return nil
+	}
+
+	start := time.Now()
+	svcInfo, err := n.monitorService.Status(ctx, filesystemService, n.config.Name, tick)
+	if err != nil {
+		if errors.Is(err, nmap_service.ErrScanFailed) {
```
JeremyTheocharis, line 134: add a comment here for what this is

umh-core/pkg/fsm/nmap/actions.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+		return fmt.Errorf("failed to get nmap metrics: %w", err)
+	}
+	metrics.ObserveReconcileTime(logger.ComponentNmapInstance, n.baseFSMInstance.GetID()+".getServiceStatus", time.Since(start))
+
+	// Stores the raw service info
+	n.ObservedState.ServiceInfo = svcInfo
+
+	if n.ObservedState.LastStateChange == 0 {
+		n.ObservedState.LastStateChange = time.Now().Unix()
```
JeremyTheocharis, line 145: this is copied from S6, but it should be ticks instead, can you create a linear issue to fix it here and in S6, and assign it to me?

JeremyTheocharis: (it should also be ticks in S6)

JeremyTheocharis (approved):

umh-core/pkg/fsm/nmap/actions.go
led0nk: created

kanapuli (approved):

Scarjit (approved):

umh-core/pkg/fsm/nmap/actions.go
```diff
@@ -0,0 +1,192 @@
...
+	}
```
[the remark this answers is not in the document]
led0nk: already recognized and addressed in a linear-ticket

PR #1921, opened 2025-04-09
[ENG-2806] Implement DFC FSM Reconcile logic (WIP)

kanapuli:
## Description

Implement the Reconcile logic for DFC FSM. Needs self review one more time

### DIFF
--- a/umh-core/pkg/fsm/dataflowcomponent/actions.go
+++ b/umh-core/pkg/fsm/dataflowcomponent/actions.go
@@ -1,0 +1,251 @@
+// Copyright 2025 UMH Systems GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dataflowcomponent
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+
+	internalfsm "github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/internal/fsm"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/config/dataflowcomponentconfig"
+	benthosfsm "github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsm/benthos"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/logger"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/metrics"
+	dataflowcomponentservice "github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/service/dataflowcomponent"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/service/filesystem"
+)
+
+// The functions in this file define heavier, possibly fail-prone operations
+// (for example, network or file I/O) that the DataflowComponent's Benthos FSM might need to perform.
+// They are intended to be called from Reconcile.
+//
+// IMPORTANT:
+//   - Each action is expected to be idempotent, since it may be retried
+//     multiple times due to transient failures.
+//   - Each action takes a context.Context and can return an error if the operation fails.
+//   - If an error occurs, the Reconcile function must handle
+//     setting error state and scheduling a retry/backoff.
+
+// initiateDataflowComponentCreate attempts to add the DataflowComponent to the Benthos manager.
+func (d *DataflowComponentInstance) initiateDataflowComponentCreate(ctx context.Context, filesystemService filesystem.Service) error {
+	d.baseFSMInstance.GetLogger().Debugf("Starting Action: Adding DataflowComponent service %s to Benthos manager ...", d.baseFSMInstance.GetID())
+
+	err := d.service.AddDataFlowComponentToBenthosManager(ctx, filesystemService, &d.config, d.baseFSMInstance.GetID())
+	if err != nil {
+		if err == dataflowcomponentservice.ErrServiceAlreadyExists {
+			d.baseFSMInstance.GetLogger().Debugf("DataflowComponent service %s already exists in Benthos manager", d.baseFSMInstance.GetID())
+			return nil // do not throw an error, as each action is expected to be idempotent
+		}
+		return fmt.Errorf("failed to add DataflowComponent service %s to Benthos manager: %w", d.baseFSMInstance.GetID(), err)
+	}
+
+	d.baseFSMInstance.GetLogger().Debugf("DataflowComponent service %s added to Benthos manager", d.baseFSMInstance.GetID())
+	return nil
+}
+
+// initiateDataflowComponentRemove attempts to remove the DataflowComponent from the Benthos manager.
+// It requires the service to be stopped before removal.
+func (b *DataflowComponentInstance) initiateDataflowComponentRemove(ctx context.Context, filesystemService filesystem.Service) error {
+	b.baseFSMInstance.GetLogger().Debugf("Starting Action: Removing DataflowComponent service %s from Benthos manager ...", b.baseFSMInstance.GetID())
+
+	// Remove the Benthos from the S6 manager
+	err := b.service.RemoveDataFlowComponentFromBenthosManager(ctx, filesystemService, b.baseFSMInstance.GetID())
+	if err != nil {
+		if err == dataflowcomponentservice.ErrServiceNotExist {
+			b.baseFSMInstance.GetLogger().Debugf("DataflowComponent service %s not found in Benthos manager", b.baseFSMInstance.GetID())
+			return nil // do not throw an error, as each action is expected to be idempotent
+		}
+		return fmt.Errorf("failed to remove DataflowComponent service %s from Benthos manager: %w", b.baseFSMInstance.GetID(), err)
+	}
+
+	b.baseFSMInstance.GetLogger().Debugf("DataflowComponent service %s removed from Benthos manager", b.baseFSMInstance.GetID())
+	return nil
+}
+
+// initiateDataflowComponentStart to start the DataflowComponent by setting the desired state to running for the given instance
+func (d *DataflowComponentInstance) initiateDataflowComponentStart(ctx context.Context, filesystemService filesystem.Service) error {
+	d.baseFSMInstance.GetLogger().Debugf("Starting Action: Starting DataflowComponent service %s ...", d.baseFSMInstance.GetID())
+
+	// TODO: Add pre-start validation
+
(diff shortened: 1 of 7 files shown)


umh-core/pkg/fsm/dataflowcomponent/actions.go
```diff
@@ -0,0 +1,251 @@
...
+			d.baseFSMInstance.GetLogger().Debugf("Config differences detected but service does not exist yet, skipping update")
+		}
+	}
+
+	return nil
+}
+
+// getServiceStatus gets the status of the DataflowComponent service
+// its main purpose is to handle the edge cases where the service is not yet created or not yet running
+func (d *DataflowComponentInstance) getServiceStatus(ctx context.Context, filesystemService filesystem.Service, tick uint64) (dataflowcomponentservice.ServiceInfo, error) {
```
led0nk, line 169: although this sounds stupid, can we move this above `updateObservedState` ? (consistent with benthos-fsm)

umh-core/pkg/service/dataflowcomponent/dataflowcomponent.go
```diff
@@ -165,17 +165,17 @@ func (s *DataFlowComponentService) GenerateBenthosConfigForDataFlowComponent(dat
 
 // GetConfig returns the actual DataFlowComponent config from the Benthos service
 // Expects benthosName (e.g. "dataflow-myservice") as defined in the UMH config
-func (s *DataFlowComponentService) GetConfig(ctx context.Context, filesystemService filesystem.Service, componentName string) (dataflowcomponentconfig.DataFlowComponentConfig, error) {
+func (s *DataFlowComponentService) GetConfig(ctx context.Context, filesystemService filesystem.Service, componentName string) (*dataflowcomponentconfig.DataFlowComponentConfig, error) {
 	if ctx.Err() != nil {
-		return dataflowcomponentconfig.DataFlowComponentConfig{}, ctx.Err()
+		return nil, ctx.Err()
```
led0nk, line 170: why are we returning `nil` here instead of an empty config?

umh-core/pkg/service/dataflowcomponent/dataflowcomponent.go
```diff
@@ -165,17 +165,17 @@ func (s *DataFlowComponentService) GenerateBenthosConfigForDataFlowComponent(dat
...
 	}
 
 	benthosName := s.getBenthosName(componentName)
 
 	// Get the Benthos config
 	benthosCfg, err := s.benthosService.GetConfig(ctx, filesystemService, benthosName)
 	if err != nil {
-		return dataflowcomponentconfig.DataFlowComponentConfig{}, fmt.Errorf("failed to get benthos config: %w", err)
+		return nil, fmt.Errorf("failed to get benthos config: %w", err)
```
led0nk, line 178: as above

umh-core/pkg/service/dataflowcomponent/mock.go
```diff
@@ -121,16 +121,16 @@ func (m *MockDataFlowComponentService) GenerateBenthosConfigForDataFlowComponent
 }
 
 // GetConfig mocks getting the DataFlowComponent configuration
-func (m *MockDataFlowComponentService) GetConfig(ctx context.Context, filesystemService filesystem.Service, componentName string) (dataflowcomponentconfig.DataFlowComponentConfig, error) {
+func (m *MockDataFlowComponentService) GetConfig(ctx context.Context, filesystemService filesystem.Service, componentName string) (*dataflowcomponentconfig.DataFlowComponentConfig, error) {
```
led0nk, line 124: why are we using a pointer here? We should adjust the interface to return the config instead of a pointer (also consistency with other fsms)

umh-core/pkg/fsm/dataflowcomponent/reconcile.go
```diff
@@ -29,6 +38,343 @@ import (
...
+		}
+
+		return d.baseFSMInstance.SendEvent(ctx, EventStartDone), true
+	default:
+		return fmt.Errorf("invalid starting state: %s", currentState), false
+	}
+}
+
+// reconcileRunningState handles the various running states when transitioning to Active.
+func (d *DataflowComponentInstance) reconcileRunningState(ctx context.Context, filesystemService filesystem.Service, currentState string, currentTime time.Time) (err error, reconciled bool) {
```
led0nk, line 321: @JeremyTheocharis We should definitely decide on either renaming this to `reconcileActiveState` for all fsms - or rename the func `reconcileTransitionToActive` to `reconcileTransitionToRunning`.

imo `reconcileRunningState` is reconciling all running states 
`reconcileTransitionToActive` is reconciling the transition to any state considered as `running` but desired here is active


`// reconcileTransitionToActive handles transitions when the desired state is Active.`
It makes sense in this specific case, but across all fsms we then would have to call this after the desired state, which would pretty inconsistent.

led0nk (commented):
sorry for being picky here, i think we just have to be precise throughout the fsms, even though we might adjust some things just later

umh-core/pkg/fsm/dataflowcomponent/actions.go
```diff
@@ -0,0 +1,251 @@
...
+			d.baseFSMInstance.GetLogger().Debugf("Service not found, will be created during reconciliation: %v", err)
+			return nil
+		} else {
+			return fmt.Errorf("failed to get observed DataflowComponent config: %w", err)
+		}
+	}
+
+	// Detect a config change - but let the Benthos manager handle the actual reconciliation
+	// Use new ConfigsEqual function that handles Benthos defaults properly
+	if !dataflowcomponentconfig.CompareConfigs(&d.config, d.ObservedState.ObservedDataflowComponentConfig) {
```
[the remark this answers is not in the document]
kanapuli: Cool. Names Changed

umh-core/pkg/fsm/dataflowcomponent/actions.go
kanapuli: Done

umh-core/pkg/service/dataflowcomponent/dataflowcomponent.go
kanapuli: Commented below the reason for returning a pointer

umh-core/pkg/service/dataflowcomponent/mock.go
led0nk: > DataflowComponentConfig could be huge (depends on the underlying benthos config) and it is better to pass it as pointer instead of copying the struct to every function and wasting memory.

I totally get this point here, but since it depends on the underlying benthos-config it's basically the same for benthos, I think we should just keep it the same for all fsms. Maybe we need to adjust this on the other fsms

wdyt @Scarjit  @JeremyTheocharis ?

umh-core/pkg/fsm/dataflowcomponent/actions.go
```diff
@@ -0,0 +1,251 @@
...
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
+
+	start := time.Now()
+	info, err := d.getServiceStatus(ctx, filesystemService, tick)
+	if err != nil {
+		return fmt.Errorf("error while getting service status: %w", err)
+	}
+	metrics.ObserveReconcileTime(logger.ComponentBenthosInstance, d.baseFSMInstance.GetID()+".getServiceStatus", time.Since(start))
```
led0nk, line 152: this should also be `logger.ComponentDataFlowComponentInstance`

umh-core/pkg/fsm/dataflowcomponent/actions.go
```diff
@@ -0,0 +1,251 @@
...
+			return fmt.Errorf("failed to get observed DataflowComponent config: %w", err)
+		}
+	}
+
+	// Detect a config change - but let the Benthos manager handle the actual reconciliation
+	// Use new ConfigsEqual function that handles Benthos defaults properly
+	if !dataflowcomponentconfig.ConfigsEqual(&d.config, d.ObservedState.ObservedDataflowComponentConfig) {
+		// Check if the service exists before attempting to update
+		if d.service.ServiceExists(ctx, filesystemService, d.baseFSMInstance.GetID()) {
+			d.baseFSMInstance.GetLogger().Debugf("Observed DataflowComponent config is different from desired config, updating DataflowComponent configuration")
```
led0nk, line 178: updating `Benthos` configuration

led0nk (changes requested):

umh-core/pkg/fsm/dataflowcomponent/actions.go
kanapuli: Thanks fixed

led0nk (approved):

umh-core/pkg/service/dataflowcomponent/mock.go
Scarjit: I think ptrs are the better solution here, but that would as you say need an update for the other fsm.

umh-core/pkg/fsm/dataflowcomponent/reconcile.go
Scarjit: It might make sense to merge it now down, and then do a consolidated cleanup of all fsm tomorrow, so that we only have to modify the codebase once

umh-core/pkg/service/dataflowcomponent/mock.go
JeremyTheocharis: then lets keep it consistent for now

umh-core/pkg/fsm/dataflowcomponent/reconcile.go
JeremyTheocharis: I already spent an hour going through all of that and we also already discussed this in person @led0nk, and found out that the current approach is the most easily understood. 

We already discussed it in this PR here: https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/1907

A small consitency fix here is to rename this to `reconcileRunningStateS`

Background: 
image
image

we can see here, that reconcileOperationalStates reconciles everything if the current state is an operational one. From then one, it switches the view: instead of considering what the current state is, the next distinction is made what the desiredState is. Therefore, it makes sense to call those functions `reconcileTransitionToActive` (as it has everything that is needed to reconcile to Active)

then in there, we again check based on the current state, and therefore it makes sense to call it `reconcileStartingStates`

image


We could then also switch it, but then it becomes slightly unclear at different points in the codebase. Therefore, I recommend only doing this minor change here as keep it consistent across all FSMs.

umh-core/pkg/fsm/dataflowcomponent/actions.go
```diff
@@ -16,46 +16,236 @@ package dataflowcomponent
...
+//
+// IMPORTANT:
+//   - Each action is expected to be idempotent, since it may be retried
+//     multiple times due to transient failures.
+//   - Each action takes a context.Context and can return an error if the operation fails.
+//   - If an error occurs, the Reconcile function must handle
+//     setting error state and scheduling a retry/backoff.
+
+// initiateDataflowComponentCreate attempts to add the DataflowComponent to the Benthos manager.
+func (d *DataflowComponentInstance) initiateDataflowComponentCreate(ctx context.Context, filesystemService filesystem.Service) error {
```
JeremyTheocharis, line 45: you would need to implement the FSMActionsInterface here

JeremyTheocharis (changes requested):
I find this difficult to review if most of the reconcile logic will change anyway, let me wait for review until you have finished that

kanapuli:
This PR has got obsolete code. Closing this in favour of https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/1928.

PR #1963, opened 2025-04-24
[ENG-2858] Use ServiceRegistry instead of filesystem.Service

kanapuli:
## Summary by CodeRabbit

- **Refactor**
  - Unified service access across the application by introducing a centralized service registry abstraction.
  - Updated all manager, instance, and service methods to use the service registry interface for accessing core services instead of direct dependencies.
  - Adjusted method signatures and internal logic to support the new service registry approach, improving modularity and maintainability.

- **Tests**
  - Refactored all test suites to use a mock service registry for dependency injection, replacing previous direct service mocks.
  - No changes to test logic or assertions; tests continue to validate all workflows as before.

- **New Features**
  - Added a centralized service registry, including singleton and mock implementations, to streamline access to core services.


FreiPaul (approved):

umh-core/pkg/control/loop.go
```diff
@@ -281,7 +285,7 @@ func (c *ControlLoop) Reconcile(ctx context.Context, ticker uint64) error {
 	newSnapshot.CurrentConfig = cfg
 
 	// 4) If your filesystem service is a buffered FS, sync once per loop:
-	bufferedFs, ok := c.filesystemService.(*filesystem.BufferedService)
+	bufferedFs, ok := c.services.GetFileSystem().(*filesystem.BufferedService)
```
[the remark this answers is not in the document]
led0nk: this might make sense now

led0nk (changes requested):

led0nk (approved):
looks good from my perspective even though i'm unsure why we need to panic in factory. But looking forward to address this in my pr

Scarjit (approved):

kanapuli:
> looks good from my perspective even though i'm unsure why we need to panic in factory. But looking forward to address this in my pr

@led0nk I understand your thoughts. Usually I am against using panic but I feel panic makes more sense for this internal package. It forces that the other parts of the code should always have the serviceregistry through dependency injection and any violation on this would panic which will be caught in tests before moving to production. So I would say it is safe.

umh-core/pkg/fsm/dataflowcomponent/machine.go
```diff
@@ -26,13 +26,16 @@ import (
...
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/portmanager"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/service/dataflowcomponent"
 )
 
 // NewDataflowComponentInstance creates a new DataflowComponentInstance with a given ID and service path
 func NewDataflowComponentInstance(
 	s6BaseDir string,
-	config config.DataFlowComponentConfig) *DataflowComponentInstance {
+	config config.DataFlowComponentConfig,
+	portManager portmanager.PortManager,
```
JeremyTheocharis, line 37: for what is the port manager needed here?

JeremyTheocharis (changes requested):

umh-core/pkg/fsm/dataflowcomponent/machine.go
kanapuli: This is an unused parameter. Removed it and cleaned up the tests

JeremyTheocharis (approved):

PR #1972, opened 2025-04-25
[ENG-2913] System State Mocking

FreiPaul:
This PR introduces a stateMocker that provides a mocked system state manager for the unit tests (of the actions). It runs in a separate goroutine and checks the config of the config manager, it is given and derives a mocked system state from it that is then accessible via a stateManager. Thereby, the execute steps of actions and the waitForDfcToBecomeActive etc. can be tested. It also features detection of adding/removing/editing a dfc and performing state transitions close to how the real system would do

## Summary by CodeRabbit

- **New Features**
  - Introduced a mock system state simulator for unit testing, enabling dynamic simulation and management of component states.
  - Added helper functions for consuming outbound messages in tests.
  - Extended action handling to include system snapshot management for edit instance actions.

- **Bug Fixes**
  - Standardized ticker intervals for component actions using a new constant for improved consistency.

- **Tests**
  - Enhanced and refactored test suites to use the new state simulation tools, improving reliability and maintainability.
  - Added comprehensive tests for the new state simulation features.

- **Chores**
  - Updated example configuration to set the internal Redpanda component to an active state by default.
  - Added new debug configurations for Go test debugging in the development environment.


.DS_Store
Scarjit, line 1: Let's not push .DS_Store files into the repo

umh-core/pkg/communicator/actions/delete-dataflowcomponent_test.go
```diff
@@ -152,12 +159,18 @@ var _ = Describe("DeleteDataflowComponent", func() {
 			// Reset tracking for this test
 			mockConfig.ResetCalls()
 
+			// Start the state mocker
+			stateMocker.Start()
```
Scarjit, line 163: Is there a reason why these are not called in BeforeEach/AfterEach ?

umh-core/pkg/communicator/actions/delete-dataflowcomponent_test.go
```diff
@@ -152,12 +159,18 @@ var _ = Describe("DeleteDataflowComponent", func() {
...
+
 			// Execute the action
 			result, metadata, err := action.Execute()
 			Expect(err).NotTo(HaveOccurred())
 			Expect(result).To(ContainSubstring("Successfully deleted data flow component"))
 			Expect(metadata).To(BeNil())
 
+			// Stop the state mocker
+			stateMocker.Stop()
```
Scarjit, line 172: Is there a reason why these are not called in BeforeEach/AfterEach ?

umh-core/pkg/communicator/actions/stateMocker.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+// Tick advances the state mocker by one tick and updates the system state
+func (s *StateMocker) Tick() {
+	s.TickCounter++
+	s.UpdateState()
+}
+
+// here, the actual state update logic is implemented
+func (s *StateMocker) UpdateState() {
```
Scarjit, line 83: Since this only affects the DFC's, maybe a better name like UpdateDFCState ?

umh-core/pkg/communicator/actions/stateMocker.go
```diff
@@ -0,0 +1,252 @@
...
+	if s.State.Managers == nil {
+		s.State.Managers = make(map[string]fsm.ManagerSnapshot)
+	}
+	s.State.Managers[constants.DataflowcomponentManagerName] = managerSnapshot
+}
+
+// Start spawns a goroutine that updates the state of the system periodically
+// It returns immediately and can be stopped by calling Stop()
+func (s *StateMocker) Run() {
+	ticker := time.NewTicker(1 * time.Second)
```
Scarjit, line 147: This tick rate is different from our normal control loop (1000 vs 100ms)

umh-core/pkg/communicator/actions/stateMocker.go
```diff
@@ -0,0 +1,252 @@
...
+			case <-s.done:
+				return
+			}
+		}
+	}()
+}
+
+// Start is a convenience method that runs the state mocker in a new goroutine
+func (s *StateMocker) Start() {
+	go s.Run()
```
Scarjit, line 165: For safety it could make sense to throw an error if this is started twice without stopping first.

Scarjit (changes requested):

umh-core/pkg/communicator/actions/stateMocker.go
```diff
@@ -0,0 +1,283 @@
...
+	return nil
+}
+
+// Stop signals the state mocker to stop updating and waits for it to complete
+// It is safe to call Stop even if the mocker isn't running
+func (s *StateMocker) Stop() {
+	if s.running.Load() {
+		close(s.done)
+		// Create a new channel for next run
+		s.done = make(chan struct{})
```
Scarjit, line 198: s.running can now also be set to false ?

Scarjit (changes requested):

Scarjit (approved):

umh-core/pkg/communicator/actions/stateMocker.go
```diff
@@ -0,0 +1,486 @@
...
+		}
+
+		if instance, ok := instances[curDataflowcomponent.Name]; ok {
+			currentState = instance.CurrentState
+			instanceExistsInCurrentSnapshot = true
+		}
+
+		// Apply any pending transitions
+		currentState = s.checkPendingTransitions(curDataflowcomponent.Name, currentState)
+
```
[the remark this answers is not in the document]
FreiPaul: how does checkPendingTransitions acquire a lock?

umh-core/.DS_Store
JeremyTheocharis, line 1: pls remove this file

umh-core/pkg/communicator/actions/stateMocker.go
```diff
@@ -0,0 +1,486 @@
...
+	return s.ConfigManager
+}
+
+// Tick advances the state mocker by one tick and updates the system state
+func (s *StateMocker) Tick() {
+	s.TickCounter++
+	s.UpdateDfcState()
+}
+
+// here, the actual state update logic is implemented
```
JeremyTheocharis, line 94: add docstring comment what is happening here (can be made based on the inline comments)

umh-core/pkg/communicator/actions/stateMocker.go
```diff
@@ -0,0 +1,486 @@
...
+		Managers: map[string]fsm.ManagerSnapshot{
+			constants.DataflowcomponentManagerName: managerSnapshot,
+		},
+	})
+}
+
+// detect config events (add, remove, edit) and add the corresponding state transitions to the pending transitions
+func (s *StateMocker) detectConfigEvents(curDfcConfig []config.DataFlowComponentConfig, lastDfcConfig []config.DataFlowComponentConfig) {
+
+	if !s.LastConfigSet {
```
JeremyTheocharis, line 134: for what is this case? when does it happen?

umh-core/pkg/communicator/actions/stateMocker.go
```diff
@@ -0,0 +1,486 @@
...
+}
+
+// Tick advances the state mocker by one tick and updates the system state
+func (s *StateMocker) Tick() {
+	s.TickCounter++
+	s.UpdateDfcState()
+}
+
+// here, the actual state update logic is implemented
+func (s *StateMocker) UpdateDfcState() {
```
JeremyTheocharis, line 95: Can you split up UpdateDfcState into multiple functions? It would make it easier to understand. 

What I would do:
1. Rename UpdateDfcState to "AdvanceTick" or something
2. Split up the content of UpdateDfcState into buildNextSnapshot

Here a proposal:
```
package actions

import (
	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/config"
	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsm"
)

/*
StateMocker is a lightweight, in-memory stand-in for the real
manager/scheduler which runs in production.  
The only public method that mutates state is `AdvanceTick`
(you could also call it `Update` or `Step` – pick one and keep it
consistent across the code-base).

The *heavy* work – comparing the latest config with the previous tick,
figuring out which instances should appear/disappear, and deciding
their next lifecycle state – is delegated to a pure helper called
`buildNextSnapshot`.

Separating **calculation** from **persistence** gives us:

  * deterministic, side-effect-free logic that is trivial to unit-test
  * an explicit data flow (inputs → pure function → outputs)
  * freedom to reuse the calculation in benchmarks or offline tools
*/
type StateMocker struct {
	// --- injected (immutable) ---
	ConfigManager ConfigManager       // provides FullConfig
	StateManager  *fsm.SnapshotManager // owns the in-memory SystemSnapshot

	// --- internal (mutated under mu) ---
	tick              int
	lastDfcConfig     []config.DataFlowComponentConfig
	pendingTransitions map[string][]StateTransition
	ignoreUntilTick    map[string]int

	mu sync.RWMutex
}

/*
AdvanceTick is the *only* method that mutates the internal snapshot.
It performs three steps:

  1. Pull the freshest config from the ConfigManager.
  3. Call the pure helper `buildNextSnapshot` to decide what the world
     *should* look like on this tick.
  4. Persist the returned snapshot into StateManager and store the
     updated transition queues.

Callers never touch StateManager directly; they merely call
`AdvanceTick()` and then query the snapshot through StateManager’s
read-only API.
*/
func (s *StateMocker) AdvanceTick() {
	s.mu.Lock()
	defer s.mu.Unlock()

	s.tick++

	curCfg := s.ConfigManager.GetDataFlowConfig()

	newSnap, newPending, newIgnore := buildNextSnapshot(
		s.tick,
		curCfg,
		s.lastDfcConfig,
		s.StateManager.GetDeepCopySnapshot(),
		s.pendingTransitions,
		s.ignoreUntilTick,
	)

	s.StateManager.UpdateSnapshot(newSnap)
	s.lastDfcConfig = curCfg               // remember for the next tick
	s.pendingTransitions = newPending
	s.ignoreUntilTick = newIgnore
}

/*
buildNextSnapshot is a *pure* function – no global state, no locks, no
I/O.  
It receives *everything* it needs as parameters and returns brand-new
data structures.

Arguments  
  tick               – the logical time-step we are computing  
  curCfg / prevCfg   – data-flow-component sections of FullConfig  
  prevSnap           – deep copy of the previous SystemSnapshot  
  pending            – outstanding lifecycle transitions (copied in)  
  ignoreUntil        – map of instance-IDs → tick to suppress deletion

Return values  
  nextSnap           – how the system should look *after* this tick  
  nextPending        – updated transition queues  
  nextIgnore         – updated ignore-map
*/
func buildNextSnapshot(
	tick int,
	curCfg, prevCfg []config.DataFlowComponentConfig,
	prevSnap *fsm.SystemSnapshot,
	pending map[string][]StateTransition,
	ignoreUntil map[string]int,
) (
	nextSnap *fsm.SystemSnapshot,
	nextPending map[string][]StateTransition,
	nextIgnore map[string]int,
) {
	// -----------------------------------------------------------------
	// 1. Diff configs (added / removed / edited / renamed components)
	//    and update `pending` & `ignoreUntil`.
	// -----------------------------------------------------------------
	// TODO: implement the diff logic described in detectConfigEvents()

	// -----------------------------------------------------------------
	// 2. Start with a shallow copy of prevSnap, then iterate over every
	//    component that *should* exist according to curCfg.
	//    For each component:
	//      a) look up current lifecycle state from prevSnap
	//      b) consult pending[tick] to see if a transition is due
	//      c) create a fresh FSMInstanceSnapshot with new CurrentState
	// -----------------------------------------------------------------
	// TODO: implement reconciliation loop

	// -----------------------------------------------------------------
	// 3. Handle components that have been removed from the config but
	//    are still inside ignoreUntil – keep them around until the
	//    grace period expires.
	// -----------------------------------------------------------------
	// TODO: implement ignore-window logic

	// Return fresh copies so the caller can assign without aliasing.
	return nextSnap, nextPending, nextIgnore
}

/*
StateTransition is a single “at tick T the instance must reach state S”
instruction.  It is kept outside the FSM snapshot so tests can inject
custom sequences without fiddling with the snapshot directly.
*/
type StateTransition struct {
	TickAt int
	State  string
}

```

umh-core/pkg/communicator/actions/stateMocker.go
```diff
@@ -0,0 +1,486 @@
...
+}
+
+// GetConfigManager returns the config manager that is being used by the state mocker
+func (s *StateMocker) GetConfigManager() ConfigManager {
+	return s.ConfigManager
+}
+
+// Tick advances the state mocker by one tick and updates the system state
+func (s *StateMocker) Tick() {
+	s.TickCounter++
```
JeremyTheocharis, line 90: doesn't this require uisng the mutex?

JeremyTheocharis (changes requested):

JeremyTheocharis (approved):

PR #1981, opened 2025-04-28
Eng 2876 improve remove handling & fixed bug with DFCs not going out of starting state

JeremyTheocharis:
Benthos/S6 teardown is now *deterministic, idempotent, and fully cleaned-up*

### Why

* A configuration change that deletes a Benthos instance should remove  
  *all* traces of  
  `run/service/benthos-<name>` **and**  
  `run/service/benthos-monitor-<name>` **and** the corresponding log trees.  
  In practice the main service vanished, but the monitor or its S6 supervise
  directory kept lingering, which prevented later re-creates and polluted
  `/run/service` as well as the log volume.

* The root cause:  
  1.  `RemoveBenthosFromS6Manager` and its monitor cousin returned **nil** as
      soon as the config slice was trimmed – long before the child S6-FSM
      had actually executed its own `Remove()`.  
      The parent FSM therefore fired *remove_done* immediately, tore down
      the instance record and went on.  
  2.  `DefaultService.Remove()` and `ForceRemove()` considered their work
      done after deleting the *service* directory, ignoring the rotated log
      directory.  
      Failures during the delete went unnoticed, so half-deleted trees were
      silently accepted.

Did then the same logic with the remaining fsms (except Redpanda, container and agent, which will never be deleted)

The bug with DFCs not getting otu of starting state is addressed here as well, changed a `28` to a `29`

## What is missing and should be delivered in a second PR

- there needs to be a S6 clean up routine, that finds dangling S6 services and logs and kills them. It is useful for agent crashes

## Summary by CodeRabbit

- **New Features**
  - Added a "Force Remove" option for S6 services, enabling unconditional and robust teardown of services.
  - Improved log parsing for S6 services with structured timestamped entries.
  - Introduced retryable removal pending error states for multiple service types to enhance removal workflows.

- **Bug Fixes**
  - Enhanced idempotency and error handling for service removal across Benthos, Nmap, DataflowComponent, and Connection managers.
  - Service removal actions now gracefully handle ongoing removals, preventing premature errors.

- **Refactor**
  - Streamlined and clarified error handling in removal methods for multiple service types using explicit error categorization.
  - Exported and documented key service name and removal methods for better maintainability.
  - Refactored removal logic to be allocation-efficient and idempotent.

- **Tests**
  - Added comprehensive tests for S6 service removal and log parsing.
  - Expanded integration tests to verify removal hooks and idempotent behaviors.
  - Updated existing tests to align with new idempotent removal logic.

- **Chores**
[...]

PR #1982, opened 2025-04-29
ENG-2882 Reliable remove function

JeremyTheocharis:
Benthos/S6 teardown is now *deterministic, idempotent, and fully cleaned-up*

### Why

* A configuration change that deletes a Benthos instance should remove  
  *all* traces of  
  `run/service/benthos-<name>` **and**  
  `run/service/benthos-monitor-<name>` **and** the corresponding log trees.  
  In practice the main service vanished, but the monitor or its S6 supervise
  directory kept lingering, which prevented later re-creates and polluted
  `/run/service` as well as the log volume.

* The root cause:  
  1.  `RemoveBenthosFromS6Manager` and its monitor cousin returned **nil** as
      soon as the config slice was trimmed – long before the child S6-FSM
      had actually executed its own `Remove()`.  
      The parent FSM therefore fired *remove_done* immediately, tore down
      the instance record and went on.  
  2.  `DefaultService.Remove()` and `ForceRemove()` considered their work
      done after deleting the *service* directory, ignoring the rotated log
      directory.  
      Failures during the delete went unnoticed, so half-deleted trees were
      silently accepted.

Did then the same logic with the remaining fsms (except Redpanda, container and agent, which will never be deleted)

The bug with DFCs not getting otu of starting state is addressed here as well, changed a `28` to a `29`

## What is missing and should be delivered in a second PR

- there needs to be a S6 clean up routine, that finds dangling S6 services and logs and kills them. It is useful for agent crashes

## Summary by CodeRabbit

- **New Features**
  - FSMs now detect and handle instances stuck in transient lifecycle states, triggering automatic or forced removal.
  - Added no-op creation checks across multiple FSM types for improved lifecycle management.
  - Introduced a new debug configuration for Go tests targeting specific service directories.
- **Bug Fixes**
  - Enhanced idempotent and gradual removal processes to prevent premature deletion while FSMs remain active.
  - Improved error handling and logging during service removal, status checks, and monitoring.
  - Refined log parsing with nanosecond precision timestamps for better diagnostics.
- **Refactor**
  - Centralized FSM lifecycle reconciliation using a service registry provider interface.
  - Replaced deprecated error variables with standardized error handling.
  - Rewrote service removal to be concurrent, timeout-aware, and robust with forced removal fallback.
  - Updated monitoring scripts to capture error output and improved parsing for metrics and configs.
- **Tests**
  - Added tests verifying FSM removal hooks and removal behavior.
  - Adjusted existing tests to accommodate idempotent removal semantics and updated method signatures.
- **Chores**
  - Updated configuration files and IDE launch settings.
  - Enhanced system information output with container-level CPU cgroup statistics.


JeremyTheocharis:
This PR build upon https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/1981, but because I added a lot of bug fixes here I am using this one now

umh-core/pkg/service/s6/s6.go
```diff
@@ -1101,31 +1177,136 @@ func (s *DefaultService) GetS6ConfigFile(ctx context.Context, servicePath string
...
+			return nil // no supervisor ⇒ nothing to kill
+		}
+		if pid, err := strconv.Atoi(strings.TrimSpace(string(data))); err == nil && pid > 0 {
+			err = syscall.Kill(pid, syscall.SIGTERM)
+			if err != nil {
+				return fmt.Errorf("failed to kill supervisor for service %s: %w", dir, err)
+			}
+		}
+		return nil
 	}
```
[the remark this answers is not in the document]
JeremyTheocharis: it is fine in force remove

umh-core/internal/fsm/baseFSM.go
```diff
@@ -393,3 +418,67 @@ func (s *BaseFSMInstance) HandlePermanentError(
...
+			}
+			return err, false
+		}
+		return s.SendEvent(ctx, LifecycleEventRemoveDone), true
+	case LifecycleStateRemoved:
+		return standarderrors.ErrInstanceRemoved, true
+	default:
+		// If we are not in a lifecycle state, just continue
+		return nil, false
+	}
```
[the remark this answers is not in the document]
led0nk: This might make sense (throughout all fsms later on), because whenever the `SendEvent` fails i woulnd't consider this as reconciled.

umh-core/pkg/fsm/base_manager.go
```diff
@@ -656,3 +664,118 @@ func (m *BaseFSMManager[C]) schedule(base uint64) uint64 {
...
+		if err != nil {
+			sentry.ReportFSMErrorf(
+				m.logger,
+				instanceName,
+				m.managerName,
+				"remove_error",
+				"Error removing instance: %v",
+				err,
+			)
+			if forceRemover, ok := inst.(interface {
```
led0nk, line 759: maybe we should throw/report an error if `!ok` otherwise i'd assume you wouldn't recognize, or am i on the wrong track?

umh-core/pkg/service/s6/s6.go
```diff
@@ -1101,31 +1184,136 @@ func (s *DefaultService) GetS6ConfigFile(ctx context.Context, servicePath string
...
+	// 2. SIGTERM lingering s6-supervise processes to avoid resurrection
+	//--------------------------------------------------------------------
+	killSupervise := func(dir string) error {
+		pidFile := filepath.Join(dir, "supervise", "pid")
+		data, err := fsService.ReadFile(ctx, pidFile)
+		if err != nil || len(data) == 0 {
+			return nil // no supervisor ⇒ nothing to kill
+		}
+		if pid, err := strconv.Atoi(strings.TrimSpace(string(data))); err == nil && pid > 0 {
+			err = syscall.Kill(pid, syscall.SIGTERM)
```
Scarjit, line 1244: I assume SIGTERM is enough here, but since this is force remove a SIGKILL might be an option too.

Scarjit (approved):

umh-core/pkg/service/s6/s6.go
JeremyTheocharis: i think for now sigterm is enough and actually better as it might close other things as well: https://skarnet.org/software/s6/s6-supervise.html (signals chapter)

led0nk (approved):

umh-core/internal/fsm/baseFSM.go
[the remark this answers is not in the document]
JeremyTheocharis: i marked it as reconciled, as it already tried to reconcile something in this tick and took maybe some significant time, so we should skip the remaining managers and wait for the next tick

umh-core/pkg/fsm/base_manager.go
JeremyTheocharis: good point, added a sentry issue

PR #1996, opened 2025-05-06
Explicitly set redpanda config

Scarjit:
This pr sets the redpanda default configs (and our core/mem config) explicitly in our config if no config is present.

## Summary by CodeRabbit

- **New Features**
  - Added default configuration settings for Redpanda, including topic retention and resource limits.


umh-core/pkg/config/env.go
```diff
@@ -105,6 +106,16 @@ func LoadConfigWithEnvOverrides(ctx context.Context, configManager *FileConfigMa
 				FSMInstanceConfig: FSMInstanceConfig{
 					DesiredFSMState: "active", // Default desired state for Redpanda
 				},
+				RedpandaServiceConfig: redpandaserviceconfig.RedpandaServiceConfig{
+					Topic: redpandaserviceconfig.TopicConfig{
+						DefaultTopicRetentionMs:    604800000,
```
kanapuli, line 111: Better to add a comment that we are targeting 7 days. 60480000 looks like a magic number on the first look

umh-core/pkg/config/env.go
```diff
@@ -105,6 +106,16 @@ func LoadConfigWithEnvOverrides(ctx context.Context, configManager *FileConfigMa
...
+						DefaultTopicRetentionBytes: 0,
```
kanapuli, line 112: (suggestion) Shouldn't we have a number here, atleast a big number like a `1 GB` ? Because an high throughput topic could quickly fill up the disk and with the retention policy of 7 days, the application would be already down with low disk.

Scarjit: We currently have no limit (per-default) on mgmtcompanion, so i would leave it as it for now.

kanapuli (approved):

PR #1999, opened 2025-05-06
[ENG-2947] Set snappy as default compression type

Scarjit:
## Summary by CodeRabbit

- **New Features**
  - Added support for configuring the default compression type for Redpanda topics, with "snappy" set as the default value.
  - The compression type is now included in generated YAML configurations and validated during configuration processing.

- **Bug Fixes**
  - Improved detection and reporting of differences in topic compression settings during configuration comparisons.

- **Tests**
  - Extended test coverage to verify correct handling and rendering of the compression type in configurations and YAML output.

- **Documentation**
  - Updated configuration defaults and examples to reflect the new compression type setting.


umh-core/pkg/config/redpandaserviceconfig/comparator.go
```diff
@@ -66,6 +66,11 @@ func (c *Comparator) ConfigDiff(desired, observed RedpandaServiceConfig) string
 			normDesired.Topic.DefaultTopicRetentionBytes, normObserved.Topic.DefaultTopicRetentionBytes))
 	}
 
+	if normDesired.Topic.DefaultTopicCompressionType != normObserved.Topic.DefaultTopicCompressionType {
```
JeremyTheocharis, line 69: this results in a breaking change right?

old configs will not have snappy, so this will always fail, but the system can also not reconcile towards the new state as the updating does not work.

so it will keep failing.

JeremyTheocharis: i would like to prevent issues like this so that updating does not cause a bricked instance.

JeremyTheocharis (changes requested):

led0nk (approved):

PR #2001, opened 2025-05-06
[ENG-2949] fix fatal when there is a wrong dfc desired state

FreiPaul:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling in the control loop to log a warning and continue operation when encountering "invalid desired state" errors, preventing unnecessary interruptions.

### DIFF
--- a/umh-core/pkg/control/loop.go
+++ b/umh-core/pkg/control/loop.go
@@ -38,6 +38,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/tiendc/go-deepcopy"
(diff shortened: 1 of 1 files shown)


umh-core/pkg/control/loop.go
```diff
@@ -189,6 +190,10 @@ func (c *ControlLoop) Execute(ctx context.Context) error {
 				if errors.Is(err, context.DeadlineExceeded) {
 					// For timeouts, log warning but continue
 					sentry.ReportIssuef(sentry.IssueTypeWarning, c.logger, "Control loop reconcile timed out: %v", err)
+				} else if strings.Contains(err.Error(), "invalid desired state") {
```
JeremyTheocharis, line 193: can you in the sources maybe create an error for it, and then match it here? then it becomes easier to find where the invalid desired state is coming from

JeremyTheocharis (changes requested):

JeremyTheocharis:
Closed in Favor of https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/2003

PR #2004, opened 2025-05-07
ENG-2952 Fix redpanda config application

Scarjit:


umh-core/pkg/fsm/redpanda/reconcile.go
```diff
@@ -220,16 +220,40 @@ func (r *RedpandaInstance) reconcileOperationalStates(ctx context.Context, servi
...
 		return r.reconcileTransitionToStopped(ctx, services, currentState)
+	case OperationalStateRestarting:
+		return r.reconcileTransitionToRestarting(ctx, services, currentState)
 	default:
 		return fmt.Errorf("invalid desired state: %s", desiredState), false
 	}
 }
 
+// reconcileTransitionToRestarting handles transitions when the desired state is Restarting.
+func (r *RedpandaInstance) reconcileTransitionToRestarting(ctx context.Context, services serviceregistry.Provider, currentState string) (err error, reconciled bool) {
```
JeremyTheocharis, line 238: How about instead of introducing a restarting state, simply set it once to "stopping" by sending the stop event, then it should restart automatically (should finish the stopping, and should then start again)

JeremyTheocharis: because restarting is not a desired state, it is just an action.

or alternatively, just work with an event, that can be send in running states only:
- reconcileTransitionToActive (and if the current state is then to_be_restarted, then call the restart function, if it is restarting, check whether the restart was completed)

mostly the logic remains the same, just that we don't have a desired state of restarting

umh-core/pkg/fsm/redpanda/actions.go
```diff
@@ -321,7 +334,12 @@ func (r *RedpandaInstance) UpdateObservedStateOfInstance(ctx context.Context, se
 			r.baseFSMInstance.GetLogger().Debugf("Configuration differences: %s", diffStr)
 
 			// Update the config in the S6 manager
-			err := r.service.UpdateRedpandaInS6Manager(ctx, &r.config, r.baseFSMInstance.GetID())
+			var needsRestart bool
+			needsRestart, err = r.service.UpdateRedpandaInS6Manager(ctx, &r.ObservedState.ObservedRedpandaServiceConfig, &r.config, r.baseFSMInstance.GetID(), tick)
+			if needsRestart {
+				r.baseFSMInstance.GetLogger().Debugf("Redpanda service %s needs to be restarted", r.baseFSMInstance.GetID())
+				return r.RestartInstance(ctx, services.GetFileSystem())
```
JeremyTheocharis, line 341: this here could then send a "restart" event, which then causes it to be moved into to_be_restarted. after that it goes into restarting, and as soon as it is confirmed that the restart worked, it should go into starting state (so we can leverage the safety checks there)

Scarjit: Not sure if we even need that.
Since RestartInstance now sends the Stop event, S6 will shut it down, and once it is in the shutdown state, we will automatically reconcile to active again.

PR #2006, opened 2025-05-07
Eng 2955 provide redpanda status in status message

JeremyTheocharis:
Now one can see the redpanda status in the console. What can still happen is that redpanda remains degraded in the UI. We would need to address that

@ricci2511 will take care of changing msg/sec to bytes/sec in the frontend

### DIFF
--- a/umh-core/pkg/communicator/pkg/generator/generator.go
+++ b/umh-core/pkg/communicator/pkg/generator/generator.go
@@ -25,6 +25,7 @@ import (
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsm/agent_monitor"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsm/container"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsm/dataflowcomponent"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsm/redpanda"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/logger"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/models"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/sentry"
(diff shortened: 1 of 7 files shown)


umh-core/pkg/communicator/pkg/generator/generator.go
```diff
@@ -385,6 +399,66 @@
...
+		}
+
+		// Now fetch the last observed service info
+		observedStateServiceInfo := observedState.ServiceInfoSnapshot
+
+		// Now calculate the throughput metrics if available
+		if observedStateServiceInfo.RedpandaStatus.RedpandaMetrics.MetricsState != nil {
+			redpandaData.AvgIncomingThroughputPerMinuteInMsgSec = float64(observedStateServiceInfo.RedpandaStatus.RedpandaMetrics.MetricsState.Input.BytesPerTick) / constants.DefaultTickerTime.Seconds()
+			redpandaData.AvgOutgoingThroughputPerMinuteInMsgSec = float64(observedStateServiceInfo.RedpandaStatus.RedpandaMetrics.MetricsState.Output.BytesPerTick) / constants.DefaultTickerTime.Seconds()
+		}
```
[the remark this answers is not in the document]
JeremyTheocharis: input/output are not pointers, therefore they cannot be nil

umh-core/pkg/models/status_message.go
```diff
@@ -164,8 +164,8 @@ type DfcBridgeInfo struct {
 
 type Redpanda struct {
 	Health                                 *Health `json:"health"`
-	AvgIncomingThroughputPerMinuteInMsgSec float64 `json:"avgIncomingThroughputPerMinuteInMsgSec"` // Incoming messages per second, averaged over a minute
-	AvgOutgoingThroughputPerMinuteInMsgSec float64 `json:"avgOutgoingThroughputPerMinuteInMsgSec"` // Outgoing messages per second, averaged over a minute
+	AvgIncomingThroughputPerMinuteInMsgSec float64 `json:"avgIncomingThroughputPerMinuteInMsgSec"` // Incoming bytes per second, averaged over a minute
```
Scarjit, line 167: Variable name does not match description

Scarjit (changes requested):

umh-core/pkg/models/status_message.go
ricci2511: We can change this safely without much backwards compatibility issues for clients, since this field was never set until now

Scarjit (approved):

umh-core/pkg/models/status_message.go
JeremyTheocharis: ok perfect, this was where I was worrying about and why I did not change the field

PR #2007, opened 2025-05-07
Eng 2953 better user message for log degration

FreiPaul:


JeremyTheocharis:
Included in this PR here: https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/2010

PR #2016, opened 2025-05-08
[ENG-2971] Set the Protocolconverter output to uns plugin

kanapuli:
## Summary by CodeRabbit

- **Bug Fixes**
  - The output configuration for the ProtocolConverter now defaults to using the "uns" plugin for consistent message processing.
- **Tests**
  - Test cases were updated to confirm the automatic setting of the "uns" plugin as the default output.


JeremyTheocharis (changes requested):
what is here the default topic (so topic, not kafka_topic)

--> the plugin should fail (refuse message, log it, etc.) when that variable is not set)

led0nk:
Just FYI don't rely on the comment too much, just have a look how it's done for Redpanda with the templating and the yaml generation, that's how we need it here.

Scarjit:
Resolved by https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/2028/files, which also does this

PR #2028, opened 2025-05-09
Eng 2845 feat pc service (WIP)

led0nk:


JeremyTheocharis:
still need to push some of the fixes, but now unit tests are failing, but I need to go into a meeting, will do it later

umh-core/pkg/config/config.go
```diff
@@ -32,13 +33,14 @@ type FullConfig struct {
...
 
 type InternalConfig struct {
-	Services        []S6FSMConfig           `yaml:"services,omitempty"`        // Services to manage, can be updated while running
-	Benthos         []BenthosConfig         `yaml:"benthos,omitempty"`         // Benthos services to manage, can be updated while running
-	Nmap            []NmapConfig            `yaml:"nmap,omitempty"`            // Nmap services to manage, can be updated while running
-	Redpanda        RedpandaConfig          `yaml:"redpanda,omitempty"`        // Redpanda config, can be updated while running
-	BenthosMonitor  []BenthosMonitorConfig  `yaml:"benthosMonitor,omitempty"`  // BenthosMonitor config, can be updated while running
-	Connection      []ConnectionConfig      `yaml:"connection,omitempty"`      // Connection services to manage, can be updated while running
-	RedpandaMonitor []RedpandaMonitorConfig `yaml:"redpandaMonitor,omitempty"` // RedpandaMonitor config, can be updated while running
+	Services          []S6FSMConfig             `yaml:"services,omitempty"`          // Services to manage, can be updated while running
```
Scarjit, line 36: Just as a note: We should keep one standard of indentation/formatting.
I would just go for https://pkg.go.dev/cmd/gofmt without any special rules.

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -0,0 +1,756 @@
...
+	foundConn := false
+	indexConn := -1
+	for i, config := range p.connectionConfig {
+		if config.Name == underlyingName {
+			foundConn = true
+			indexConn = i
+			break
+		}
+	}
+	// Check if the dfcconfig exists
```
Scarjit, line 451: You could early abort here, instead of also looping the dfc configs, in case of no conn found

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -0,0 +1,756 @@
...
+
+	p.logger.Infof("Removing dataflowcomponent %s", protConvName)
+
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
+
+	underlyingName := p.getUnderlyingName(protConvName)
+
+	connSliceRemoveByName := func(in []config.ConnectionConfig, name string) []config.ConnectionConfig {
```
Scarjit, line 517: It could make sense to give all of our configs an implementation of an interface to return Name.
This would allow us to have a generic filter function.

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -0,0 +1,756 @@
...
+	// Find and update our cached config
+	connFound := false
+	for i, config := range p.connectionConfig {
+		if config.Name == underlyingName {
+			p.connectionConfig[i].DesiredFSMState = connectionfsm.OperationalStateUp
+			connFound = true
+			break
+		}
+	}
+
```
Scarjit, line 587: Here again could be an early return

Scarjit (changes requested):

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -0,0 +1,756 @@
...
+	protConvName string,
+) bool {
+	if ctx.Err() != nil {
+		return false
+	}
+	underlyingName := p.getUnderlyingName(protConvName)
+
+	// Check if the actual service exists
+	connExists := p.connectionService.ServiceExists(ctx, filesystemService, underlyingName)
+	dfcExists := p.dataflowComponentService.ServiceExists(ctx, filesystemService, underlyingName)
```
FreiPaul, line 724: Dont we need to use `protConvName` here instead of `underlyingName`?

FreiPaul (changes requested):

umh-core/pkg/config/config.go
JeremyTheocharis: it was reformatted because ProtocolConverterConfig is logner than the rest so all the `yaml:...` were aligned, this is default behaviour

umh-core/pkg/service/protocolconverter/protocolconverter.go
JeremyTheocharis: minor thing but done

umh-core/pkg/service/protocolconverter/protocolconverter.go
JeremyTheocharis: also thought about that, but decided not to use generics (as they are a more complicated pattern and it is not really much duplicated code)

umh-core/pkg/config/protocolconverterserviceconfig/normalizer.go
```diff
@@ -28,13 +28,15 @@ func NewNormalizer() *Normalizer {
 }
 
 // NormalizeConfig applies ProtocolConverter defaults to a structured config
-func (n *Normalizer) NormalizeConfig(cfg *ProtocolConverterServiceConfig) {
+func (n *Normalizer) NormalizeConfig(cfg ProtocolConverterServiceConfig) ProtocolConverterServiceConfig {
 
 	// We need to first normalize the underlying DFCServiceConfig
```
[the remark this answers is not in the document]
JeremyTheocharis: not all of them (e.g., DFC wasnt doing it) but standardized it now

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -0,0 +1,756 @@
...
+	// on disk.
+	GetConfig(ctx context.Context, filesystemService filesystem.Service, protConvName string) (protocolconverterserviceconfig.ProtocolConverterServiceConfig, error)
+
+	// Status aggregates health from Connection, DFC and Redpanda into a single
+	// snapshot.  The returned structure is **read-only** – callers must not
+	// mutate it.
+	Status(ctx context.Context, services serviceregistry.Provider, snapshot fsm.SystemSnapshot, protConvName string, tick uint64) (ServiceInfo, error)
+
+	// AddToManager adds a ProtocolConverter to the Connection & DFC manager
+	AddToManager(ctx context.Context, filesystemService filesystem.Service, protConvCfg *protocolconverterserviceconfig.ProtocolConverterServiceConfig, protConvName string) error
```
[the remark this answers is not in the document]
JeremyTheocharis: this is from original Janik proposal, and I took it over because it adds it to multiple managers (and not just a single one)

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -0,0 +1,756 @@
...
+
+	// UpdateInManager updates an existing ProtocolConverter in the Connection & DFC manager
+	UpdateInManager(ctx context.Context, filesystemService filesystem.Service, protConvCfg *protocolconverterserviceconfig.ProtocolConverterServiceConfig, protConvName string) error
```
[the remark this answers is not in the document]
JeremyTheocharis: see above

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -0,0 +1,756 @@
...
+
+	// RemoveFromManager removes a ProtocolConverter from the Connection & DFC manager
+	RemoveFromManager(ctx context.Context, filesystemService filesystem.Service, protConvName string) error
```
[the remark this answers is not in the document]
JeremyTheocharis: see above

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -0,0 +1,756 @@
...
+
+	// Start starts a ProtocolConverter
+	Start(ctx context.Context, filesystemService filesystem.Service, protConvName string) error
+
+	// Stop stops a ProtocolConverter
+	Stop(ctx context.Context, filesystemService filesystem.Service, protConvName string) error
+
+	// ForceRemove removes a ProtocolConverter from the Connetion & DFC manager
+	ForceRemove(ctx context.Context, filesystemService filesystem.Service, protConvName string) error
```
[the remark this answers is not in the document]
JeremyTheocharis: adjusted

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -0,0 +1,756 @@
...
+	AddToManager(ctx context.Context, filesystemService filesystem.Service, protConvCfg *protocolconverterserviceconfig.ProtocolConverterServiceConfig, protConvName string) error
+
+	// UpdateInManager updates an existing ProtocolConverter in the Connection & DFC manager
+	UpdateInManager(ctx context.Context, filesystemService filesystem.Service, protConvCfg *protocolconverterserviceconfig.ProtocolConverterServiceConfig, protConvName string) error
+
+	// RemoveFromManager removes a ProtocolConverter from the Connection & DFC manager
+	RemoveFromManager(ctx context.Context, filesystemService filesystem.Service, protConvName string) error
+
+	// Start starts a ProtocolConverter
+	Start(ctx context.Context, filesystemService filesystem.Service, protConvName string) error
```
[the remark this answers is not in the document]
JeremyTheocharis: adjusted

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -0,0 +1,756 @@
...
+	// returns:
+	//   err        – unrecoverable problem (configuration bug, ctx cancellation)
+	//   reconciled – true when *any* child manager made progress
+	ReconcileManager(ctx context.Context, services serviceregistry.Provider, tick uint64) (error, bool)
+}
+
+// ServiceInfo holds information about the ProtocolConverters underlying health states.
+type ServiceInfo struct {
+	// ConnectionObservedState is the last sample from the connection manager.
+	ConnectionObservedState connectionfsm.ConnectionObservedState
```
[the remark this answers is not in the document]
JeremyTheocharis: there is only one connection per PC?

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -0,0 +1,756 @@
...
+	// ConnectionFSMState is the *current* FSM state string (e.g. "up", "stopped").
+	ConnectionFSMState string
+
+	// DataflowComponentObservedState mirrors the DFC manager.
+	DataflowComponentObservedState dfcfsm.DataflowComponentObservedState
```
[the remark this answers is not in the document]
JeremyTheocharis: theoretically yes, but this is coming from Janiks original proposal and I did not want to refactor this whole PR. can still add that at a later stage and don't want to overengineer right now.

still, i spend now 1 hour to add at least the base implementation (e.g., the config, the comparisons, the service) without actually implementing it

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -0,0 +1,756 @@
...
+}
+
+// getUnderlyingName converts a protConvName to its underlying service name
+func (p *ProtocolConverterService) getUnderlyingName(protConvName string) string {
+	return fmt.Sprintf("protocolconverter-%s", protConvName)
+}
+
+// getConnectionName converts a protConvName to the name of the connection service
+func (p *ProtocolConverterService) getConnectionName(protConvName string) string {
+	return fmt.Sprintf("connection-%s", p.getUnderlyingName(protConvName))
```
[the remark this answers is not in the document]
JeremyTheocharis: yes

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -0,0 +1,756 @@
...
+	generatedConfig, err := p.generateProtocolConverterYaml(protConvConfig)
+	if err != nil {
+		return connectionserviceconfig.ConnectionServiceConfig{},
+			dataflowcomponentserviceconfig.DataflowComponentServiceConfig{},
+			fmt.Errorf("failed to generate ProtocolConverterServiceConfig: %w", err)
+	}
+
+	protConvConfig.DataflowComponentServiceConfig = dataflowcomponentserviceconfig.DataflowComponentServiceConfig{
+		BenthosConfig: dataflowcomponentserviceconfig.BenthosConfig{
+			Output: map[string]any{"uns": generatedConfig},
```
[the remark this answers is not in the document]
JeremyTheocharis: good point, that line here did not make sense

umh-core/pkg/service/protocolconverter/protocolconverter.go
JeremyTheocharis: no, underlying name is the correct one

umh-core/pkg/service/protocolconverter/protocolconverter_test.go
```diff
@@ -0,0 +1,778 @@
...
+	connfsm "github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsm/connection"
+	dfcfsm "github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsm/dataflowcomponent"
+	redpandafsm "github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsm/redpanda"
+	connservice "github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/service/connection"
+	dfcservice "github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/service/dataflowcomponent"
+	redpandaservice "github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/service/redpanda"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/serviceregistry"
+)
+
+var _ = Describe("DataFlowComponentService", func() {
```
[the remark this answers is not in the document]
JeremyTheocharis: there are now tests for the BuildRuntimeConfig function, which is the successor to GenerateConfig

umh-core/pkg/config/templating.go
```diff
@@ -0,0 +1,180 @@
...
+//
+//   - **Anchor**  – a node that can later be referenced (`&base …`)
+//   - **Alias**   – a node that *is* such a reference (`*base`)
+//
+// Why we need it
+// --------------
+// The UMH configuration file lets operators use YAML templating.
+// We therefore want to detect whether a mapping still
+// contains anchors/aliases and, if so, refuse automatic mutations
+// (Add/Edit/Delete) so that we never flatten or overwrite a handcrafted
```
FreiPaul, line 39: Really also add and delete? We currently only protect the edit action, right?

umh-core/pkg/config/protocolconverterserviceconfig/generator.go
```diff
@@ -51,42 +52,106 @@ func (g *Generator) RenderConfig(cfg ProtocolConverterServiceConfig) (string, er
...
 	}
 
-	normalizedDFCConfig := dataflowcomponentserviceconfig.NormalizeConfig(dfcConfig)
+	// Process location map correctly to ensure it's a proper map[string]string
+	// This handles converting location keys (like level numbers) to the correct format
+	locationMap := make(map[string]string)
+	if rawLocation, ok := raw["location"].(map[string]any); ok {
+		// Convert map[string]any to map[string]string
+		for k, v := range rawLocation {
+			// Convert any value type to string if possible
```
FreiPaul, line 103: Can't we do this way easier with `locationMap[k] = fmt.Sprint(v)`?

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -0,0 +1,1170 @@
...
+			break
+		}
+	}
+
+	foundWriteDFC := false
+	indexWriteDFC := -1
+	for i, config := range p.dataflowComponentConfig {
+		if config.Name == p.getUnderlyingDFCWriteName(protConvName) {
+			foundWriteDFC = true
+			indexWriteDFC = i
```
FreiPaul, line 808: break

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -0,0 +1,1170 @@
...
+// It delegates to the ProtocolConverter service's reconciliation function,
+// which ensures that the actual state matches the desired state
+// for all services.
+//
+// This should be called periodically by a control loop.
+func (p *ProtocolConverterService) ReconcileManager(
+	ctx context.Context,
+	services serviceregistry.Provider,
+	tick uint64,
+) (error, bool) {
```
FreiPaul, line 1062: error should be the last return parameter? For me golangci always complains if error is not the last return type

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -0,0 +1,1170 @@
...
+	//    • Add bridged_by header derived from nodeName and pcName
+	//
+	// 3. Render all three sub-templates:
+	//    • Connection
+	//    • read-DFC   (with UNS **output** enforced)
+	//    • write-DFC  (with UNS **input**  enforced)
+	//
+	// Notes
+	// ─────
+	//   - The function is pure: it performs no side-effects and never mutates *Spec*.
```
FreiPaul, line 83: Here you say that the function never mutates Spec. I also had such a case in the stateMocker where I thought, I made a copy but go internally referred to the same memory address, resulting in the effect that the original variable was mutated. This might maybe happen here as well: https://chatgpt.com/share/e/682dc8ff-faac-8007-b44f-4bd802c6792e
Consider, making a deep copy

FreiPaul (changes requested):

umh-core/pkg/service/protocolconverter/protocolconverter.go
JeremyTheocharis: this is true, but we started once with it and now we need to be consistent (or create a PR adjusting it)

umh-core/pkg/service/protocolconverter/protocolconverter.go
JeremyTheocharis: good point, i ensured that i dont pass a pointer from now on

umh-core/pkg/communicator/pkg/generator/generator.go
```diff
@@ -17,28 +17,12 @@ package generator
...
-	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/constants"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsm"
-	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/logger"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/models"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/sentry"
 	"go.uber.org/zap"
 )
 
-const (
-	// Manager name constants
```
Scarjit, line 29: We now have definitions for these in both constants/managernames.go and fsm/snapshot.go.
Both of them are activly referenced.

umh-core/pkg/config/config.go
```diff
@@ -19,16 +19,19 @@ import (
...
-	DataFlow []DataFlowComponentConfig `yaml:"dataFlow,omitempty"` // DataFlow components to manage, can be updated while running
-	Internal InternalConfig            `yaml:"internal,omitempty"` // Internal config, not to be used by the user, only to be used for testing internal components
+	Agent             AgentConfig               `yaml:"agent"`                       // Agent config, requires restart to take effect
+	DataFlow          []DataFlowComponentConfig `yaml:"dataFlow,omitempty"`          // DataFlow components to manage, can be updated while running
+	ProtocolConverter []ProtocolConverterConfig `yaml:"protocolConverter,omitempty"` // ProtocolConverter config, can be updated while runnnig
+	Internal          InternalConfig            `yaml:"internal,omitempty"`          // Internal config, not to be used by the user, only to be used for testing internal components
+	Templates         []any                     `yaml:"templates,omitempty"`         // proof of concept for general yaml templates, where anchor can be placed, see also examples/example-config-dataflow-templated.yaml
```
Scarjit: Must this be of type []any or would a map[string]interface{} (or []map[string]interface{}) better capture this ?

umh-core/pkg/config/connectionserviceconfig/package.go
```diff
@@ -29,7 +29,7 @@ type ConnectionServiceConfig struct {
 }
 
 // Equal checks if two ConnectionServiceConfigs are equal
-func (c *ConnectionServiceConfig) Equal(other *ConnectionServiceConfig) bool {
```
Scarjit, line 32: This method does not have a ptr receiver, while the GetNmapServiceConfig in utils.go does.
This is not recommended, as it can easily lead to confusion, which callers copy and which reference the underlying types.

This issue is also present in other package.go equal methods

umh-core/pkg/config/dataflowcomponentserviceconfig/normalizer.go
```diff
@@ -25,16 +25,16 @@ func NewNormalizer() *Normalizer {
...
 
 // NormalizeConfig applies DFC defaults to a structured config
-func (n *Normalizer) NormalizeConfig(cfg *DataflowComponentServiceConfig) {
-	// Create a copy
-	if cfg == nil {
-		return
-	}
+func (n *Normalizer) NormalizeConfig(cfg DataflowComponentServiceConfig) DataflowComponentServiceConfig {
+
+	// create a copy
```
Scarjit: I know this is copied, but can we maybe replace it with "create a shallow copy" ?

umh-core/pkg/config/protocolconverterserviceconfig/normalizer_test.go
```diff
@@ -69,66 +71,74 @@ var _ = Describe("ProtocolConverter YAML Normalizer", func() {
...
+						},
 					},
 				},
 			}
 
 			normalizer := NewNormalizer()
-			normalizer.NormalizeConfig(config)
+			config = normalizer.NormalizeConfig(config)
+
+			Expect(config.Template.DataflowComponentReadServiceConfig.BenthosConfig.Input).NotTo(BeNil())
```
Scarjit, line 111: Above you say these are "// Empty but not nil", you could then directly test for this behaviour by checking To(BeEmpty())

umh-core/pkg/config/protocolconverterserviceconfig/package.go
```diff
@@ -17,51 +17,113 @@ package protocolconverterserviceconfig
...
 }
 
 // ConfigDiff is a package-level function for easy config diff generation
-func ConfigDiff(desired, observed *ProtocolConverterServiceConfig) string {
+func ConfigDiff(desired, observed ProtocolConverterServiceConfigSpec) string {
 	return defaultComparator.ConfigDiff(desired, observed)
 }
+
+// ConfigsEqualRuntime is a package-level function for comparing runtime configurations
+func ConfigsEqualRuntime(desired, observed ProtocolConverterServiceConfigRuntime) bool {
```
Scarjit, line 108: This function is unused.

umh-core/pkg/config/protocolconverterserviceconfig/package.go
```diff
@@ -17,51 +17,113 @@ package protocolconverterserviceconfig
...
+	// Convert runtime configs to spec configs for comparison
+	// This allows us to reuse the existing comparison logic
+	desiredSpec := ProtocolConverterServiceConfigSpec{Template: ProtocolConverterServiceConfigTemplate(desired)}
+	observedSpec := ProtocolConverterServiceConfigSpec{Template: ProtocolConverterServiceConfigTemplate(observed)}
+	return defaultComparator.ConfigsEqual(desiredSpec, observedSpec)
+}
+
+// ConfigDiffRuntime is a package-level function for generating diffs between runtime configurations
+func ConfigDiffRuntime(desired, observed ProtocolConverterServiceConfigRuntime) string {
```
Scarjit, line 117: This function is unused.

umh-core/pkg/config/protocolconverterserviceconfig/package.go
```diff
@@ -17,51 +17,113 @@ package protocolconverterserviceconfig
...
+	// Convert runtime configs to spec configs for diffing
+	// This allows us to reuse the existing diff generation logic
+	desiredSpec := ProtocolConverterServiceConfigSpec{Template: ProtocolConverterServiceConfigTemplate(desired)}
+	observedSpec := ProtocolConverterServiceConfigSpec{Template: ProtocolConverterServiceConfigTemplate(observed)}
+	return defaultComparator.ConfigDiff(desiredSpec, observedSpec)
+}
+
+// SpecToRuntime converts a ProtocolConverterServiceConfigSpec to a ProtocolConverterServiceConfigRuntime
+// This is commonly used when we need to convert from the config structure to the runtime structure
+func SpecToRuntime(spec ProtocolConverterServiceConfigSpec) ProtocolConverterServiceConfigRuntime {
```
Scarjit, line 127: This function is unused.

umh-core/pkg/config/templating.go
```diff
@@ -0,0 +1,180 @@
...
+// know *if* templating was used, because:
+//
+//   - If **yes** → the API helpers (`AtomicEdit…`, `AtomicDelete…`)
+//     must refuse to modify this DFC automatically.
+//   - If **no**  → the helpers may proceed and rewrite the config.
+//
+// The top-level loader (`parseConfig`) doesn't need to be aware of any
+// of this: it simply decodes into `FullConfig` and the magic happens
+// inside every DFC.
+func (d *DataFlowComponentConfig) UnmarshalYAML(value *yaml.Node) error {
```
Scarjit, line 87: Ptr and value receives are mixed again.
https://go.dev/doc/faq#methods_on_values_or_pointers:~:text=Next%20is%20consistency.%20If%20some%20of%20the%20methods%20of%20the%20type%20must%20have%20pointer%20receivers%2C%20the%20rest%20should%20too%2C%20so%20the%20method%20set%20is%20consistent%20regardless%20of%20how%20the%20type%20is%20used.%20See%20the%20section%20on%20method%20sets%20for%20details.

umh-core/pkg/config/templating.go
```diff
@@ -0,0 +1,180 @@
...
+	var tmp plain
+
+	// 1. decode into the temporary value just like the default behaviour
+	if err := value.Decode(&tmp); err != nil {
+		return err
+	}
+
+	// 2. locate the child node with key "dataFlowComponentConfig"
+	var cfgNode *yaml.Node
+	for i := 0; i < len(value.Content)-1; i += 2 {
```
Scarjit: Can you document why you jump by 2 for each loop call ?

Scarjit: Can this ever become uneven ?

umh-core/pkg/config/templating.go
```diff
@@ -0,0 +1,180 @@
...
+	var tmp plain
+
+	// 1. decode into the temporary value just like the default behaviour
+	if err := value.Decode(&tmp); err != nil {
+		return err
+	}
+
+	// 2. locate the child node with key "protocolConverterConfig"
+	var cfgNode *yaml.Node
+	for i := 0; i < len(value.Content)-1; i += 2 {
```
Scarjit: See above :)

umh-core/pkg/config/templating.go
```diff
@@ -0,0 +1,180 @@
...
+// {{ … }} actions, renders it with text/template and returns the same
+// struct type fully materialised.
+//
+// Callers **must** supply a fully-merged variable scope; the function does
+// not fetch or inject `.global`, `.internal`, or `.location` keys.
+func RenderTemplate[T any](tmpl T, scope map[string]any) (T, error) {
+	// A. serialise to YAML – keeps anchors & order stable for diffing
+	raw, err := yaml.Marshal(tmpl)
+	if err != nil {
+		return *new(T), err
```
Scarjit: It could help if we wrap the errors here, for easier debugging later.

umh-core/pkg/config/templating.go
```diff
@@ -0,0 +1,180 @@
...
+	d.hasAnchors = hasAnchors(cfgNode)
+
+	return nil
+}
+
+// RenderTemplate takes an *arbitrary* struct that still contains
+// {{ … }} actions, renders it with text/template and returns the same
+// struct type fully materialised.
+//
+// Callers **must** supply a fully-merged variable scope; the function does
```
Scarjit, line 161: This should still have a basic scope != nil validation, preventing accidential panics

umh-core/pkg/fsm/snapshot.go
```diff
@@ -21,6 +21,25 @@ import (
...
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/config"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/config/dataflowcomponentserviceconfig"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/constants"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/logger"
+)
+
+// These are the names of the managers and instances that are part of the system snapshot
+// They are quite inconsistent and need to be cleaned up
+const (
+	// Manager name constants
```
Scarjit: See my comment at the top. this might be a duplicate.

umh-core/pkg/service/protocolconverter/runtime_config/runtime_config.go
```diff
@@ -0,0 +1,243 @@
...
+
+	// 1c) fill gaps up to the highest defined level with "unknown"
+	maxLevel := -1
+	for k := range loc {
+		level, err := strconv.Atoi(k)
+		if err == nil && level > maxLevel {
+			maxLevel = level
+		}
+	}
+	for i := 0; i <= maxLevel; i++ {
```
Scarjit, line 128: Can you document, why do we choose to fill with unknown here, instead of returning an error ?

Scarjit (changes requested):

umh-core/pkg/config/config.go
```diff
@@ -147,6 +172,12 @@ type ConnectionConfig struct {
 	ConnectionServiceConfig connectionserviceconfig.ConnectionServiceConfig `yaml:"connectionServiceConfig"`
 }
 
+// TemplateVariable is a variable that can be used in templating
+type TemplateVariable struct {
```
led0nk: this is still unused - not sure if we should keep this here, maybe we change plans or sth like that

umh-core/pkg/communicator/actions/edit-dataflowcomponent.go
```diff
@@ -665,5 +665,5 @@ func CompareSnapshotWithDesiredConfig(dfcSnapshot *dataflowcomponent.DataflowCom
 			Buffer:             observedConfig.Buffer,
 		},
 	}
-	return dataflowcomponentserviceconfig.NewComparator().ConfigsEqual(&observedConfigInDfcConfig, &desiredConfig.DataFlowComponentServiceConfig)
+	return dataflowcomponentserviceconfig.NewComparator().ConfigsEqual(observedConfigInDfcConfig, desiredConfig.DataFlowComponentServiceConfig)
```
led0nk, line 668: what was the point in moving away from ptrs as especially dfc & pc can have "more huge" configs ?

umh-core/pkg/communicator/pkg/generator/generator.go
JeremyTheocharis: what do you mean by that?

umh-core/pkg/config/nmapserviceconfig/normalizer.go
```diff
@@ -24,5 +24,7 @@ func NewNormalizer() *Normalizer {
 
 // NormalizeConfig doesn't do anything, there is no normalization needed here
 func (n *Normalizer) NormalizeConfig(cfg NmapServiceConfig) NmapServiceConfig {
-	return cfg
+	// create a copy
+	normalized := cfg
```
led0nk: think we might remove this, since there is no normalizing process and we don't need to do another copy here

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -0,0 +1,903 @@
...
+	DataflowComponentWriteObservedState dfcfsm.DataflowComponentObservedState
+	DataflowComponentWriteFSMState      string
+
+	// RedpandaObservedState is included so a protocol-converter can degrade
+	// itself when the message bus is down.
+	RedpandaObservedState redpandafsm.RedpandaObservedState
+	RedpandaFSMState      string
+
+	// LastChange is the global tick when any of the above fields last changed.
+	LastChange uint64
```
led0nk: do we ever access this?

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -0,0 +1,903 @@
...
+
+// getUnderlyingConnectionName returns the external name handed to the **Connection
+// manager**. For the Connection we do **not** add any read/write qualifier –
+// the underlying base name is already unique across all Connection instances.
+//
+// Example:
+//
+//	protConvName = "mixing-station"
+//	→ "protocolconverter-mixing-station"
+func (p *ProtocolConverterService) getUnderlyingConnectionName(protConvName string) string {
```
led0nk, line 215: imo this is just a wrapper for `getUnderlyingName` - we should get rid of one of those

led0nk: might make also sense to put those `underlyingName` funcs together (as we always call all of them), so there cannot be any confusion of naming and return the same modified-input

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -0,0 +1,903 @@
...
+
+// Status returns information about the connection health for the specified connection.
+// It queries the underlying Connection service and then enhances the result with
+// additional context like flakiness detection based on historical data.
+func (p *ProtocolConverterService) Status(
+	ctx context.Context,
+	services serviceregistry.Provider,
+	snapshot fsm.SystemSnapshot,
+	protConvName string,
+	tick uint64,
```
led0nk, line 300: could be a potential leftover where we use double `tick` and `snapshot` - didn't see that fix in #2060

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -0,0 +1,939 @@
...
+	underlyingDFCWriteName := p.getUnderlyingDFCWriteName(protConvName)
+
+	// Check if the connection already exists in our configs
+	for _, existingConfig := range p.connectionConfig {
+		if existingConfig.Name == underlyingConnectionName {
+			return ErrServiceAlreadyExists
+		}
+	}
+
+	connServiceConfig := cfg.ConnectionServiceConfig
```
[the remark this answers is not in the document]
led0nk: why don't we check for that also?

umh-core/pkg/communicator/pkg/generator/generator.go
Scarjit: Screenshot 2025-06-03 at 16 33 24
They both seem to define the same constants.

umh-core/pkg/config/protocolconverterserviceconfig/package.go
JeremyTheocharis: will be used in the fsm

umh-core/pkg/config/protocolconverterserviceconfig/package.go
JeremyTheocharis: will be used in the fsm

umh-core/pkg/config/protocolconverterserviceconfig/package.go
JeremyTheocharis: will be used in the fsm

umh-core/pkg/service/protocolconverter/protocolconverter.go
JeremyTheocharis: i thought long about this and it makes sense to keep it seperated. there are cases where you need underlying name, and there are cases where you need the other version. if we now have one case of a conncetion where underlying is then something else, it becomes difficult to spot mistakes.

believe me, this part here was the casue of at least 8 bugs and I spent at least 4h just on the naming here. decided for consistency and similarity instead of optimization because of it.

umh-core/pkg/communicator/actions/edit-dataflowcomponent.go
JeremyTheocharis: so that it is all standardized. there is not much performance improvement gained here / the performance improvements are somewhere else entirely and orders of magnitude bigger. and we loose consistency, which can cause subtle bugs. so in this case i decided against performance so that the code is easier to maintain

umh-core/pkg/service/protocolconverter/protocolconverter.go
JeremyTheocharis: good point, fixed that in that PR

umh-core/pkg/config/config.go
JeremyTheocharis: good point, removed

umh-core/pkg/service/protocolconverter/protocolconverter.go
JeremyTheocharis: good point, removed

umh-core/pkg/service/protocolconverter/protocolconverter.go
led0nk: okay i'm fine with that, but even though this is separated, it's still just a wrapper for the other function, so we could get rid of one

led0nk (commented):
looks good, but would make sense to merge #2060 into this now if you don't want to merge it 1 by 1 into staging

PR #2030, opened 2025-05-12
ENG-2952 Allow modifcation of existing kafka settings

Scarjit:
This pr uses a REST api call to change 0-N settings at the same time, which is also documented at https://docs.redpanda.com/api/admin-api/.

### DIFF
--- a/umh-core/integration/redpanda_config_update_test.go
+++ b/umh-core/integration/redpanda_config_update_test.go
@@ -24,7 +24,7 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-var _ = Describe("Redpanda Config Update Integration Test", Ordered, Label("integration"), func() {
+var _ = FDescribe("Redpanda Config Update Integration Test", Ordered, Label("integration"), func() {
 	const (
 		topicName         = "dfc-config-update-test-topic"
 		messagesPerSecond = 5
@@ -123,6 +123,19 @@ var _ = Describe("Redpanda Config Update Integration Test", Ordered, Label("inte
 			GinkgoWriter.Printf("Error: %v\n", err)
 			return err == nil && redpandaConfig == "7200000"
 		}, 5*time.Second, 1*time.Second).Should(BeTrue(), "Redpanda config should not be changed back")
+
+		// Let's check that we can remove keys as well
+		builder.full.Internal.Redpanda.RedpandaServiceConfig.Topic.DefaultTopicRetentionMs = 0
+		cfg = builder.BuildYAML()
+		Expect(writeConfigFile(cfg, getContainerName())).To(Succeed())
+
+		By("Checking if the config has been removed")
+		Eventually(func() bool {
+			redpandaConfig, err := getRedpandaConfig("log_retention_ms")
+			GinkgoWriter.Printf("Redpanda config: %s\n", redpandaConfig)
+			GinkgoWriter.Printf("Error: %v\n", err)
+			return err != nil && redpandaConfig == "604800000" // Default value
+		}, 5*time.Second, 1*time.Second).Should(BeTrue(), "Redpanda config should be removed")
 
 	})
 })
--- a/umh-core/integration/redpanda_config_update_test.go
+++ b/umh-core/integration/redpanda_config_update_test.go
@@ -24,7 +24,7 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-var _ = Describe("Redpanda Config Update Integration Test", Ordered, Label("integration"), func() {
+var _ = FDescribe("Redpanda Config Update Integration Test", Ordered, Label("integration"), func() {
 	const (
 		topicName         = "dfc-config-update-test-topic"
 		messagesPerSecond = 5
--- a/umh-core/integration/redpanda_config_update_test.go
+++ b/umh-core/integration/redpanda_config_update_test.go
@@ -124,19 +124,6 @@ var _ = FDescribe("Redpanda Config Update Integration Test", Ordered, Label("int
 			return err == nil && redpandaConfig == "7200000"
 		}, 5*time.Second, 1*time.Second).Should(BeTrue(), "Redpanda config should not be changed back")
 
-		// Let's check that we can remove keys as well
-		builder.full.Internal.Redpanda.RedpandaServiceConfig.Topic.DefaultTopicRetentionMs = 0
-		cfg = builder.BuildYAML()
-		Expect(writeConfigFile(cfg, getContainerName())).To(Succeed())
-
-		By("Checking if the config has been removed")
-		Eventually(func() bool {
-			redpandaConfig, err := getRedpandaConfig("log_retention_ms")
-			GinkgoWriter.Printf("Redpanda config: %s\n", redpandaConfig)
-			GinkgoWriter.Printf("Error: %v\n", err)
-			return err != nil && redpandaConfig == "604800000" // Default value
-		}, 5*time.Second, 1*time.Second).Should(BeTrue(), "Redpanda config should be removed")
-
 	})
 })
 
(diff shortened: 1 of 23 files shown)


umh-core/pkg/fsm/redpanda/reconcile.go
```diff
@@ -166,11 +167,42 @@ func (r *RedpandaInstance) reconcileExternalChanges(ctx context.Context, service
...
 
-	err := r.UpdateObservedStateOfInstance(observedStateCtx, services, tick, loopStartTime)
+	err = r.UpdateObservedStateOfInstance(observedStateCtx, services, tick, loopStartTime)
 	if err != nil {
-		return fmt.Errorf("failed to update observed state: %w", err)
+		return fmt.Errorf("failed to update observed state: %w", err), false
+	}
+
+	currentState := r.baseFSMInstance.GetCurrentFSMState()
+	r.baseFSMInstance.GetLogger().Infof("Current state: %s", currentState)
```
JeremyTheocharis, line 176: i think this doesnt need to be an info

umh-core/pkg/fsm/redpanda/reconcile.go
```diff
@@ -166,11 +167,42 @@ func (r *RedpandaInstance) reconcileExternalChanges(ctx context.Context, service
...
+	if IsRunningState(currentState) {
```
JeremyTheocharis, line 177: maybe add a comment here what this whole if clause is for:
- if there is a config diff AND redpanda is currently running, then apply it via HTTP. because we apply it via HTTP, we can only do it if redpanda is running

JeremyTheocharis: on the first look, i cannot see how the restart is done? maybe put this here into the comment as well

umh-core/pkg/service/redpanda/redpanda.go
```diff
@@ -956,3 +962,133 @@ func formatMemory(memory int) string {
...
+	if s.httpClient == nil {
+		return fmt.Errorf("http client not initialized")
+	}
+
+	// If the context is done, return an error
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
+
+	// Construct the request body
```
JeremyTheocharis, line 977: maybe it makes sense here to either have smaller sub-function like "UpdateCLusterCOnfig" and "ReadbackClusterConfig" or alternatively documentation about "Step1: update the cluster config", "step 2: readback",etc.

would make it easier to understand the whole prupose of the function

umh-core/pkg/service/redpanda/redpanda.go
```diff
@@ -956,3 +962,133 @@ func formatMemory(memory int) string {
...
+		return fmt.Errorf("failed to read readback response body: %w", err)
+	}
+
+	// Parse as JSON
+	var readbackConfig map[string]interface{}
+	if err := json.Unmarshal(readbackBody, &readbackConfig); err != nil {
+		return fmt.Errorf("failed to unmarshal readback response body: %w", err)
+	}
+
+	for key, value := range configUpdates {
```
JeremyTheocharis, line 1061: what is happening in this for loop?

umh-core/pkg/service/redpanda/redpanda.go
```diff
@@ -956,3 +962,133 @@ func formatMemory(memory int) string {
 
 	return fmt.Sprintf("%d%s", memory, units[unitIndex])
 }
+
+// updateRedpandaClusterConfig updates Redpanda's cluster config via HTTP PUT request
```
JeremyTheocharis, line 966: can you add into the docstring explaining why this works? we've put a lot of time and information in the research and tried different methods.

We should explain all of the methods (with restart), and as far as I remember this here is that we enforce the setting by reading it back immediately?

umh-core/pkg/fsm/redpanda/reconcile.go
JeremyTheocharis: ah, justn oticed that we dont need the restart anymore. Like I said further below, it makes sense to explain all of that reasoning here in docstrings and comments.

umh-core/pkg/service/redpanda/redpanda.go
JeremyTheocharis: other options were also:
1. do the restart by setting a EventDone
2. have this done via an S6 service with curl

but what doesn ow happen for things that DO require a restart according to redpanda? I think currently we can only change the ones without requiring a restart with the read back trick

umh-core/pkg/service/redpanda/redpanda.go
```diff
@@ -956,3 +962,133 @@ func formatMemory(memory int) string {
...
+func (s *RedpandaService) UpdateRedpandaClusterConfig(ctx context.Context, redpandaName string, configUpdates map[string]interface{}) error {
+	if s.httpClient == nil {
+		return fmt.Errorf("http client not initialized")
+	}
+
+	// If the context is done, return an error
+	if ctx.Err() != nil {
```
JeremyTheocharis, line 973: maybe it makes sense to define our own constant for "UpdateHTTPTimeout" or something similar so that we tell future developers that this only takes like xx ms maximum. helps to spot regressions as well

umh-core/pkg/service/redpanda/redpanda.go
```diff
@@ -956,3 +962,133 @@ func formatMemory(memory int) string {
...
+	}
+
+	// Convert to JSON
+	jsonBody, err := json.Marshal(requestBody)
+	if err != nil {
+		return fmt.Errorf("failed to marshal request body: %w", err)
+	}
+
+	// Construct the URL
+	url := "http://localhost:9644/v1/cluster_config"
```
JeremyTheocharis, line 990: maybe it makes sense to create this somehow dynamically or use a const for this? or use the same generation how we fetch the config, to show that this is the same API call

JeremyTheocharis (changes requested):

umh-core/integration/redpanda_config_update_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+func getRedpandaConfig(key string) (string, error) {
+	ctx, cncl := context.WithTimeout(context.Background(), 1*time.Second)
+	defer cncl()
+	out, err := runDockerCommandWithCtx(ctx, "exec", getContainerName(), "/opt/redpanda/bin/rpk", "cluster", "config", "get", key)
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(out), nil
+}
```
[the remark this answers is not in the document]
Scarjit: It seems to work reliably.

FreiPaul (approved):

umh-core/pkg/fsm/redpanda/reconcile.go
```diff
@@ -91,7 +91,8 @@ func (r *RedpandaInstance) Reconcile(ctx context.Context, snapshot fsm.SystemSna
 	}
 
 	// Step 2: Detect external changes.
-	if err := r.reconcileExternalChanges(ctx, services, snapshot.Tick, start); err != nil {
+	err, reconciled = r.reconcileExternalChanges(ctx, services, snapshot.Tick, start)
```
JeremyTheocharis, line 94: what is this change for?

JeremyTheocharis: it will be overrriden by `err, reconciled = r.reconcileStateTransition(ctx, services, currentTime)`

do a reconciledExternal and reconciledStateTransition and then a || if you want to do it

umh-core/pkg/service/redpanda/redpanda.go
```diff
@@ -956,3 +962,181 @@ func formatMemory(memory int) string {
...
+}
+
+// UpdateRedpandaClusterConfig updates Redpanda's cluster configuration through its admin API.
+// The function performs two key operations:
+//  1. Sends a PUT request to update the cluster configuration
+//  2. Verifies the changes by reading back and comparing the new values
+//
+// Note on restarts:
+//   - This function does not directly trigger a Redpanda restart
+//   - Restarts are handled by the reconcile loop when S6 service config changes
```
JeremyTheocharis, line 1112: what does happen first? the s6 service change, or the API call? ideally it should always do the API call first, but could there also be cases where redpanda is restarted and then the API call is attempted and fails?

JeremyTheocharis (changes requested):

umh-core/pkg/service/redpanda/redpanda.go
Scarjit: good catch

JeremyTheocharis (approved):

PR #2040, opened 2025-05-13
Eng 2977 expose config.yaml via api

FreiPaul:


umh-core/go.sum
```diff
@@ -160,11 +215,28 @@ golang.org/x/xerrors v0.0.0-20190717185122-a985d3407aa7/go.mod h1:I/5z698sn9Ka8T
...
 google.golang.org/protobuf v1.36.6/go.mod h1:jduwjTPXsFjZGTmRluh+L6NjiWu7pchiJ2/5YcXBHnY=
 gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
 gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c h1:Hei/4ADfdWqJk1ZMxUNpqntNwaWcugrBjAiHlqqRiVk=
 gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c/go.mod h1:JHkPIbrfpd72SG/EVd6muEfDQjcINNoR0C8j2r3qZ4Q=
+gopkg.in/yaml.v2 v2.4.0 h1:D8xgwECY7CYvx+Y2n4sBz93Jn9JRvxdiyyo8CTfuKaY=
+gopkg.in/yaml.v2 v2.4.0/go.mod h1:RDklbk79AGWmwhnvt/jBztapEOGDOx6ZbXqjP6csGnQ=
 gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c/go.mod h1:K4uyk7z7BCEPqu6E+C64Yfv1cQ7kz7rIZviUmN+EgEM=
 gopkg.in/yaml.v3 v3.0.1 h1:fxVm/GzAzEWqLHuvctI91KS9hhNmmWOoWu0XTYJS7CA=
 gopkg.in/yaml.v3 v3.0.1/go.mod h1:K4uyk7z7BCEPqu6E+C64Yfv1cQ7kz7rIZviUmN+EgEM=
+k8s.io/apimachinery v0.32.2 h1:yoQBR9ZGkA6Rgmhbp/yuT9/g+4lxtsGYwW6dR6BDPLQ=
```
Scarjit, line 234: Why are there so many new dependencies in here ?
incl. ones for k8s ?

Scarjit (changes requested):

Scarjit (approved):

umh-core/pkg/communicator/actions/actions.go
```diff
@@ -129,6 +129,26 @@ func HandleActionMessage(instanceUUID uuid.UUID, payload models.ActionMessagePay
 			actionLogger:          log,
 			systemSnapshotManager: systemSnapshotManager,
 		}
+
+	case models.GetConfigFile:
```
kanapuli, line 133: (style) Any special reasons why we differ here from other Handlers by calling a constructor instead of the struct Assignment directly ?

umh-core/pkg/communicator/actions/testutil/memfileinfo.go
```diff
@@ -0,0 +1,36 @@
...
+// limitations under the License.
+
+package testutil
+
+import (
+	"os"
+	"time"
+)
+
+// MemFileInfo is a simple implementation of os.FileInfo for testing
```
kanapuli, line 22: Probably the os.FileInfo should expose an interface and the interface could be used in the tests directly for mocking. 
While I agree with the current implementation, I should also point out that the struct MemFileInfo never indicates that it is intended for testing purposes only.

umh-core/pkg/communicator/actions/get-config-file.go
```diff
@@ -0,0 +1,129 @@
...
+
+	// ─── Runtime observation ────────────────────────────────────────────────
+	systemSnapshotManager *fsm.SnapshotManager
+
+	// ─── Utilities ──────────────────────────────────────────────────────────
+	actionLogger *zap.SugaredLogger
+}
+
+// NewGetConfigFileAction creates a new GetConfigFileAction with the provided parameters.
+func NewGetConfigFileAction(userEmail string, actionUUID uuid.UUID, instanceUUID uuid.UUID, outboundChannel chan *models.UMHMessage, systemSnapshotManager *fsm.SnapshotManager, configManager config.ConfigManager) *GetConfigFileAction {
```
kanapuli, line 52: (minor) This function never does any additional computation other than initializing the struct. Maybe we can remove this function and directly initialize the exported struct wherever necessary

umh-core/pkg/communicator/actions/set-config-file.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// Parse extracts the business fields from the raw JSON payload.
+func (a *SetConfigFileAction) Parse(payload interface{}) error {
+	a.actionLogger.Info("Parsing SetConfigFile payload")
+
+	// Extract SetConfigFilePayload from the interface{}
+	payloadStruct, err := ParseActionPayload[models.SetConfigFilePayload](payload)
+	if err != nil {
+		SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionFinishedWithFailure,
```
kanapuli, line 75: (style)Better to split the arguments in multiple lines for readability

umh-core/pkg/communicator/actions/set-config-file.go
```diff
@@ -0,0 +1,172 @@
...
+
+	return nil
+}
+
+// Execute takes care of updating the config file content.
+func (a *SetConfigFileAction) Execute() (interface{}, map[string]interface{}, error) {
+	a.actionLogger.Info("Executing SetConfigFile action")
+
+	// Create a context with timeout
+	ctx, cancel := context.WithTimeout(context.Background(), constants.ConfigGetConfigTimeout)
```
kanapuli, line 106: (major) Is it `ConfigGetConfigTimeout` or `ConfigSetConfigTimeout` ? 

If we use the same timeout constant for both `get-config` and `set-config` operation, then the constant must be renamed for clarity

umh-core/pkg/communicator/actions/get-config-file.go
FreiPaul: Fair point. I think, it origins from initially some action structs being declared privately (?). I created ENG-3025 for it

umh-core/pkg/communicator/actions/get-config-file.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil || fileInfo == nil {
+		errMsg := fmt.Sprintf("Failed to read config file info: %v", err)
+		SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionFinishedWithFailure,
+			errMsg, a.outboundChannel, models.GetConfigFile)
+		return nil, nil, fmt.Errorf("failed to read config file info: %w", err)
+	}
+	lastModifiedTime := fileInfo.ModTime()
+
+	// Read the file content
+	data, err := a.fsService.ReadFile(ctx, configPath)
```
JeremyTheocharis, line 106: can't we just use the config manager instead? and also to write back? then we don't use the filesystem directly, but instead the abstraction of the config manager

JeremyTheocharis: then we also have the mutex etc and we dont write into the file while another function is reading it

JeremyTheocharis: the config manager has also a cacheTime which is already the ModTime

JeremyTheocharis (changes requested):

umh-core/pkg/communicator/actions/testutil/memfileinfo.go
JeremyTheocharis, line 1: is this file still required?

umh-core/pkg/config/manager.go
```diff
@@ -686,3 +708,64 @@ func (m *FileConfigManagerWithBackoff) AtomicEditDataflowcomponent(ctx context.C
 
 	return m.configManager.AtomicEditDataflowcomponent(ctx, componentUUID, dfc)
 }
+
+// GetConfigAsString returns the current config file contents as a string
```
JeremyTheocharis, line 712: can you explain why this function exists? so explain with yaml anchors, etc.

umh-core/pkg/config/manager.go
```diff
@@ -686,3 +708,64 @@ func (m *FileConfigManagerWithBackoff) AtomicEditDataflowcomponent(ctx context.C
...
+func (m *FileConfigManager) GetConfigAsString(ctx context.Context) (string, error) {
+	// in the GetConfig method, we already read the file and cached the raw config to m.cacheRawConfig
+	_, err := m.GetConfig(ctx, 0)
```
JeremyTheocharis, line 715: i don't understand this, why are we doing a GetConfig and then reading from the cache? that doesn't feel right, even if this works. I recommend splitting it then up into functions that mutate state, and then pure functions (as one example). Otherwise it is unexpected.

umh-core/pkg/config/manager.go
```diff
@@ -686,3 +708,64 @@ func (m *FileConfigManagerWithBackoff) AtomicEditDataflowcomponent(ctx context.C
...
+	if err != nil {
+		return "", fmt.Errorf("failed to get config: %w", err)
+	}
+
+	return m.cacheRawConfig, nil
+}
+
+// GetConfigAsString returns the current config as a string with backoff logic for failures
```
JeremyTheocharis, line 723: see above

umh-core/pkg/config/manager.go
```diff
@@ -686,3 +708,64 @@ func (m *FileConfigManagerWithBackoff) AtomicEditDataflowcomponent(ctx context.C
...
+	// in the GetConfig method, we already read the file and cached the raw config to m.cacheRawConfig
+	_, err := m.GetConfig(ctx, 0)
+	if err != nil {
+		return "", fmt.Errorf("failed to get config: %w", err)
+	}
+
+	return m.configManager.cacheRawConfig, nil
+}
+
+// GetConfigModTime returns the modification time of the config file
```
JeremyTheocharis, line 734: for what? can you write the reason why this might be important?

umh-core/pkg/config/manager.go
```diff
@@ -686,3 +708,64 @@ func (m *FileConfigManagerWithBackoff) AtomicEditDataflowcomponent(ctx context.C
...
+func (m *FileConfigManager) GetCacheModTime() (time.Time, error) {
+	// read config to update the cache mod time
+	_, err := m.GetConfig(context.Background(), 0)
```
JeremyTheocharis, line 737: can this function maybe take a context instead? then we can just pass it down here

umh-core/pkg/config/manager.go
```diff
@@ -686,3 +708,64 @@ func (m *FileConfigManagerWithBackoff) AtomicEditDataflowcomponent(ctx context.C
...
+	if err != nil {
+		return time.Time{}, fmt.Errorf("failed to get config: %w", err)
+	}
+
+	m.cacheMu.RLock()
```
JeremyTheocharis, line 742: are you sure it works like this? maybe to it more explicit

umh-core/pkg/config/manager.go
```diff
@@ -686,3 +708,64 @@ func (m *FileConfigManagerWithBackoff) AtomicEditDataflowcomponent(ctx context.C
...
+
+// GetConfigAsString returns the current config file contents as a string
+func (m *FileConfigManager) GetConfigAsString(ctx context.Context) (string, error) {
+	// in the GetConfig method, we already read the file and cached the raw config to m.cacheRawConfig
+	_, err := m.GetConfig(ctx, 0)
+	if err != nil {
+		return "", fmt.Errorf("failed to get config: %w", err)
+	}
+
+	return m.cacheRawConfig, nil
```
JeremyTheocharis, line 720: so cache mod time is protected by a mutex, but this function isnt?

umh-core/pkg/config/manager.go
```diff
@@ -686,3 +708,64 @@ func (m *FileConfigManagerWithBackoff) AtomicEditDataflowcomponent(ctx context.C
...
+	return m.cacheModTime, nil
+}
+
+// GetCacheModTime delegates to the underlying FileConfigManager
+func (m *FileConfigManagerWithBackoff) GetCacheModTime() (time.Time, error) {
+	return m.configManager.GetCacheModTime()
+}
+
+// WriteConfigFromSting writes a config from a string to the config file
+func (m *FileConfigManager) WriteConfigFromSting(ctx context.Context, config string) error {
```
JeremyTheocharis, line 753: typo

JeremyTheocharis: WriteConfigFromStRing

umh-core/pkg/constants/actions.go
```diff
@@ -23,5 +23,14 @@ const (
 )
 
 const (
+	// ActionTickerTime is the time between the ticks of the action ticker
+	// especially relevant for deploy-dataflow-component or edit-dataflow-component
+	// becuase they have a ticker loop to check the status of the dataflow component
```
JeremyTheocharis, line 28: typo

JeremyTheocharis: becuase

umh-core/pkg/communicator/actions/set-config-file_test.go
```diff
@@ -0,0 +1,232 @@
...
+	"github.com/google/uuid"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/communicator/actions"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/config"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsm"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/models"
+)
+
+// memFileInfo already defined in get-config-file_test.go
```
JeremyTheocharis, line 32: what is this comment for?

umh-core/pkg/communicator/actions/set-config-file_test.go
```diff
@@ -0,0 +1,232 @@
...
+
+var _ = Describe("SetConfigFile", func() {
```
JeremyTheocharis, line 34: can you add a test to check if the YAML anchors work properly? so that they are not marshalled/interpreted and that tey are shown properly in the UI as well as written back to the UI

JeremyTheocharis (changes requested):

JeremyTheocharis (changes requested):
Here from o3:

<html><head></head><body><h3>🧩 Fresh findings (things I didn’t spot in earlier review comments)</h3>

Area | Observation | Why it matters | Suggestion
-- | -- | -- | --
Timeouts | GetOrSetConfigFileTimeout = 30 ms – both actions create a context.WithTimeout using that value. Even on a local SSD, a call-chain of yaml parse → mutex acquisition → file-read will frequently blow past 30 ms; on a network volume it’s almost guaranteed. | Intermittent “context deadline exceeded” under moderate I/O pressure. | Pick a value that’s slow-disk tolerant (e.g. 1 s) or pass the caller’s deadline down instead of inventing a new one.
String-vs-time comparison | SetConfigFileAction.Execute() string-compares currentLastModifiedTimeString with the client-supplied timestamp. RFC-3339 round-tripping is not guaranteed to preserve sub-second precision (the file’s mtime might include monotonic‐clock data that gets stripped) and local/UTC formatting differences will fail the equality test. | False-positive “concurrent modification detected” → user can’t save even though nothing changed. | Parse payload.LastModifiedTime with time.Parse and use time.Equal() or t1.Sub(t2) < ε instead of string equality.
Missing validation of timestamp format | Validate() checks that LastModifiedTime != "" but never parses it, so an arbitrary string (“foo”) passes validation and then fails inside Execute(). | Pushes error handling out of the happy path and into the action’s core logic. | In Validate() parse the string into a time.Time; store the parsed value on the struct so Execute() can reuse it.
Locking of cacheRawConfig | FileConfigManager.GetConfig() writes m.cacheRawConfig = string(data) while holding the read-lock, but later reads of that field (GetConfigAsString(), GetConfigAsString in the back-off wrapper) access it without any lock. | Two goroutines can read from and assign to the same string concurrently – the assignment itself is atomic, but not the visibility of side-effects; a race detector run will flag this. | Guard cacheRawConfig with the same cacheMu used for cacheConfig/cacheModTime or make it part of an immutable snapshot struct that you swap under lock.
Anchor-preservation asymmetry | Reading path keeps raw YAML intact (good), but the write path does yaml.Unmarshal ➔ struct ➔ writeConfig() ➔ yaml.Marshal (inside writeConfig()), which expands aliases and drops anchors. A user can therefore open a file containing anchors, press Save without changes, and lose those anchors. | Inadvertent config drift; anchors are often used to DRY-up large UMH configs. | Consider a “verbatim write” branch that skips parse-and-re-marshal when the client submits a blob that passed YAML linting and passes concurrency checks.
Duplicate helper code | New helpers (GetConfigAsString, WriteConfigFromSting, etc.) are re-implemented in both FileConfigManager and FileConfigManagerWithBackoff. | Divergence risk – one path gets a bug-fix, the other doesn’t. | Factor the shared bits into free functions or embed one struct in the other.
Large responses over the action bus | Both actions echo the full config file back to the caller (esp. SetConfigFileResponse). For multi-MB configs that doubles network traffic and clogs the AMQP channel, yet the client already owns the content it just sent. | Latency + bandwidth. | Return only metadata (lastModifiedTime, maybe a short diff summary). The UI can keep the editor buffer as the source of truth.
Hard-coded path in log message | GetConfigFileAction logs Reading config file from <defaultPath> but the concrete path could differ when a custom ConfigManager implementation is injected. | Misleading operator logs. | Ask the configManager for its actual path (expose it on the interface) or just log “reading config file” without the path.
Unit-tests miss race-conditions | Tests cover happy-path & error cases, but none run under -race, and none exercise concurrent reads/writes. | Risk that the lock gap on cacheRawConfig (see above) slips through CI. | Add a test that spawns N goroutines calling GetConfigAsString in a loop while another goroutine rewrites the config; run with go test -race.


<hr>
<p><strong>Tiny nits I also didn’t see mentioned</strong></p>
<ul>
<li>
<p><code inline="">SetConfigFileAction.Validate()</code> does two passes over the YAML (validate + later parse-and-marshal). You can reuse the <code inline="">yaml.Node</code> produced during validation to avoid parsing twice.</p>
</li>
<li>
<p><code inline="">WriteConfigFromSting</code> → typo already caught; but note it’s exported (should probably stay internal or be renamed to <code inline="">WriteConfigFromString</code> + <code inline="">//nolint:revive</code> to avoid the stutter).</p>
</li>
</ul>
<p>Hope that helps tighten things up!</p></body></html>

umh-core/pkg/communicator/actions/set-config-file_test.go
FreiPaul: Currently this only works for the get action because I added a yaml validation for the set action (parseConfig) and it currently does not accept the anchors

JeremyTheocharis: can we then fix it so that the set function also works?

PR #2045, opened 2025-05-14
Enhance Benthos health check to only proceed once the health is really passing for 5 seconds

JeremyTheocharis:
In this PR I tried several options, until I landed up with a debounce logic based on ticks:
1. first I added a logic based on the counters. then I realized that the original counter based logic is not correct, so I adjusted it. Then upon researching I realized it is measuring the same as ready/live in the first place
2. so i added instead a debouncing logic that when it turns to ready for a brief second, but then fails, the fsm doesnt proceed. this was first done with time, and I adjusted all the according tests to use currentTime as well. I made the entire logic so that it uses the time from the snapshot, therefore preventing another time.now call
3. then I realized that ticks are more stable and easier for testing, and I reverted the time change but kept the reduction of one time.now call


umh-core/pkg/fsm/benthos/actions.go
```diff
@@ -312,18 +312,30 @@ func (b *BenthosInstance) IsBenthosConfigLoaded() (bool, string) {
...
 //
 // It returns:
 //
 //	ok     – true when both probes pass, false otherwise.
 //	reason – empty when ok is true; otherwise details of the failed probe(s).
 func (b *BenthosInstance) IsBenthosHealthchecksPassed() (bool, string) {
+	connectionUpCounterInput := b.ObservedState.ServiceInfo.BenthosStatus.BenthosMetrics.Metrics.Input.ConnectionUp
+	connectionFailedCounterInput := b.ObservedState.ServiceInfo.BenthosStatus.BenthosMetrics.Metrics.Input.ConnectionFailed
+	connectionLostCounterInput := b.ObservedState.ServiceInfo.BenthosStatus.BenthosMetrics.Metrics.Input.ConnectionLost
+	isInputUp := connectionUpCounterInput > (connectionFailedCounterInput + connectionLostCounterInput)
```
FreiPaul, line 325: Maybe this is worth explaining or at least reference https://docs.redpanda.com/redpanda-connect/components/metrics/about/#inputs so that the code reader can understand what is happening

umh-core/pkg/fsm/benthos/actions.go
```diff
@@ -312,18 +312,30 @@ func (b *BenthosInstance) IsBenthosConfigLoaded() (bool, string) {
...
+
+	connectionUpCounterOutput := b.ObservedState.ServiceInfo.BenthosStatus.BenthosMetrics.Metrics.Output.ConnectionUp
+	connectionFailedCounterOutput := b.ObservedState.ServiceInfo.BenthosStatus.BenthosMetrics.Metrics.Output.ConnectionFailed
+	connectionLostCounterOutput := b.ObservedState.ServiceInfo.BenthosStatus.BenthosMetrics.Metrics.Output.ConnectionLost
+	isOutputUp := connectionUpCounterOutput > (connectionFailedCounterOutput + connectionLostCounterOutput)
```
FreiPaul, line 330: same here

umh-core/pkg/fsm/benthos/actions.go
```diff
@@ -312,18 +312,30 @@ func (b *BenthosInstance) IsBenthosConfigLoaded() (bool, string) {
...
 // It returns:
 //
 //	ok     – true when both probes pass, false otherwise.
 //	reason – empty when ok is true; otherwise details of the failed probe(s).
 func (b *BenthosInstance) IsBenthosHealthchecksPassed() (bool, string) {
+	connectionUpCounterInput := b.ObservedState.ServiceInfo.BenthosStatus.BenthosMetrics.Metrics.Input.ConnectionUp
+	connectionFailedCounterInput := b.ObservedState.ServiceInfo.BenthosStatus.BenthosMetrics.Metrics.Input.ConnectionFailed
+	connectionLostCounterInput := b.ObservedState.ServiceInfo.BenthosStatus.BenthosMetrics.Metrics.Input.ConnectionLost
+	isInputUp := connectionUpCounterInput > (connectionFailedCounterInput + connectionLostCounterInput)
+
```
FreiPaul, line 326: What happens when a connection attempt fails in the first run but then the connection goes up. Then we have the case
`connectionFailedCounterInput=1`
`connectionUpCounterInput=1`
which should result in `isInputUp=true`
,right?
With the `>` check this is not the case

FreiPaul (changes requested):

umh-core/pkg/constants/benthos.go
```diff
@@ -53,4 +53,24 @@ const (
...
+	// A pod can therefore oscillate like:
+	//
+	//   live=true, ready=true   ← connection succeeds
+	//   live=true, ready=false  ← broker drops a socket a few ms later
+	//
+	// Our FSM used to consume IsReady verbatim, so a 1-frame “true” was enough to
+	// advance the state machine.
+	//
+	// Change this constant if you need a different stability window.
+	BenthosHealthCheckStableDurationInTicks = uint64(5 * time.Second / DefaultTickerTime) // 5 seconds
```
Scarjit, line 75: BenthosHealthCheckStableDurationInTicks is 50 ticks, which ofc is then equivalent to 5 sec

PR #2049, opened 2025-05-15
[ENG-3022] Fix YAML Inject Handling

FreiPaul:
The problem was that the frontend sends the yaml inject part as `rawYAML` and not as `inject`. The tests have been updated and the parsing is fixed now


Scarjit (commented):
Does this require changes to the Get action aswell ?

FreiPaul:
In the get action, we already use `CommonDataFlowComponentCDFCProperties` which has the correct `rawYAML` field

PR #2051, opened 2025-05-15
[ENG-3007] [ENG-3014] Better Errors for DFC deploy and edit

FreiPaul:
This PR introduces early aborts on benthos config errors and provides the frontend with better error info (i.e. error code) so that it can determine weather to offer the user the retry-option or not. Therefore, we include the SendActionReplyV2 which already existed in UMH Classic. In the frontend, the fetcher checks the action reply error code for the prefix `ERR_RETRY_`  . If it finds the prefix in the error code or no error code at all, the retry option is displayed.

### DIFF
--- a/umh-core/pkg/communicator/actions/benthos-log-checker.go
+++ b/umh-core/pkg/communicator/actions/benthos-log-checker.go
@@ -29,6 +29,10 @@ var ErrBenthosLogLines = []string{
 // Benthos configuration errors. It performs case-insensitive checks for known
 // error patterns combined with the word "error". Returns true if any errors
 // are detected, false otherwise.
+// This function is used to detect fatal configuration errors that would cause
+// Benthos to enter a CrashLoop. When such errors are detected, we can immediately
+// abort the startup process rather than waiting for the full timeout period,
+// as these errors require configuration changes to resolve.
 func CheckBenthosLogLinesForConfigErrors(logs []s6.LogEntry) bool {
 	for _, log := range logs {
 		logContent := strings.ToLower(log.Content)
--- a/umh-core/pkg/communicator/actions/benthos-log-checker.go
+++ b/umh-core/pkg/communicator/actions/benthos-log-checker.go
@@ -1,0 +1,37 @@
+// Copyright 2025 UMH Systems GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package actions
+
+import (
+	"strings"
+
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/service/s6"
+)
+
+var ErrBenthosLogLines = []string{
+	"unable to infer",
+	"Config lint error",
+}
+
+func CheckBenthosLogLinesForConfigErrors(logs []s6.LogEntry) bool {
+	for _, log := range logs {
+		for _, errLine := range ErrBenthosLogLines {
+			if strings.Contains(log.Content, errLine) && strings.Contains(log.Content, "error") {
+				return true
+			}
+		}
+	}
+	return false
+}
--- a/umh-core/pkg/communicator/actions/benthos-log-checker.go
+++ b/umh-core/pkg/communicator/actions/benthos-log-checker.go
@@ -25,10 +25,19 @@ var ErrBenthosLogLines = []string{
 	"Config lint error",
 }
 
+// CheckBenthosLogLinesForConfigErrors examines a slice of S6 log entries for
+// Benthos configuration errors. It performs case-insensitive checks for known
+// error patterns combined with the word "error". Returns true if any errors
+// are detected, false otherwise.
 func CheckBenthosLogLinesForConfigErrors(logs []s6.LogEntry) bool {
 	for _, log := range logs {
+		logContent := strings.ToLower(log.Content)
+		if !strings.Contains(logContent, "error") {
+			continue
+		}
+
 		for _, errLine := range ErrBenthosLogLines {
-			if strings.Contains(log.Content, errLine) && strings.Contains(log.Content, "error") {
+			if strings.Contains(logContent, strings.ToLower(errLine)) {
 				return true
 			}
 		}
--- a/umh-core/pkg/communicator/actions/deploy-dataflowcomponent.go
+++ b/umh-core/pkg/communicator/actions/deploy-dataflowcomponent.go
@@ -532,7 +532,7 @@ func (a *DeployDataflowComponentAction) Execute() (interface{}, map[string]inter
 	}
 
 	// return success message, but do not send it as this is done by the caller
-	successMsg := Label("deploy", a.name) + "success"
+	successMsg := Label("deploy", a.name) + "component successfully deployed and activated"
 
 	return successMsg, nil, nil
 }
@@ -592,8 +592,9 @@ func (a *DeployDataflowComponentAction) waitForComponentToBeActive(ctx context.C
 			err := a.configManager.AtomicDeleteDataflowcomponent(ctx, dataflowcomponentserviceconfig.GenerateUUIDFromName(a.name))
 			if err != nil {
 				a.actionLogger.Errorf("failed to remove dataflowcomponent %s: %v", a.name, err)
+				return models.ErrRetryRollbackTimeout, fmt.Errorf("dataflow component '%s' failed to activate within timeout but could not be removed: %v. Please check system load and consider removing the component manually", a.name, err)
 			}
-			return models.ErrRetryRollbackTimeout, fmt.Errorf("dataflow component '%s' was removed because it did not become active within the timeout period", a.name)
+			return models.ErrRetryRollbackTimeout, fmt.Errorf("dataflow component '%s' was removed because it did not become active within the timeout period. Please check system load or component configuration and try again", a.name)
 
 		case <-ticker.C:
 
@@ -618,7 +619,7 @@ func (a *DeployDataflowComponentAction) waitForComponentToBeActive(ctx context.C
 						continue
 					}
 					if instance.CurrentState == "active" || instance.CurrentState == "idle" {
-						stateMessage := RemainingPrefixSec(remainingSeconds) + fmt.Sprintf("completed. is in state '%s' with correct configuration", instance.CurrentState)
+						stateMessage := RemainingPrefixSec(remainingSeconds) + fmt.Sprintf("component successfully activated with state '%s', configuration verified", instance.CurrentState)
 						SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionExecuting, stateMessage,
 							a.outboundChannel, models.DeployDataFlowComponent)
 						return "", nil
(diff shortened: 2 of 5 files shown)


umh-core/pkg/communicator/actions/benthos-log-checker.go
```diff
@@ -0,0 +1,46 @@
...
+var ErrBenthosLogLines = []string{
+	"unable to infer",
+	"Config lint error",
+}
+
+// CheckBenthosLogLinesForConfigErrors examines a slice of S6 log entries for
+// Benthos configuration errors. It performs case-insensitive checks for known
+// error patterns combined with the word "error". Returns true if any errors
+// are detected, false otherwise.
+func CheckBenthosLogLinesForConfigErrors(logs []s6.LogEntry) bool {
```
JeremyTheocharis, line 32: Can you explain here why this function is useful? 

--> When benthos starts up, it can land up in a CrashLoop because of a configuration error. THese manifest through ErrBenthosLogLines. So instead of forcing the user to wait for 20 more seconds, we use this function to immediately abort because if one of these issues appears it will require a configuration change

umh-core/pkg/communicator/actions/deploy-dataflowcomponent.go
```diff
@@ -519,10 +519,10 @@ func (a *DeployDataflowComponentAction) Execute() (interface{}, map[string]inter
 			SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionExecuting, Label("deploy", a.name)+"configuration updated; but ignoring the health check", a.outboundChannel, models.EditDataFlowComponent)
 		} else {
 			SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionExecuting, Label("deploy", a.name)+"configuration updated; waiting to become active", a.outboundChannel, models.DeployDataFlowComponent)
-			err = a.waitForComponentToBeActive()
+			errCode, err := a.waitForComponentToBeActive()
 			if err != nil {
 				errorMsg := Label("deploy", a.name) + fmt.Sprintf("failed to wait for dataflow component to be active: %v", err)
-				SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionFinishedWithFailure, errorMsg, a.outboundChannel, models.DeployDataFlowComponent)
+				SendActionReplyV2(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionFinishedWithFailure, errorMsg, errCode, nil, a.outboundChannel, models.DeployDataFlowComponent, nil)
```
JeremyTheocharis, line 525: add a note here why we can and should use V2 here

umh-core/pkg/communicator/actions/deploy-dataflowcomponent.go
```diff
@@ -551,7 +551,7 @@ func (a *DeployDataflowComponentAction) GetParsedPayload() models.CDFCPayload {
 
 // waitForComponentToBeActive polls live FSM state until the new component
 // becomes active or the timeout hits (→ delete unless ignoreHealthCheck).
```
JeremyTheocharis, line 553: can you explain what string and error in the return parametrers mean? I am a little bit confused as both look like errors

umh-core/pkg/communicator/actions/deploy-dataflowcomponent.go
```diff
@@ -615,20 +615,36 @@ func (a *DeployDataflowComponentAction) waitForComponentToBeActive() error {
...
+					stateMessage := RemainingPrefixSec(remainingSeconds) + currentStateReason
+					SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionExecuting, stateMessage, a.outboundChannel, models.DeployDataFlowComponent)
+					// send the benthos logs to the user
+					logs = dfcSnapshot.ServiceInfo.BenthosObservedState.ServiceInfo.BenthosStatus.BenthosLogs
+
+					// only send the logs that have not been sent yet
+					if len(logs) > len(lastLogs) {
+						lastLogs = SendLimitedLogs(logs, lastLogs, a.instanceUUID, a.userEmail, a.actionUUID, a.outboundChannel, models.DeployDataFlowComponent, remainingSeconds)
+					}
+					// check if the logs contain any of the error lines and if so, cancel the action with rolling back
```
JeremyTheocharis, line 633: explain here again why this is relevant (see my comments at CheckBenthosLogLinesForConfigErrors)

umh-core/pkg/communicator/actions/deploy-dataflowcomponent.go
```diff
@@ -615,20 +615,36 @@ func (a *DeployDataflowComponentAction) waitForComponentToBeActive() error {
...
+					if CheckBenthosLogLinesForConfigErrors(logs) {
+						SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionExecuting, "Failed to parse config. Removing the component.", a.outboundChannel, models.DeployDataFlowComponent)
+						ctx, cancel := context.WithTimeout(context.Background(), constants.ActionTimeout)
```
JeremyTheocharis, line 636: can't we use the overall context here? in general, we should have one context, and re-use that. context.Background should be used only at the very start of something (e.g., the user starts the overall action) and then during that action run re-used

umh-core/pkg/communicator/actions/deploy-dataflowcomponent.go
```diff
@@ -615,20 +615,36 @@ func (a *DeployDataflowComponentAction) waitForComponentToBeActive() error {
...
+						defer cancel()
+						err := a.configManager.AtomicDeleteDataflowcomponent(ctx, dataflowcomponentserviceconfig.GenerateUUIDFromName(a.name))
+						if err != nil {
+							a.actionLogger.Errorf("failed to remove dataflowcomponent %s: %v", a.name, err)
+							SendActionReplyV2(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionFinishedWithFailure, "DFC not removed. Please check your configuration and consider removing the component manually.", "ERR_CONFIG_ERROR", nil, a.outboundChannel, models.DeployDataFlowComponent, nil)
```
JeremyTheocharis, line 641: what does ERR_CONFIG_ERROR mean? what does it do? --> needs to be documented somewhere. either in SendActionReplyV2 or somewhere else

umh-core/pkg/communicator/actions/edit-dataflowcomponent.go
```diff
@@ -407,7 +407,7 @@ func (a *EditDataflowComponentAction) Execute() (interface{}, map[string]interfa
 				err := yaml.Unmarshal([]byte(processor.Data), &procConfig)
 				if err != nil {
 					errMsg := Label("edit", a.name) + fmt.Sprintf("failed to parse pipeline processor %s: %s", processorName, err.Error())
-					SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionFinishedWithFailure, errMsg, a.outboundChannel, models.EditDataFlowComponent)
+					SendActionReplyV2(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionFinishedWithFailure, errMsg, "ERR_RETRY_DFC_TIMEOUT", nil, a.outboundChannel, models.EditDataFlowComponent, nil)
```
JeremyTheocharis, line 410: see comments above

umh-core/pkg/communicator/actions/edit-dataflowcomponent.go
```diff
@@ -477,10 +477,10 @@ func (a *EditDataflowComponentAction) Execute() (interface{}, map[string]interfa
 			SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionExecuting, Label("edit", a.name)+"configuration updated; but ignoring the health check", a.outboundChannel, models.EditDataFlowComponent)
 		} else {
 			SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionExecuting, Label("edit", a.name)+"configuration updated; waiting to become active", a.outboundChannel, models.EditDataFlowComponent)
-			err = a.waitForComponentToBeActive()
+			errCode, err := a.waitForComponentToBeActive()
 			if err != nil {
 				errorMsg := Label("edit", a.name) + fmt.Sprintf("failed to wait for dataflow component to be active: %v", err)
-				SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionFinishedWithFailure, errorMsg, a.outboundChannel, models.EditDataFlowComponent)
+				SendActionReplyV2(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionFinishedWithFailure, errorMsg, errCode, nil, a.outboundChannel, models.EditDataFlowComponent, nil)
```
JeremyTheocharis, line 483: see comments above

umh-core/pkg/communicator/actions/edit-dataflowcomponent.go
```diff
@@ -519,7 +519,7 @@ func (a *EditDataflowComponentAction) GetComponentUUID() uuid.UUID {
 // Concurrency note: The method never writes to `systemSnapshot`; the FSM runtime
 // is the single writer.  We only take read‑locks while **copying** the full
 // snapshot to avoid holding the lock during YAML comparisons.
-func (a *EditDataflowComponentAction) waitForComponentToBeActive() error {
+func (a *EditDataflowComponentAction) waitForComponentToBeActive() (string, error) {
```
JeremyTheocharis, line 522: see comments above.

and by the way: doesn't it make sense to reuse this function in the first place? it is same for deploy and editing right?

umh-core/pkg/models/action_models.go
```diff
@@ -531,3 +531,63 @@ type GetLogsResponse struct {
...
+// 	"action-confirmed",
+// 	"action-executing",
+// 	"action-success",
+// 	"action-failure",
+// }
+
+type ActionReplyResponseSchemaJsonActionReplyPayloadV2Payload map[string]interface{}
+
+const (
+	ErrRetryParseFailed     = "ERR_RETRY_ACTION_PARSE_FAILED"
```
JeremyTheocharis, line 584: can you explain them and what they will do in the frontend?

JeremyTheocharis (changes requested):

umh-core/pkg/communicator/actions/edit-dataflowcomponent.go
FreiPaul: I also thought about that but there are some key differences:

- The deploy version needs to delete the component on failure
- The edit version needs to roll back to the previous configuration on failure
- The edit version additionally compares observed vs desired configs

I once tried to build a more abstract shared function but in the end it made things even more complicated in my opinion. I would suggest, considering this refactor as a separate PR since I currently this PR and the pause DFC PR open for this part of the code and the refactor would now cost some time and cause merge conficts

umh-core/pkg/communicator/actions/edit-dataflowcomponent.go
```diff
@@ -477,17 +477,20 @@ func (a *EditDataflowComponentAction) Execute() (interface{}, map[string]interfa
...
 		} else {
 			SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionExecuting, Label("edit", a.name)+"configuration updated; waiting to become active", a.outboundChannel, models.EditDataFlowComponent)
-			err = a.waitForComponentToBeActive()
+			errCode, err := a.waitForComponentToBeActive(ctx)
 			if err != nil {
 				errorMsg := Label("edit", a.name) + fmt.Sprintf("failed to wait for dataflow component to be active: %v", err)
-				SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionFinishedWithFailure, errorMsg, a.outboundChannel, models.EditDataFlowComponent)
+				// waitForComponentToBeActive gives us the error code, which we then forward to the frontend using the SendActionReplyV2 function
+				// the error code is a string that can be used to identify the error reason
+				// the main reason for this is to allow the frontend to determine weather it should offer a retry option or not
```
JeremyTheocharis, line 485: * whether not weather

JeremyTheocharis (changes requested):

JeremyTheocharis (approved):

PR #2053, opened 2025-05-19
[ENG-1242] Pause DFC

FreiPaul:
belongs to https://github.com/united-manufacturing-hub/ManagementConsole/pull/2484

### DIFF
--- a/umh-core/pkg/communicator/actions/deploy-dataflowcomponent.go
+++ b/umh-core/pkg/communicator/actions/deploy-dataflowcomponent.go
@@ -234,7 +234,7 @@ func (a *DeployDataflowComponentAction) Execute() (interface{}, map[string]inter
 			SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionExecuting, Label("deploy", a.name)+"configuration updated; waiting to become ready", a.outboundChannel, models.DeployDataFlowComponent)
 			errCode, err := a.waitForComponentToBeReady(ctx)
 			if err != nil {
-				errorMsg := Label("deploy", a.name) + fmt.Sprintf("failed to wait for dataflow component to be active: %v", err)
+				errorMsg := Label("deploy", a.name) + fmt.Sprintf("failed to wait for dataflow component to be ready: %v", err)
 				// waitForComponentToBeReady gives us the error code, which we then forward to the frontend using the SendActionReplyV2 function
 				// the error code is a string that can be used to identify the error reason
 				// the main reason for this is to allow the frontend to determine whether it should offer a retry option or not
@@ -266,7 +266,7 @@ func (a *DeployDataflowComponentAction) GetParsedPayload() models.CDFCPayload {
 }
 
 // waitForComponentToBeReady polls live FSM state until the new component
-// becomes active or the timeout hits (→ delete unless ignoreHealthCheck).
+// reaches the desired state or the timeout hits (→ delete unless ignoreHealthCheck).
 // the function returns the error code and and the error message via an error object
 // the error code is a string that is sent to the frontend to allow it to determine if the action can be retried or not
 // the error message is sent to the frontend to allow the user to see the error message
@@ -274,7 +274,7 @@ func (a *DeployDataflowComponentAction) waitForComponentToBeReady(ctx context.Co
 	// checks the system snapshot
 	// 1. waits for the instance to appear in the system snapshot
 	// 2. takes the logs of the instance and sends them to the user in 1-second intervals
-	// 3. waits for the instance to be in state "active"
+	// 3. waits for the instance to reach the desired state
 	// 4. takes the residual logs of the instance and sends them to the user
 	// 5. returns nil
 
@@ -297,7 +297,7 @@ func (a *DeployDataflowComponentAction) waitForComponentToBeReady(ctx context.Co
 
 		select {
 		case <-timeout:
-			stateMessage := Label("deploy", a.name) + "timeout reached. it did not start in time. removing"
+			stateMessage := Label("deploy", a.name) + "timeout reached. it did not reach the desired state in time. removing"
 			SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionExecuting, stateMessage,
 				a.outboundChannel, models.DeployDataFlowComponent)
 			// Create a fresh context for cleanup operation since the original ctx has timed out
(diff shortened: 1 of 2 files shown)


Scarjit (approved):

umh-core/pkg/communicator/actions/deploy-dataflowcomponent.go
```diff
@@ -618,8 +627,17 @@ func (a *DeployDataflowComponentAction) waitForComponentToBeActive(ctx context.C
 							a.outboundChannel, models.DeployDataFlowComponent)
 						continue
 					}
-					if instance.CurrentState == "active" || instance.CurrentState == "idle" {
-						stateMessage := RemainingPrefixSec(remainingSeconds) + fmt.Sprintf("component successfully activated with state '%s', configuration verified", instance.CurrentState)
+					// Compare current state with the desired state
+					var acceptedStates []string
+					switch a.state {
+					case models.ComponentStateActive:
+						acceptedStates = []string{"active", "idle"}
```
JeremyTheocharis: maybe not hardcode this, but instead leverage like dataflowcomoponent.OperationalStateActive or something like that

JeremyTheocharis: maybe even list them all (even though the strings behind it are the same) --> it will make it more maintainable, should we choose the rename a state

umh-core/pkg/communicator/actions/edit-dataflowcomponent.go
```diff
@@ -184,6 +187,15 @@ func (a *EditDataflowComponentAction) Parse(payload interface{}) error {
 		return errors.New("missing required field Name")
 	}
 
+	a.state = models.ComponentState(topLevel.State)
+	if a.state == "" {
+		a.state = models.ComponentStateActive
+	}
+	if !a.state.IsValid() {
+		SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionFinishedWithFailure, "invalid state: "+string(a.state), a.outboundChannel, models.EditDataFlowComponent)
```
JeremyTheocharis: why is this not V2?

umh-core/pkg/communicator/actions/edit-dataflowcomponent.go
```diff
@@ -596,44 +610,50 @@ func (a *EditDataflowComponentAction) waitForComponentToBeActive(ctx context.Con
...
-						if instance.CurrentState == "active" || instance.CurrentState == "idle" {
-							stateMessage := RemainingPrefixSec(remainingSeconds) + fmt.Sprintf("component successfully activated with state '%s', configuration verified", instance.CurrentState)
-							SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionExecuting,
-								stateMessage, a.outboundChannel, models.EditDataFlowComponent)
-							return "", nil
+						// depending on the desired state, we accept different states
+						var acceptedStates []string
+						switch a.state {
+						case models.ComponentStateActive:
+							acceptedStates = []string{"active", "idle"}
```
JeremyTheocharis: see also above. maybe it makes sense also to combine it in a states.go subfile here in the actions folder as it is the same code as above

umh-core/pkg/communicator/actions/edit-dataflowcomponent.go
JeremyTheocharis: wait, wasnt there a logic like this:

1. send action replies within the function
2. but if it fails, do not send the action reply with failure but instead return the error (which then will be automatically translated into another action message?

i remember it was quite weird from the beginning

umh-core/pkg/models/action_models.go
```diff
@@ -598,3 +601,23 @@ const (
 	// ErrConfigFileInvalid is sent when the deployment of a dfc fails because the config file is invalid.
 	ErrConfigFileInvalid = "ERR_CONFIG_FILE_INVALID"
 )
+
+// ComponentState represents the desired state of a data flow component
+type ComponentState string
+
+const (
+	// ComponentStateActive indicates the component should be running
+	ComponentStateActive ComponentState = "active"
```
JeremyTheocharis: maybe make this dependent on dfc and protocol converter, and then re-use it

umh-core/pkg/communicator/actions/edit-dataflowcomponent.go
```diff
@@ -515,7 +530,7 @@ func (a *EditDataflowComponentAction) GetComponentUUID() uuid.UUID {
 	return a.oldComponentUUID
 }
 
-// waitForComponentToBeActive polls the live FSM state until either
+// waitForComponentToBeReady polls the live FSM state until either
 //   - the component shows up **active** with the *expected* configuration or
```
JeremyTheocharis: shows up as active or equivalent

umh-core/pkg/communicator/actions/edit-dataflowcomponent.go
```diff
@@ -184,6 +187,15 @@ func (a *EditDataflowComponentAction) Parse(payload interface{}) error {
 		return errors.New("missing required field Name")
 	}
 
+	a.state = models.ComponentState(topLevel.State)
+	if a.state == "" {
```
JeremyTheocharis: I am not sure if this should be the default error handling. if an invalid state is selected by the user, it should be refused immediately

JeremyTheocharis: also doesn't make sense that we then afterwards check if its valid, if we always set it right?

JeremyTheocharis (changes requested):

JeremyTheocharis:
bugbot run

JeremyTheocharis (changes requested):
## O3 feedback

(i removed obvious stupid stuff, the rest is worth looking into though)

```
a) DesiredFSMState propagation
You now thread DesiredFSMState: a.state through the config objects. Good. Two follow-ups:

Defaulting – In both actions you still set desiredState := a.state but a.state is always set already. The extra var isn’t needed.

Logs – Success message in Edit uses a map to translate state to “activated”/“stopped”; Deploy keeps “…successfully updated and activated” even when state=stopped. Align the two.

b) Wait-helper refactor
Renaming to waitForComponentToBeReady is clearer. However comments and some error messages still say “become active”. Search-and-replace to avoid misleading logs.
```

--> looks like quick minor changes

```
2 – Parsing & validation
go
Copy
a.state = topLevel.State
if a.state != dataflowcomponent.OperationalStateStopped &&
   a.state != dataflowcomponent.OperationalStateActive {
        return fmt.Errorf("invalid state: %s", a.state)
}
Duplication – The same validation snippet is now in both Deploy and Edit actions. Extract a helper (e.g. validateOperationalState) so the rule lives in one place.
```
--> makes sense to prevent code duplication

FreiPaul:
@coderabbitai full review

PR #2056, opened 2025-05-21
[ENG-2977] expose config.yaml via API

FreiPaul:
recovered the broken branch from https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/2040

### DIFF
--- a/umh-core/pkg/config/parse_config_templated_test.go
+++ b/umh-core/pkg/config/parse_config_templated_test.go
@@ -27,7 +27,7 @@ var _ = Describe("ParseConfigTemplated", func() {
 			cfg, err := os.ReadFile("../../examples/example-config-protocolconverter-templated.yaml")
 			Expect(err).To(BeNil())
 
-			parsedConfig, err := parseConfig(cfg)
+			parsedConfig, err := parseConfig(cfg, false)
 			Expect(err).To(BeNil())
 
 			Expect(parsedConfig.ProtocolConverter).To(HaveLen(3))


umh-core/pkg/communicator/actions/set-config-file.go
```diff
@@ -0,0 +1,177 @@
...
+		errMsg := fmt.Sprintf("Failed to write config file: %v", err)
+		SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionFinishedWithFailure,
+			errMsg, a.outboundChannel, models.SetConfigFile)
+		return nil, nil, fmt.Errorf("failed to write config file: %w", err)
+	}
+
+	newLastModifiedTime, err := a.configManager.UpdateAndGetCacheModTime(ctx)
+	if err != nil {
+		errMsg := fmt.Sprintf("Failed to get cache mod time, refresh the page and double check if the file has been modified. Consider rolling back to the previous version if issues persist. Error: %v", err)
+		SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionFinishedWithFailure,
```
JeremyTheocharis, line 155: doesnt it make sense to use the new V2 here to indicate whether this could be retried?

umh-core/pkg/communicator/actions/set-config-file.go
```diff
@@ -0,0 +1,177 @@
...
+			errMsg, a.outboundChannel, models.SetConfigFile)
+		return nil, nil, fmt.Errorf("concurrent modification detected: file modified at %v, client has version from %v",
+			currentLastModified, a.payload.LastModifiedTime)
+	}
+
+	// Write the new content to the file
+	err = a.configManager.WriteConfigFromString(ctx, a.payload.Content)
+	if err != nil {
+		errMsg := fmt.Sprintf("Failed to write config file: %v", err)
+		SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionFinishedWithFailure,
```
JeremyTheocharis, line 147: maybe use V2?

umh-core/pkg/config/mock.go
```diff
@@ -84,9 +84,6 @@ func (m *MockConfigManager) GetFileSystemService() filesystem.Service {
 
 // WriteConfig implements the ConfigManager interface
 func (m *MockConfigManager) writeConfig(ctx context.Context, cfg FullConfig) error {
-	m.mutexReadOrWrite.Lock()
-	defer m.mutexReadOrWrite.Unlock()
-
 	m.Config = cfg
```
JeremyTheocharis, line 88: why is this not protected with a mutex anymore? if this is only called from functions that already have the mutex, then this should be documented in the docstring ala "the calling functions needs to hold the mutex mutexReadOrWrite"

JeremyTheocharis (changes requested):
Note from o3 on my feedback:

```
MockConfigManager.writeConfig now assumes the caller already holds mutexReadOrWrite (which is the case for all production-equivalent paths – WriteConfigFromString, Atomic* helpers, etc.). Locking twice was redundant and risked deadlock; removing it is fine as long as the contract is documented in the function comment.
```

JeremyTheocharis (approved):

umh-core/pkg/communicator/actions/get-config-file.go
```diff
@@ -0,0 +1,124 @@
...
+	// Get the config file content as string using the new method
+	content, err := a.configManager.GetConfigAsString(ctx)
+	if err != nil {
+		errMsg := fmt.Sprintf("Failed to read config file: %v", err)
+		SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionFinishedWithFailure,
+			errMsg, a.outboundChannel, models.GetConfigFile)
+		return nil, nil, fmt.Errorf("failed to read config file: %w", err)
+	}
+
+	lastModifiedTime, err := a.configManager.UpdateAndGetCacheModTime(ctx)
```
Scarjit: This seems inefficient, as it calls GetConfig internally, resulting in two filesystem reads.
Why isn't GetConfigAsString internally also updating the mod time ?

umh-core/pkg/communicator/actions/set-config-file.go
```diff
@@ -0,0 +1,155 @@
...
+	// Write the new content to the file with atomic concurrent modification check
+	err := a.configManager.WriteConfigFromString(ctx, a.payload.Content, a.payload.LastModifiedTime)
+	if err != nil {
+		errMsg := fmt.Sprintf("Failed to write config file: %v", err)
+		SendActionReplyV2(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionFinishedWithFailure,
+			errMsg, models.ErrRetryConfigWriteFailed, nil, a.outboundChannel, models.SetConfigFile, nil)
+		return nil, nil, fmt.Errorf("failed to write config file: %w", err)
+	}
+
+	newLastModifiedTime, err := a.configManager.UpdateAndGetCacheModTime(ctx)
```
Scarjit: Same as above, WriteConfig could directly also set the mod time.

Scarjit (changes requested):

PR #2059, opened 2025-05-23
ENG-3034 Generic get-metrics action

ricci2511:
In this pull request I've done 3 main things:
- Implement a generic `get-metrics` action to retrieve metrics from different sources with an enforced response structure
- Expose Redpanda metrics alongside existing DFC metrics through this new action
- Vastly improve DevEx by introducing [GoMock](https://github.com/uber-go/mock) to make unit testing more straightforward

### Action: `get-metrics`
This action follows the same pattern as `get-logs`. The request payload requires specifying the metric `type` and an optional `uuid` for sources that need it (currently only DFC metrics).

The response structure matches the existing `get-data-flow-component-metrics` action, providing metrics as a flattened array with all necessary metadata.

**Backwards Compatibility**: The old `get-data-flow-component-metrics` action remains functional and its related structs are marked as deprecated. DFC metrics can be retrieved through both actions equally.

**New Capability**: Redpanda metrics can now be fetched with the following action payload: `{"type": "redpanda"}`

### Provider Pattern
I introduced a "providers" package within actions to implement the `MetricsProvider` interface. This pattern:
- **Offloads business logic** from actions into focused, testable interfaces that handle domain-specific operations
- **Makes unit testing dramatically simpler** through interface mocking
- **Provides a reusable pattern** for other actions that need similar logic separation (e.g., LogsProvider, ConfigProvider, etc.)

### GoMock for Better Testing
The existing test setup (see the [old DFC metrics test](https://github.com/united-manufacturing-hub/united-manufacturing-hub/blob/staging/umh-core/pkg/communicator/actions/get-dataflowcomponent-metrics_test.go)) required extensive manual mocking of complex snapshot structures. With the provider interface + GoMock:
- Simple `go:generate` directive creates mocks automatically
- Clean, focused tests that mock only what matters
- Easy to test error scenarios without complex setup

**Quick Start**: Add `//go:generate mockgen ...` to your interface file, run `go generate ./...`, then use `NewMockInterfaceName` in tests. See `get-metrics_test.go` for examples.


umh-core/vendor/go.uber.org/mock/gomock/matchers.go
```diff
@@ -0,0 +1,447 @@
...
+
+	// Check if types assignable and convert them to common type
+	x1Val := reflect.ValueOf(e.x)
+	x2Val := reflect.ValueOf(x)
+
+	if x1Val.Type().AssignableTo(x2Val.Type()) {
+		x1ValConverted := x1Val.Convert(x2Val.Type())
+		return reflect.DeepEqual(x1ValConverted.Interface(), x2Val.Interface())
+	}
+
```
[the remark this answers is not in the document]
ricci2511: don't bother looking into vendor

umh-core/pkg/communicator/actions/providers/metrics_provider.go
```diff
@@ -0,0 +1,205 @@
...
+		return res, err
+	}
+
+	observedState, ok := dfcInstance.LastObservedState.(*dataflowcomponent.DataflowComponentObservedStateSnapshot)
+	if !ok || observedState == nil {
+		err = fmt.Errorf("DFC instance %s has no observed state", uuid)
+		return res, err
+	}
+
+	metrics := observedState.ServiceInfo.BenthosObservedState.ServiceInfo.BenthosStatus.BenthosMetrics.Metrics
```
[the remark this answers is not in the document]
ricci2511: Actually `dfcInstance` needs a nil check.
Observed state is already checked, but everything from `ServiceInfo` down `Metrics` is not a pointer.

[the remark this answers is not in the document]
ricci2511: The helper `FindDfcInstanceByUUID` already does the nil checking internally by skipping nil instances, which will return an error due to the instance not being found.

umh-core/pkg/communicator/actions/actions.go
```diff
@@ -138,6 +138,8 @@ func HandleActionMessage(instanceUUID uuid.UUID, payload models.ActionMessagePay
 			actionLogger:          log,
 			systemSnapshotManager: systemSnapshotManager,
 		}
+	case models.GetMetrics:
+		action = NewGetMetricsAction(sender, payload.ActionUUID, instanceUUID, outboundChannel, systemSnapshotManager, log)
```
FreiPaul, line 163: Is there a reason why you do not do it as above with direct initialization?

umh-core/pkg/communicator/actions/get-metrics.go
```diff
@@ -0,0 +1,127 @@
...
+	payload models.GetMetricsRequest
+
+	// ─── Utilities ──────────────────────────────────────────────────────────
+	provider     providers.MetricsProvider
+	actionLogger *zap.SugaredLogger
+}
+
+// NewGetMetricsAction creates a new GetMetricsAction with the provided parameters.
+// Caller needs to invoke Parse and Validate before calling Execute.
+func NewGetMetricsAction(userEmail string, actionUUID uuid.UUID, instanceUUID uuid.UUID, outboundChannel chan *models.UMHMessage, systemSnapshotManager *fsm.SnapshotManager, logger *zap.SugaredLogger) *GetMetricsAction {
```
Scarjit, line 51: This differs from other ones.

umh-core/pkg/communicator/actions/providers/metrics_provider_mock.go
```diff
@@ -0,0 +1,71 @@
...
+func (m *MockMetricsProvider) GetMetrics(payload models.GetMetricsRequest, snapshot fsm.SystemSnapshot) (models.GetMetricsResponse, error) {
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "GetMetrics", payload, snapshot)
+	ret0, _ := ret[0].(models.GetMetricsResponse)
+	ret1, _ := ret[1].(error)
+	return ret0, ret1
+}
+
+// GetMetrics indicates an expected call of GetMetrics.
+func (mr *MockMetricsProviderMockRecorder) GetMetrics(payload, snapshot any) *gomock.Call {
```
Scarjit: Can we rename this to something else, otherwise both very different calls will be called GetMetrics

Scarjit (approved):

umh-core/pkg/communicator/actions/get-metrics.go
ricci2511: I guess the differences are:
- doesn't include the `configManager` because it's not needed, read-only action
- includes a `logger` param to use it as the `actionLogger`

I have [here](https://github.com/united-manufacturing-hub/united-manufacturing-hub/compare/eng-3034-generic-get-metrics-action...actions-factory-di-pattern) a brief poc which avoids this problem of possibly having different signatures. It uses dependency injection with a factory pattern. Might be worth a look.

umh-core/pkg/communicator/actions/actions.go
ricci2511: I know it looks inconsistent right now, but i think we should enforce using constructors to encapsulate initialization logic and provide better separation of concerns.

For example, in this case with direct struct initialization, the action handler has to specify the `provider` field - but provider selection is an implementation detail that the action handler shouldn't care about. The constructor handles all dependency setup including the default provider that satisfies the interface, so the action handler just says 'give me a metrics action' without needing to know about provider internals.
I used the `provider` as an example, but it would also go for any other "extra" fields that are not the action dependencies.

If still required, i can just switch to struct initialization.

umh-core/pkg/communicator/actions/providers/metrics_provider_mock.go
ricci2511: This is auto-generated gomock code. The mock recorder method names always match the interface method names. But perhaps I've misunderstood the request?
The go directive at the top of the metrics_provider file: `//go:generate mockgen -destination=metrics_provider_mock.go -package=providers github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/communicator/actions/providers MetricsProvider`

umh-core/pkg/communicator/actions/actions.go
ricci2511: Maybe you can check out https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/2059#discussion_r2120669651 for a similar issue

FreiPaul (approved):

PR #2069, opened 2025-06-11
ENG-3050 implement re-authentification routine

Scarjit:
This implements a small routine which will attempt to update the auth token every 1 hours.
Our JWT currently has an expiry of 24 hours, giving it ~24 attempts to refresh before expiry (and even if expired it will continue to re-try).
Imo this is a good balance between reliabilty and keeping the API traffic low.
It also matches our old implementation

See also the companion implementation in https://github.com/united-manufacturing-hub/ManagementConsole/blob/staging/companion/cmd/appstate/appstate.go#L395

I did not implement the part where it decodes the expiry time, as for now i would try to keep this simple.
If on a later date we want to make the refresh interval dynamic based on token expiry time, it should be easily implementable.

It also uses the Watchdog to prevent any "hangs".

### DIFF
--- a/umh-core/Makefile
+++ b/umh-core/Makefile
@@ -225,7 +225,7 @@ integration-test: stop-all-pods vet nilaway lint
 pre-test: stop-all-pods vet nilaway lint
 
 # Meta target to setup the environment for the tests in CI environments
-pre-test-ci: vet nilaway lint
+pre-test-ci: vet lint
 
 integration-test: pre-test
 	ginkgo -r -v --tags=test --label-filter='integration' --fail-fast ./...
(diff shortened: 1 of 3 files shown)


Scarjit:
bugbot run

Scarjit:
bugbot run

umh-core/Makefile
```diff
@@ -225,7 +225,7 @@ integration-test: stop-all-pods vet nilaway lint
 pre-test: stop-all-pods vet nilaway lint
 
 # Meta target to setup the environment for the tests in CI environments
-pre-test-ci: vet nilaway lint
```
Scarjit, line 228: This was removed from the integration tests, due to memory constraints.
It is still run on the go unit test, which uses a different runner

Scarjit: This is actually part of https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/2070, but this had the wrong base branch.

PR #2078, opened 2025-06-17
add templating for protocol converter

FreiPaul:


JeremyTheocharis (commented):
From O3

<html><head></head><body><p>Below is a <strong>full‑stack review</strong> that covers design, correctness, maintainability, test coverage and style.  I focus on the <em>new</em> templating functionality but also call out pre‑existing problems that the PR touches.</p>
<hr>
<h2>1 &nbsp;High‑level assessment</h2>

Area | Verdict | Rationale (most important items are expanded in later sections)
-- | -- | --
Architecture & design | 🟢 Sound idea | Moving templates to a dedicated templates: section keeps YAML concise and makes inheritance explicit.
Correctness | 🟠 Some edge‑case bugs | Inconsistent root detection, possible orphaned children, cache coherency temporarily disabled.
Backward compatibility | 🟠 | Renaming .template → .config is a breaking change for users and for any tooling that relied on the old field.
Concurrency & atomicity | 🟢 | New mutexAtomicUpdate guarding add/edit/delete looks correct.
Tests | 🟢 | Large number of new unit tests, good round‑trip checks.
Documentation | 🔴 | No migration note for the breaking YAML change; Godoc not updated everywhere.
Style & naming | 🟠 | Mixed semantics for “root” vs “stand‑alone”.  Field promotion done only in mocks.
Performance | 🟢 | Overhead negligible (single marshal‑unmarshal per template).


<hr>
<h3>Conclusion</h3>
<p>The PR lays solid groundwork for <strong>first‑class protocol‑converter templates</strong> and is largely well‑engineered.  Addressing the inconsistencies above (especially root detection and cache handling) will make the feature production‑ready and minimise surprises for users upgrading from previous releases.</p></body></html>

umh-core/pkg/config/manager.go
```diff
@@ -446,26 +440,47 @@ func (m *FileConfigManager) writeConfig(ctx context.Context, config FullConfig)
...
+		m.cacheRawConfig = string(data)
+		m.cacheModTime = info.ModTime()
+		m.cacheConfig = config
+		m.cacheMu.Unlock()
+	*/
 	m.logger.Infof("Successfully wrote config to %s", m.configPath)
 	return nil
 }
 
+// since we use this function in runtime_config_test to best cover the functionality, we export it
```
JeremyTheocharis: can you add a docstring here

umh-core/pkg/config/variables/example_test.go
```diff
@@ -28,7 +28,7 @@ var _ = Describe("VariableBundle User Experience Examples", func() {
 		// Create a bundle with all namespaces (as would happen internally)
 		vb := variables.VariableBundle{
 			User: map[string]any{
-				"HOST":    "localhost",
+				"HOST":    "wttr.in",
```
Scarjit, line 31: This can bite us later, if wttr.in is down.

umh-core/pkg/config/manager.go
```diff
@@ -836,11 +851,11 @@ func (m *FileConfigManagerWithBackoff) UpdateAndGetCacheModTime(ctx context.Cont
 // If expectedModTime is provided, it will check that the file hasn't been modified since that time
 func (m *FileConfigManager) WriteConfigFromString(ctx context.Context, configStr string, expectedModTime string) error {
```
JeremyTheocharis: maybe instead of "FromString", use the shared wording with "YAML" and "Spec"?

so writeConfigFromYaml (and then dosctring that this is used to write into the templates?) and writeConfigFromSpec?

maybe something like this would help understand what is there for what

umh-core/pkg/config/manager_test.go
```diff
@@ -261,26 +262,27 @@ agent:
 				Expect(config.Agent.Location[1]).To(Equal("Site"))
 			})
 
-			It("should handle empty input", func() {
-				config, err := parseConfig([]byte{}, false)
-				Expect(err).To(HaveOccurred())
-				Expect(config).To(Equal(FullConfig{}))
-			})
+			//TODO: fix this test cases
```
JeremyTheocharis: TODO :)

umh-core/pkg/config/protocolconverter.go
```diff
@@ -0,0 +1,268 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/google/uuid"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/config/dataflowcomponentserviceconfig"
+)
+
+// AtomicAddProtocolConverter adds a protocol converter to the config atomically
```
JeremyTheocharis: maybe mention in one additional sentence the special case now with templates, etc.

umh-core/pkg/config/protocolconverter.go
```diff
@@ -0,0 +1,268 @@
...
+
+	// Check if context is already cancelled
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
+
+	return m.configManager.AtomicAddProtocolConverter(ctx, pc)
+}
+
+// AtomicEditProtocolConverter edits a protocol converter in the config atomically
```
JeremyTheocharis: also here add a sentence with the special case

umh-core/pkg/config/protocolconverter.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	defer m.mutexAtomicUpdate.Unlock()
+
+	// get the current config
+	config, err := m.GetConfig(ctx, 0)
+	if err != nil {
+		return ProtocolConverterConfig{}, fmt.Errorf("failed to get config: %w", err)
+	}
+
+	// Find target index via GenerateUUIDFromName(Name) == componentUUID
```
JeremyTheocharis, line 124: why should we find the target index?

umh-core/pkg/config/protocolconverter.go
```diff
@@ -0,0 +1,268 @@
...
+
+	// write the config
+	if err := m.writeConfig(ctx, config); err != nil {
+		return ProtocolConverterConfig{}, fmt.Errorf("failed to write config: %w", err)
+	}
+
+	return oldConfig, nil
+}
+
+// AtomicDeleteProtocolConverter deletes a protocol converter from the config atomically
```
JeremyTheocharis: also add here an explanation what happens now with the golden instance. does it also delete the template?

JeremyTheocharis: in general explain in all of these functions a little bit the business logic, so when does editing will work and when not

umh-core/pkg/config/manager_test.go
Scarjit: Is this TODO still relevant ?

umh-core/pkg/config/variables/example_test.go
FreiPaul: The Host is never requested. This test is just for parsing and variable handling. I switched away from localhost because there was an issue with hosts that contained a dot

umh-core/pkg/config/yamlParsing.go
```diff
@@ -0,0 +1,215 @@
...
+// This translates between the "unrendered" config (with templateRef) and "rendered" config (with actual template content)
+func convertYamlToSpec(config FullConfig) (FullConfig, error) {
+	// Create a copy to avoid mutating the original
+	processedConfig := config.Clone()
+
+	// Build a map of available protocol converter templates for quick lookup
+	templateMap := make(map[string]protocolconverterserviceconfig.ProtocolConverterServiceConfigTemplate)
+
+	// Process protocol converter templates from the enforced structure
+	for templateName, templateContent := range config.Templates.ProtocolConverter {
```
Scarjit: This returns a ref/weak-copy to the original config, instead of the cloned one.
Probably not an issue in our single threaded approach, but using processedConfig could be safer in the future.

JeremyTheocharis:
bugbot run

JeremyTheocharis:
@coderabbitai full review

JeremyTheocharis (commented):
Update from O3 ( I think some of them were catched by coderabbit as well)

Below is a **follow‑up code review** for the delta `1f48c47…2f68013`.
I check whether the open points from the previous review were addressed and look for regressions or newly‑introduced problems.

---

## 1  Items that were **fixed / improved**

|  Topic                                        | Status       | Notes                                                                                          |
| --------------------------------------------- | ------------ | ---------------------------------------------------------------------------------------------- |
| `Clone()` omits Templates                     | **Fixed**    | `FullConfig.Clone()` now deep‑copies `ProtocolConverter` *and* the new `Templates` field.      |
| Unit‑test coverage                            | **Improved** | New `runtime_config` suite verifies variable expansion, location merging and error paths.      |
| Compile‑time duplication of `Templates` field | **Resolved** | The obsolete `[]map[string]interface{}` field was removed.                                     |
| `parseConfig` vs `ParseConfig`                | **Unified**  | Private helper dropped; exported `ParseConfig` is used everywhere; tests & benchmarks updated. |
| YAML `omitempty`                              | **Added**    | `BenthosConfig.{Input,Pipeline,Output}` and template fields now omit empties → smaller YAML.   |
| Docstrings                                    | **Added**    | `ParseConfig`, `WriteYAMLConfigFromString`, protocol‑converter FSM code now documented.        |

Great progress – code compiles, tests pass.

---

## 2  Issues that **remain outstanding**

|  #      | Finding                                                                                                                                                                          | Impact                                                                                 | Suggested fix                                                                                                                                            |
| ------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **2‑A** | `AtomicAddProtocolConverter` still treats **stand‑alone** converters (`TemplateRef==""`) as *children* and fails with `template "" not found`.  Mock implementation allows them. | Adds a blocker for users that do not use templates; mocks hide the bug → prod only.    | Change the condition to:<br>`if pc.ProtocolConverterServiceConfig.TemplateRef != "" && pc.ProtocolConverterServiceConfig.TemplateRef != pc.Name { ... }` |
| **2‑B** | `convertYamlToSpec` silently **overwrites an inline `config`** when `templateRef` is set, even if the two differ.                                                                | User edits are lost without warning.                                                   | Detect conflict and return an explicit error when inline ≠ template.                                                                                     |
| **2‑C** | After resolving a template, `TemplateRef` **is left in place**, so downstream code now sees *both* `Config` **and** `TemplateRef`.                                               | Ambiguity (root? child? stand‑alone?) can leak into other components.                  | Clear `TemplateRef` (or document & test why it must stay).                                                                                               |
| **2‑D** | Duplicate‑root detection still uses `reflect.DeepEqual`, so `{}` vs `nil` maps flag false positives.                                                                             | Spurious errors when the same template is serialised through different code paths.     | Normalise with `cmpopts.EquateEmpty()` or manual `nil→empty` helper before comparison.                                                                   |
| **2‑E** | Cache fast‑path in `writeConfig` is re‑enabled ✅, but the same path in `WriteYAMLConfigFromString` is still **disabled** → stale cache when editing via UI.                      | Users may read an old config immediately after saving.                                 | Either re‑enable or force a read‑after‑write before returning.                                                                                           |
| **2‑F** | Comment of `WriteYAMLConfigFromString` says “*bypasses template processing*” but the code calls `ParseConfig`, which **does** process templates.                                 | Mis‑leading docs.                                                                      | Update comment or fall back to a strict YAML‑only validation path.                                                                                       |
| **2‑G** | Root → stand‑alone transition (remove `TemplateRef` while children exist) is still accepted.                                                                                     | Leaves orphan children.                                                                | Reject edit if old instance was a root and still has dependants.                                                                                         |
| **2‑H** | `convertYamlToSpec` iterates over `config.Templates` (the *original* struct) instead of `processedConfig`. Scarjit’s remark persists.                                            | If `Templates` is modified in‑place earlier in the pipeline, resolution could miss it. | Use the local copy (`processedConfig.Templates`).                                                                                                        |
| **2‑I** | `WriteYAMLConfigFromString` lost the early `ctx.Err()` cancellation check.                                                                                                       | Long file writes may ignore cancelled requests.                                        | Re‑add the check before heavy work.                                                                                                                      |

---

## 3  Newly‑introduced issues

|  ID | Description                                                                                                                                                                                                                                                                   | Severity                                                                                       |
| --- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------- |
| N‑1 | Renamed variables (`renderedConfig`→`runtimeConfig`, `config`→`specConfig`) in FSM: good, but one leftover call in `manager_mock.go` still uses `configsEqual := protocolconverterserviceconfig.ConfigsEqual(protocolConverterInstance.config, …)` – should use `specConfig`. | Low (test‑only).                                                                               |
| N‑2 | `ParseConfig` now always calls `convertYamlToSpec`; if callers need the **verbatim** YAML (including anchors) they no longer have an API.  The removed `parseConfig` fulfilled that role.                                                                                     | Medium – affects tools that rely on anchor positions. Consider keeping a strict‑decode helper. |
| N‑3 | `convertYamlToSpec` clears `processedConfig.Templates` unconditionally, so a config that mixes protocol‑converter templates with **other future template sections** will lose them.                                                                                           | Low now, but future foot‑gun.                                                                  |

---

## 4  Minor observations / nits

* **Terminology**: code still uses both *normalizer* and *normaliser* – pick one.
* **Typos**: “diabled” → “disabled” (multiple comments).
* File headers still state *Copyright 2025*; company template says 2024‑(c).
* `yamlParsing.go` exceeds 700 lines; consider splitting into `toSpec.go` / `toYaml.go`.

---

## 5  Action checklist

1. **Fix 2‑A** (stand‑alone add) – users will hit this first.
2. Align real & mock semantics after (1).
3. Decide on `TemplateRef` clearing (2‑C) and implement.
4. Restore cache update (2‑E) or document why it is disabled.
5. Correct misleading comment in `WriteYAMLConfigFromString`.
6. Add regression tests for:

   * stand‑alone add succeeds,
   * root→stand‑alone edit with children fails,
   * inline‑config/TemplateRef conflict error.
7. Address minor clean‑ups and documentation items.

---

### Overall verdict

Most structural refactors were integrated correctly and the codebase now compiles and tests pass, but **critical edge‑case bugs are still present** (notably the stand‑alone add failure and orphan‑child scenario). Addressing the checklist above should bring the PR to a production‑ready state.

Scarjit (approved):

PR #2079, opened 2025-06-17
Protocol converter actions communicator

FreiPaul:


JeremyTheocharis (commented):
From O3

Below is a **structured, end‑to‑end review** of the pull‑request *“Protocol converter actions communicator”*.
I focus on API‐surface, correctness, maintainability, tests, performance, logging/security and style.
(Line numbers refer to the diff excerpts you provided; *fn* = function, *pkg* = package.)

---

## 1. Top‑level architecture

| Aspect                     | Assessment                                                                                                                                                                                    | Suggestions                                                                                                                                                                             |
| -------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **Separation‑of‑concerns** | ✅  The refactor moves generic parsing/validation/helpers into `dataflow-component-parser.go`, shrinking the action files and removing code duplication in *deploy/edit‑dataflowcomponent.go*. | –                                                                                                                                                                                       |
| **Extensibility**          | ✅  New actions (Deploy/Edit/Delete/Get ProtocolConverter) follow the existing `Action` interface, so the Communicator loop remains unchanged.                                                 | –                                                                                                                                                                                       |
| **Coupling**               | ⚠️  `EditProtocolConverterAction.Execute` now contains a complex decision tree (child/root, variable merging, rollback, health‑check).  The method is \~330 LOC and hard to reason about.     | Split the method:<br>• `applyMutation()` – returns new *spec* and *runtime* intents.<br>• `persistConfig()` – wraps the two *atomic* edits.<br>• `awaitRollout()` – only health‑checks. |

---

## 2. Shared helper (`dataflow-component-parser.go`)

### Positives

* **Single source of truth** for CDFC parsing, validation and Benthos conversion.
* Uses **`yaml.Unmarshal` on arbitrary fragments** ⇒ good validation depth.
* Introduces **order‑checking** for numeric processor keys (`CheckIfOrderedNumericKeys`).

### Issues & improvements

| Line                                                                                                                                     | Issue                                                                                                                                                                                                                                         | Recommendation                                                                                                          |
| ---------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------- |
| 115                                                                                                                                      | `customPayloadMap["customDataFlowComponent"]` – string literal duplicated in several places.                                                                                                                                                  | Extract to `const keyCustomDFC = "customDataFlowComponent"`.                                                            |
| 228                                                                                                                                      | `err := yaml.Unmarshal([]byte(proc.Data), &procConfig)` – YAML decoding inside hot loop; worst‑case O(N²) allocations if processors large.                                                                                                    | Decode once and **cache** in `proc` struct or stream processor YAMLs into a slice before the loop.                      |
| 283                                                                                                                                      | `hasNumericKeys := CheckIfOrderedNumericKeys(payload.Pipeline)` … then **require** that `hasNumericKeys` is *true*, else error. This changes behaviour for existing configs that used string keys intentionally (e.g. “dedupe”, “transform”). | Keep the strictness **only for new API versions** (v2) or gate behind a feature‑flag; otherwise the change is breaking. |
| 412‑416                                                                                                                                  | `inputType` / `outputType` heuristics rely on *first map key*.  For nested `"input"` wrapper that Benthos introduces, heuristic is fragile.                                                                                                   | \`\`\`go                                                                                                                |
| for path := \[]string{"input", "output"}; …\`\`\` – walk known wrapper levels or attach explicit `type` annotation during normalisation. |                                                                                                                                                                                                                                               |                                                                                                                         |

---

## 3. Actions

### 3.1  `DeployProtocolConverterAction`

* **Good:** deterministic UUID via `GenerateUUIDFromName` – matches existing pattern.
* **Error propagation:** wraps ConfigManager errors with context ✅.

| Line | Issue                                                                                                                                                                                                              | Recommendation                                                                                                                |
| ---- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | ----------------------------------------------------------------------------------------------------------------------------- |
| 135  | Validation rejects `Connection.IP == ""`, but IP will be templated (e.g. `"{{ .IP }}"`).  Rejecting empty string is okay, yet *non‑templated* “localhost” is accepted while invalid literal “foo bar” also passes. | Add **net.ParseIP** check *iff* not templated.                                                                                |
| 238  | `pcConfig.ProtocolConverterServiceConfig.TemplateRef = pcConfig.Name` every time ⇒ **child instances can never re‑use** a shared template.                                                                         | Keep the author’s intention? If templates should be reusable, omit this assignment for roots (the deploy is building a root). |
| 316  | `variables.User` only contains `"IP"`/`"PORT"`. All other user‑supplied variables in payload (`TemplateInfo.Variables`) are ignored at deploy time.                                                                | Merge the inbound variable array into `User` rather than ignoring.                                                            |
| 420  | Wait‑loop aborts after 30 s, rolls back, but returns **generic** error code `ERR_RETRY_ROLLBACK_TIMEOUT`. Frontend cannot distinguish “nmap can’t connect” vs “DFC failed”.                                        | Bubble up finer‑grained reason (`ErrConnectionTimeout`, `ErrDFCStuck` etc.).                                                  |

### 3.2  `EditProtocolConverterAction`

Positive patterns

* **Rollback on timeout & config errors** – good safety net.
* Uses **spec variable rendering** to compare desired vs observed configs 👏.

Issues

| Line | Issue                                                                                                                                                                                                                          | Suggestion                                                    |
| ---- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------- |
| 137  | `var newVB map[string]any` but later `maps.Copy(newVB, …)` without `make` when `isChild` => **nil‑map panic**.                                                                                                                 | Initialise with `make(map[string]any)` before `maps.Copy`.    |
| 147  | For child cases you compute `newVB` but then `maps.Copy(newVB, targetPC.ProtocolConverterServiceConfig.Variables.User)`, however `newVB` is still **nil** (see above).                                                         |                                                               |
| 263  | `dfcServiceConfig := dataflowcomponentserviceconfig.DataflowComponentServiceConfig{ BenthosConfig: benthosConfig }` – OK, but you drop other fields (e.g. **`ResourcesConfig`** in future).  Use full struct or a constructor. |                                                               |
| 380  | Health‑check loop duplicates large chunks of logic from Deploy.                                                                                                                                                                | Extract to common helper (`waitForProtocolConverterHealthy`). |
| 489  | `CheckBenthosLogLinesForConfigErrors(logs)` – function not in diff; ensure it is **unit‑tested** for false positives.                                                                                                          |                                                               |

### 3.3  `DeleteProtocolConverterAction`

*Pure & focused.*
One concern: after deleting the config you merely **sleep‐poll** snapshot until removal. That is fine, but you treat absence of manager as *still waiting*. In cases where manager crashes, action will wait full 30 s then fail – acceptable.

### 3.4  `GetProtocolConverterAction`

*Complex but necessary translation*.
Two accuracy issues:

| Code                                                                                                                                                | Problem |
| --------------------------------------------------------------------------------------------------------------------------------------------------- | ------- |
| `templateConfig.ConnectionServiceConfig.NmapTemplate.Port` – this is a *string* but later parsed as uint.  Missing path check: template may be nil. |         |
| `determineProcessingMode` looks **only at readDFC**. For write‑only PCs (rare but possible) meta will remain default.                               |         |

---

## 4. Generator changes (`pkg/generator`)

*Brings Protocol Converters into the DFS health view.*

| Line | Issue                                                                                                                                                                       | Recommendation |
| ---- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------- |
| 137  | Status message for state “startingFailedDFCMissing” is “No DFC added yet”. ❓  Should be *initialised but idle*— consider renaming for clarity.                              |                |
| 174  | `AvgInputThroughputPerMinuteInMsgSec` uses  `m.Input.MessagesPerTick /  constants.DefaultTickerTime.Seconds()`.  The **unit name** is confusing (per‑minute vs per‑second). |                |

---

## 5. Tests

* Large amount of ⚙️ **BDD tests** added – great!
* Coverage includes happy paths & many failures (duplicate name, config manager errors).
* Missing: **concurrency tests** (race detector) for variable merge & rollback path.

---

## 6. Logging & security

* **Logs leak user email?** All `SendActionReply` include `userEmail` as meta – acceptable because only sender receives the websocket, but double‑check server side.
* Convert `err.Error()` into user‑facing message – be sure no internal host paths leak.

---

## 7. Style / Minor nit‑picks

* Prefer `errors.Is(err, context.DeadlineExceeded)` to string compare when timeouts bubble up.
* `RemainingPrefixSec()` should probably return **plural‑aware** string (“1 s remaining”).
* Update `constants.DataflowComponentWaitForActiveTimeout` (20 s→30 s) – ensure docs & UI spinner align.

---

## 8. Migration / compatibility

* Backwards‑compatible at **storage layer** – existing `full-config.yaml` still valid.
* API surface: new endpoints (`action-deploy-protocol-converter`, etc.) don’t break old clients.

---

## 9. Summary & Priority actions

1. **Fix nil‑map panic** in variable merge (`EditProtocolConverterAction`, §3.2).
2. **Refactor** giant `Execute()` bodies – testability & readability.
3. Re‑evaluate **numeric‑key enforcement** in CDFC parsing to avoid silent breaking change.
4. Add **unit tests for template variable rendering** path and for `CheckBenthosLogLinesForConfigErrors`.
5. Clarify status/metric naming and logging of sensitive data.

With those addressed the PR looks solid and will greatly reduce duplicate logic around data‑flow components and protocol‑converters.

umh-core/pkg/communicator/pkg/generator/protocolconverter.go
```diff
@@ -0,0 +1,240 @@
...
+		case "closed_filtered", "closed|filtered":
+			nmapSuffix = " (port is closed or filtered)"
+		case "starting":
+			nmapSuffix = " (nmap is starting)"
+		case "stopped":
+			nmapSuffix = " (nmap is stopped)"
+		case "degraded":
+			nmapSuffix = " (nmap execution failed)"
+		default:
+			nmapSuffix = fmt.Sprintf(" (nmap: %s)", nmapState)
```
Scarjit: maybe add here, that this state is unexpected ?

umh-core/pkg/communicator/actions/get-protocolconverter.go
```diff
@@ -0,0 +1,396 @@
...
+
+	processors := readDFC.Pipeline.Processors
+
+	// If more than one processor, return custom
+	if len(processors) > 1 {
+		return "custom"
+	}
+
+	// If exactly one processor, check its type
+	if len(processors) == 1 {
```
Scarjit, line 147: This check is not needed, as a zero lenght array/map will just be skipped.

umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -0,0 +1,604 @@
...
+	instanceUUID uuid.UUID
+
+	outboundChannel chan *models.UMHMessage
+	configManager   config.ConfigManager
+
+	// Parsed request payload (only populated after Parse)
+	protocolConverterUUID uuid.UUID
+	name                  string // protocol converter name (optional for updates)
+	dfcPayload            models.CDFCPayload
+	dfcType               string // "read" or "write"
```
Scarjit: maybe use an enum ?

umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -0,0 +1,639 @@
...
+			return config.ProtocolConverterConfig{}, uuid.Nil, fmt.Errorf("root template %s not found for child protocol converter %s",
+				targetPC.ProtocolConverterServiceConfig.TemplateRef, targetPC.Name)
+		}
+
+		// Apply mutations to the root, but keep child's variables
+		instanceToModify = rootPC
+		atomicEditUUID = dataflowcomponentserviceconfig.GenerateUUIDFromName(rootPC.Name)
+
+		// Add the new variables and preserve existing child variables
+		newVB = make(map[string]any)
```
Scarjit, line 357: https://pkg.go.dev/maps#Copy might be the cleaner option :)

Copy copies all key/value pairs in src adding them to dst. When a key in src is already present in dst, the value in dst will be overwritten by the value associated with the key in src.

JeremyTheocharis:
bugbot run

JeremyTheocharis:
@coderabbitai full review

JeremyTheocharis (commented):
New things from O3

Below is a focused “delta‑review” of the range **adaeb35 … 69db107** that you asked for.
I check each item that was called‑out in the first review, verify whether it is now fixed / addressed, and highlight any **new issues** (mostly compile‑time problems introduced while refactoring).

---

## 1 – Previously‑reported problems ✔️/❌

| # | Issue raised in first review                                                           | Status in 69db107                                                                                                                                                                                             | Notes |
| - | -------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----- |
| 1 | **nil‑map panic** when merging `newVB` in **EditProtocolConverterAction** (child path) | **Fixed** ✔️ – `newVB = make(map[string]any)` added before `maps.Copy`.                                                                                                                                       |       |
| 2 | User‑supplied template variables ignored in **DeployProtocolConverterAction**          | **Fixed** ✔️ – variables from `payload.TemplateInfo.Variables` are now merged into `userVars`.                                                                                                                |       |
| 3 | **Validate name** for spaces/special characters                                        | **Added** ✔️ – new `ValidateProtocolConverterName` helper in `dataflow-component-parser.go`; invoked in Deploy **and** Edit actions.                                                                          |       |
| 4 | Giant `Execute()` in Edit action                                                       | **Addressed** ✔️ – split into `applyMutation`, `persistConfig`, `awaitRollout`.                                                                                                                               |       |
| 5 | Location / connection changes not reflected on **edit**                                | **Fixed** ✔️ – `location` map and connection `IP/PORT` are propagated and previous auto‑generated vars stripped before persisting.                                                                            |       |
| 6 | Duplicate **log‑check rollback** code                                                  | **Still duplicated** ❌ – logic is now in two helpers (`waitForComponentToBeActive` and inline rollback in `awaitRollout`). Having one source would remain cleaner, but it is no longer blocking or incorrect. |       |
| 7 | Tests used outdated error strings                                                      | **Fixed** ✔️ – assertions now compare lower‑case messages produced after refactor.                                                                                                                            |       |

**Bottom line:** all *functional* bugs that blocked the happy‑path appear addressed.
However several *new* compile‑time / style issues sneaked in during the hurried refactor.

---

## 2 – New problems introduced in 69db107

### 2.1  Compile‑time errors (the build will fail)

| File / Line                                                    | Problem                                                                                                                                                                                                                                                                                                                                       |
| -------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **deploy‑protocolconverter.go** – literal for `variableBundle` | Composite literal contains the field `User` **twice**:  `variables.VariableBundle{ User: … , User: userVars }`.  Go rejects duplicate fields → compilation error. Remove the first initialiser and keep only `User: userVars`.                                                                                                                |
| **edit‑protocolconverter.go – applyMutation**                  | Legacy error‑handling lines were left in place when the function signature changed. Ex.:  `return nil, nil, fmt.Errorf("%s", errorMsg)` followed by `return config.ProtocolConverterConfig{}, uuid.Nil, …` – the first return makes the second unreachable → compile‑time *“unreachable code”*. Remove the obsolete `return nil, nil…` lines. |
| **edit‑protocolconverter.go – not‑found branch**               | Same duplicate‐return pattern as above; delete the obsolete line.                                                                                                                                                                                                                                                                             |
| **edit‑protocolconverter.go – method duplicates**              | `getUserEmail`, `getUuid`, `GetParsedPayload`, `GetProtocolConverterUUID`, `GetDFCType` are each defined **twice** (once above and once at the bottom of the file). Go does not allow duplicate method names on the same type. Remove the lower duplicates.                                                                                   |

> **Impact:** The current HEAD would not compile (`go vet` would stop).

### 2.2  Interface drift

* `ConfigManager` now exposes **WriteYAMLConfigFromString** instead of **WriteConfigFromString**.
  Most callers were updated, but `FileConfigManager` still keeps the *old* `WriteConfigFromString` (and `FileConfigManagerWithBackoff` delegates to it). That is okay – extra methods do not break the interface – **but** double‑check every call‑site (search for `WriteConfigFromString(`) to ensure none remain. The mock implementation was updated correctly.

### 2.3  Unused / dead code

* In **DeployProtocolConverterAction** the local variable `variableBundle` is now **never used**, so the compiler will emit `variableBundle declared but not used` once the duplicate‑field issue is fixed. Either return it or delete the declaration.

### 2.4  Behavioural / corner‑case observations

1. **Name validation** now forbids underscores (`_`) which older configs allowed. If existing converter names contain underscores, the edit path will start failing. Decide if the stricter rule is intentional; if not, allow `_` or grandfather existing names.
2. When you **rename a root template**, `TemplateRef` for each child is not automatically updated (yet). The code comments say this will be handled in `AtomicEditProtocolConverter`, but the implementation still performs only a single‐instance save. Verify that the recursive update logic exists (or add a TODO).
3. `awaitRollout` performs two almost‑identical location/rollback cleanup chunks ― could be refactored (non‑blocking).

---

## 3 – Tests

* New tests were adjusted for changed error strings – good.
* **Compilation failures above will cause the whole test suite to stop**; fix those first.
* No new coverage for the `ValidateProtocolConverterName` helper – recommend adding small table‑driven test.

---

## 4 – Recommended quick‑fix list

1. **Deploy action**

   ```go
   variableBundle := variables.VariableBundle{
       User: userVars,   // <-- keep only this
   }
   // remove the earlier duplicate initialiser OR delete variableBundle if unused
   ```

2. **Edit action – applyMutation & not‑found branch**
   Delete the first `return nil, nil, fmt.Errorf("%s", errorMsg)` in each case.

3. **Edit action – duplicate methods**
   Remove the second block (lines 560‑end in the diff) that re‑defines the same getters.

4. Run `go vet ./...` and `go test ./...` – should compile cleanly again.

---

## 5 – Conclusion

*✅  Previously‑reported runtime bugs are addressed and the refactor is cleaner.*
*❌  A few mechanical mistakes will currently break the build (duplicate fields, duplicate methods, unreachable code).*

Once the four quick fixes above are applied, I do not see any new functional blockers introduced in this delta.

umh-core/pkg/communicator/pkg/generator/protocolconverter.go
```diff
@@ -0,0 +1,240 @@
...
+	// }
+
+	//check if the protocol converter is initialized by checking if a read dfc is present
+	isInitialized := false
+	input := observed.ObservedProtocolConverterRuntimeConfig.DataflowComponentReadServiceConfig.BenthosConfig.Input
+	if len(input) > 0 {
+		isInitialized = true
+	}
+
+	dfc := models.Dfc{
```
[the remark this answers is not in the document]
FreiPaul: This will be added when we introduce writeDFC

umh-core/pkg/communicator/actions/get-protocolconverter.go
FreiPaul: I'll just leave it here for better readability

umh-core/pkg/communicator/actions/edit-protocolconverter.go
FreiPaul: I already did that right below but vb is not a map

umh-core/pkg/fsm/protocolconverter/manager.go
```diff
@@ -70,7 +71,12 @@ func NewProtocolConverterManager(name string) *ProtocolConverterManager {
 			if !ok {
 				return false, fmt.Errorf("instance is not a ProtocolConverterInstance")
 			}
-			return protocolConverterInstance.specConfig.Equal(cfg.ProtocolConverterServiceConfig), nil
+			equal := protocolConverterInstance.specConfig.Equal(cfg.ProtocolConverterServiceConfig)
+			if !equal {
+				diff := protocolconverterserviceconfig.ConfigDiff(protocolConverterInstance.specConfig, cfg.ProtocolConverterServiceConfig)
+				logger.For(logger.ComponentProtocolConverterManager).Debugf("diff: %s", diff)
```
Scarjit: We probably don't need this anymore.

Scarjit (approved):

PR #2090, opened 2025-06-23
[ENG-3090] Update GetLogs implementation to handle log rotation gracefully.

Scarjit:


Scarjit:
bugbot review

umh-core/main
JeremyTheocharis, line 1: :)

umh-core/pkg/service/s6/s6.go
```diff
@@ -1455,18 +1620,28 @@ func (s *DefaultService) GetLogs(ctx context.Context, servicePath string, fsServ
...
-		st.inode, st.offset = ino, 0
+		s.logger.Debugf("Detected rotation for log file %s (inode: %d->%d, offset: %d, size: %d)",
+			logFile, st.inode, ino, st.offset, size)
 
-		if st.logs == nil {
-			st.logs = make([]LogEntry, 0, constants.S6MaxLines)
-		} else {
-			st.logs = st.logs[:0] // reuse backing array
-		}
+		// Phase 1: Finish reading the rotated file to prevent protobuf message loss
```
JeremyTheocharis: maybe not use protobuf here in the comments, use prevent loosing old logs

umh-core/pkg/service/s6/s6.go
```diff
@@ -1359,7 +1369,149 @@ func (s *DefaultService) ForceRemove(
 	return nil
 }
 
-// GetLogs reads “just the new bytes” of the log file located at
+// finishRotatedFile reads remaining data from rotated file to prevent protobuf message loss.
```
JeremyTheocharis: maybe lets not use protobuf, but instead logs. so prevent log loss

maybe explain at least once how this log loss can happen without this code: "without this log files that are being rotated could be never fully read, as the rest of the implementation here only reads current. so when a log rotation occurs, we need to read the remaining old file and append that first to our buffer"

umh-core/pkg/service/s6/s6.go
```diff
@@ -1359,7 +1369,149 @@ func (s *DefaultService) ForceRemove(
...
+// This is critical for topic browser output parsing where binary protobuf messages
```
JeremyTheocharis: this is critical e.g., for the topic browser

umh-core/pkg/service/s6/s6.go
```diff
@@ -1455,18 +1620,28 @@ func (s *DefaultService) GetLogs(ctx context.Context, servicePath string, fsServ
...
+	// which is the initialization case for a newly created current file with no rotated files.
 	if st.inode != ino || st.offset > size {
-		st.inode, st.offset = ino, 0
+		s.logger.Debugf("Detected rotation for log file %s (inode: %d->%d, offset: %d, size: %d)",
+			logFile, st.inode, ino, st.offset, size)
 
-		if st.logs == nil {
-			st.logs = make([]LogEntry, 0, constants.S6MaxLines)
-		} else {
-			st.logs = st.logs[:0] // reuse backing array
```
JeremyTheocharis, line 1465: where did this code go? i remember it was important for memory reasons

JeremyTheocharis: i see the code comment further below, but I remember that i spent a good amount of time on this logic here so I am a little bit worried that the AI just simply removed this, but there was an actual purpose behind it

umh-core/pkg/service/s6/s6.go
```diff
@@ -1455,58 +1504,69 @@ func (s *DefaultService) GetLogs(ctx context.Context, servicePath string, fsServ
...
+		rotatedFile := s.findLatestRotatedFile(ctx, logDir, fsService)
+		if rotatedFile != "" {
+			var err error
+			rotatedContent, _, err = fsService.ReadFileRange(ctx, rotatedFile, st.offset)
+			if err != nil {
+				s.logger.Warnf("Failed to read rotated file %s from offset %d: %v", rotatedFile, st.offset, err)
+			} else if len(rotatedContent) > 0 {
+				s.logger.Debugf("Read %d bytes from rotated file %s", len(rotatedContent), rotatedFile)
+			}
+		}
```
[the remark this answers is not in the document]
Scarjit: The current is rotated to the rotated file, therefore this is exactly what we want.

umh-core/pkg/service/s6/s6.go
Scarjit: I now added a comment why my current implementation does not need to do this.
The gist of it is: We track the state of the ring buffer using st.head and st.full and do not need to reset the arrays internal size.

Scarjit:
bugbot run

JeremyTheocharis (approved):

PR #2091, opened 2025-06-24
Allow benthos commit hash instead of released versions for testing

Scarjit:
This allows runs like: `make BENTHOS_UMH_VERSION=sha-f1b3a27 test-no-copy` to quickly test out an unreleased benthos version.
It modifies the existing logic by loading the binary from the docker container.

This proved usefull when testing https://github.com/united-manufacturing-hub/benthos-umh/pull/162


JeremyTheocharis (changes requested):
besides the AI feedback, can you document how to use it? maybe somewhere in the docs here? https://docs.umh.app/reference/environment-variables

Scarjit:
@JeremyTheocharis Those are internal variables of our build system, i therefore just documented it in the makefile.

FreiPaul (approved):

PR #2094, opened 2025-06-24
ENG-2987 Enable compaction for umh-core

Scarjit:


umh-core/examples/example-config-redpanda-with-delete.yaml
```diff
@@ -0,0 +1,51 @@
...
+        # this corresponds to retention_bytes (See also: https://docs.redpanda.com/current/manage/cluster-maintenance/disk-utilization/)
+        defaultTopicRetentionBytes: 0
+
+        # this corresponds to log_compression_type (See also: https://docs.redpanda.com/current/manage/cluster-maintenance/compression/)
+        defaultTopicCompressionAlgorithm: "snappy"
+
+        # this corresponds to log_cleanup_policy (See also: https://docs.redpanda.com/current/manage/cluster-maintenance/compaction-settings/)
+        # Options: "delete" (time/size-based deletion), "compact" (keep latest per key), "compact,delete" (both)
+        # Using "delete" for traditional time-based retention
+        defaultTopicCleanupPolicy: "delete"
```
JeremyTheocharis, line 38: why a delete one and not compact?

JeremyTheocharis: got it, so it can be used in a test to check if the change is working, maybe but this as package comment / top comment

JeremyTheocharis (commented):
i dont see any big issues here, but let me wait until it is readty for review :)

PR #2096, opened 2025-06-24
[ENG-3081] Topic Browser Communicator

FreiPaul:


JeremyTheocharis (changes requested):
## No Real Data Retrieved for Topic Browser

The PR introduces a TopicBrowserCache and updates it every second, but it never actually retrieves real observed state from the Benthos “topic browser” plugin. Both the cache updater and status message generator pass an empty ObservedState instead of using the systemSnapshotManager to fetch actual topic data
[github.com](https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/2096#:~:text=StartTopicBrowserCacheUpdater%28systemSnapshotManager%20,w%22%2C%20err)
[github.com](https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/2096#:~:text=%2F%2F%20,isBootstrapped)
. This means the topic browser will always be populated with empty or stale data. The systemSnapshotManager parameter in StartTopicBrowserCacheUpdater is completely unused, indicating an incomplete integration. As a result, the UI’s topic browser will not reflect real-time topics/events from Benthos, contradicting the plugin’s design of emitting a full topic map of current topics
[GitHub](https://github.com/united-manufacturing-hub/benthos-umh/blob/9c8217056c19ae51e368ba94991bb1d3a75456a3/topic_browser_plugin/metadata.go)
. In short, the feature is non-functional until the code fetches and uses real ObservedState data instead of a placeholder.

--> Simply add a TODO there that this will be integrated once the FSM is ready and available


## Goroutine and Ticker Leak

The StartTopicBrowserCacheUpdater spawns a goroutine that runs forever, ticking every second without any shutdown mechanism
[github.com](https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/2096/files/53818b44225ebdd8f56aea3186d4f4761e718734#:~:text=func%20%28c%20,Failed%20to%20update%20topic)
. There is no context cancellation or stop condition. If the UMH core shuts down or restarts the communicator, this goroutine (and its time.Ticker) will continue running until the process exits. This is a resource leak: it creates an ever-running loop that cannot be cleanly stopped, potentially preventing garbage collection of related objects. Over time or in tests, these leaked goroutines could accumulate. Best practice is to tie such background loops to a context or provide a way to cancel the ticker (e.g. on shutdown or via defer ticker.Stop() and a break). As is, the code ignores this, so the cache updater will run indefinitely. This is especially problematic if multiple CommunicationState instances are ever created (each would spawn its own orphaned updater). It’s also a design inconsistency since other components likely honor context cancellation.

## Potential Race Conditions 
It’s not clear if TopicBrowserCache is internally thread-safe. The cache is updated in the background goroutine, and it’s read when generating status messages or possibly by other threads (e.g. an HTTP endpoint or subscriber handler). In the added code, there are no locks or synchronization around cache access. For instance, the cache updater could be modifying internal structures at the same time a status message is being composed. If TopicBrowserCache uses simple data structures (slice, map, etc.) without locks, this can cause race conditions or data corruption. The Benthos plugin code uses a mutex to guard its topic metadata cache
[GitHub](https://github.com/united-manufacturing-hub/benthos-umh/blob/9c8217056c19ae51e368ba94991bb1d3a75456a3/topic_browser_plugin/metadata.go)
; the UMH Core should similarly protect shared state. Failing to do so might lead to intermittent panics or wrong data in the topic browser output under high load or concurrent use. This is a code safety issue that should be addressed by using synchronization (e.g. a mutex inside TopicBrowserCache.Update or locking around uses of the cache).

## Lack of Documentation and Clarity
Key functions added by this PR lack explanatory comments about their purpose, limitations, or usage. For example, StartTopicBrowserCacheUpdater has no comment describing how it works or that it runs forever. Likewise, if TopicBrowserCache.Update expects a certain structure from ObservedState, there’s no comment in this PR about what ObservedState contains (since it’s passed as an empty struct, a maintainer might be puzzled). Important design details – such as the fact that the topic browser data comes from a Benthos plugin and that this cache updater is a temporary stand-in – are not documented in the code. This hampers readability and maintenance, as new contributors won’t know the intent or correct usage of these components. Comments or TODOs should clarify that integration with real data is pending, and how the cache is supposed to function (e.g. “cache retains last N topic events for quick retrieval by UI”).

## Error Handling and Consistency
The code handles certain error conditions superficially, which could lead to silent failures. In InitialiseAndStartSubscriberHandler, after adding the new TopicBrowserCache parameter, the code checks if c.SubscriberHandler == nil and simply reports an issue to Sentry
[github.com](https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/2096/files/53818b44225ebdd8f56aea3186d4f4761e718734#:~:text=Expand%20Down%20%20100%20Expand,Failed%20to%20create)
. That indicates a failure to create the subscriber handler (perhaps due to misconfiguration), but after logging, it does nothing to stop or retry. The system would continue running without a functional subscriber handler, likely breaking any feature that relies on it. This approach is inconsistent with robust error handling – critical initialization failures should propagate or halt the program rather than just logging. Similarly, the cache updater logs an error every time Update fails but does not attempt any recovery or exit, meaning it could spam the logs continuously without escalation. In general, errors are not well-surfaced: for example, if the topic browser plugin’s data cannot be parsed, the system would just keep printing “Failed to update topic browser cache” every second
[github.com](https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/2096/files/53818b44225ebdd8f56aea3186d4f4761e718734#:~:text=go%20func%28%29%20,w%22%2C%20err%29)
 with no back-off or notification to the rest of the system. This could overwhelm logs/Sentry and make real issues harder to spot (a performance and observability concern). The error handling strategy should be revisited to align with project conventions (e.g. fail fast on critical init errors, implement backoff or one-time reporting for repetitive errors, etc.).

## Unbounded Data Growth
Maybe add some protections here against unrealistic large topic maps (e.g., over 1 million topics), or very large events

## Style and Naming
Overall, the code added appears to follow the project’s style (e.g. using zap.SugaredLogger for logging, the license headers, etc.), but there are a few minor inconsistencies:
Exported types and funcs vs. unexported: The PR adds TopicBrowserCache *topicbrowser.Cache to CommunicationState (exported field) and a public StartTopicBrowserCacheUpdater method
[github.com](https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/2096#:~:text=StartTopicBrowserCacheUpdater%28systemSnapshotManager%20,w%22%2C%20err)
. Given this is internal to the communicator, one might question if it needs to be exported or could be unexported (if not used outside the package). Not a major issue, but keeping the API surface minimal is a convention.
The new test suite files (e.g. topicbrowser_suite_test.go) follow the naming convention but contain only the boilerplate. That’s fine, but it’s worth noting that the actual test coverage of complex logic (like merging bundles) is in a single test function. Possibly more granular tests (for edge cases like empty input, cache overflow, etc.) would align better with how other parts of the project are tested.
Error messages: The project guidelines (as shown in the CodeRabbit snippet) emphasize clear, structured error messages. The new errors (e.g., “Failed to update topic browser cache: %w”) are concise but lack context or mitigation suggestions. While these are internal errors for Sentry, ensuring consistency (maybe include what was being updated or which component failed) could help ops troubleshooting. It’s a minor style point, but aligning with the error-message conventions would improve consistency.

umh-core/pkg/communicator/communication_state/communication_state.go
```diff
@@ -48,6 +51,8 @@ type CommunicationState struct {
 	ConfigManager         config.ConfigManager
 	ApiUrl                string
 	Logger                *zap.SugaredLogger
+	TopicBrowserCache     *topicbrowser.Cache
+	TopicBrowserSimulator *topicbrowser.Simulator
```
JeremyTheocharis, line 56: explain why there is a simulator here, and what its used for. that everyone knows not to depend on this except in testing

umh-core/pkg/communicator/communication_state/communication_state.go
```diff
@@ -145,6 +152,42 @@ func (c *CommunicationState) InitialiseAndStartRouter() {
 	c.Router.Start()
 }
 
+func (c *CommunicationState) StartTopicBrowserCacheUpdater(systemSnapshotManager *fsm.SnapshotManager, ctx context.Context) {
```
JeremyTheocharis: docstring :)

JeremyTheocharis: not a big one, but just 1-3 sentences

umh-core/pkg/communicator/communication_state/communication_state.go
```diff
@@ -145,6 +152,42 @@ func (c *CommunicationState) InitialiseAndStartRouter() {
...
+
+	runSimulator := false
+	c.TopicBrowserSimulator = topicbrowser.NewSimulator()
+
+	ctxCfg, cncl := tools.Get1SecondContext()
+	defer cncl()
+	configCopy, err := c.ConfigManager.GetConfig(ctxCfg, 0)
```
JeremyTheocharis: do we really need to call this, or can we not also pass in the config from the central main.go, where the config is already read in

umh-core/pkg/communicator/communication_state/communication_state.go
```diff
@@ -145,6 +152,42 @@ func (c *CommunicationState) InitialiseAndStartRouter() {
...
+		for {
+			select {
+			case <-ctx.Done():
+				ticker.Stop()
+				return
+			case <-ticker.C:
+				if runSimulator {
+					c.TopicBrowserSimulator.Tick()
+				}
+				err := c.TopicBrowserCache.Update(c.TopicBrowserSimulator.GetSimObservedState())
```
JeremyTheocharis, line 178: does this mean this only works for the simulator?  how would you integrate it with real data?

umh-core/pkg/communicator/pkg/generator/topicbrowser.go
```diff
@@ -0,0 +1,329 @@
...
+	MaxBufferSize = 100 * 1024 * 1024
+
+	// MaxBundlesPerRequest limits bundles returned in a single request
+	MaxBundlesPerRequest = 5
+)
+
+// GenerateTopicBrowser is the main entry point for generating TopicBrowser content.
+// It implements the core business logic for topic browser cache management:
+//
+// The cache (topicbrowser.Cache) maintains exactly one event per topic to provide a snapshot
```
JeremyTheocharis: maybe lets avoid usign hte word snapshot, as this is used throguhotu the code base with a different meaning

umh-core/pkg/communicator/pkg/generator/topicbrowser.go
```diff
@@ -0,0 +1,329 @@
...
+// of all topics with their latest values. This cache is used to efficiently serve
```
JeremyTheocharis: is this really the purpose? or is the purpose not more that each subscriber has for each topic at least once value available

umh-core/pkg/communicator/pkg/generator/topicbrowser.go
```diff
@@ -0,0 +1,329 @@
...
+			TopicCount: cache.Size(),
+			UnsBundles: make(map[int][]byte),
+		}
+	}
+
+	if isBootstrapped {
+		// Existing subscriber: Get only pending bundles from observed state
+		return GenerateTbContent(cache, obs, logger)
+	} else {
+		// New subscriber: Get the full cache PLUS any new bundles since cache was updated
```
JeremyTheocharis: I uynderstand the comments ,but a little bit confused about the code.

shouldnt it then be called 

generateTbContentForNewSubscriber(cache, obs, logger)
GenerateTbContent(cache, obs, logger)

then we could also have a logic like

if !isBootstrapped:
   append(generateTbContentForNewSubscriber)

append(GenerateTbContent)

could be easier to understand from the code

umh-core/pkg/communicator/pkg/generator/topicbrowser.go
```diff
@@ -0,0 +1,329 @@
...
+
+// getPendingBundlesFromObservedState retrieves bundles for existing subscribers.
+//
+// FILTERING LOGIC:
+// - Uses lastSentTimestamp to determine which bundles are still pending
+// - Only includes bundles with timestamp > lastSentTimestamp
+// - This ensures existing subscribers get only new data they haven't seen yet
+//
+// This implements the incremental update strategy for existing subscribers.
+func getPendingBundlesFromObservedState(cache *topicbrowser.Cache, obs *topicbrowser.ObservedState, logger *zap.SugaredLogger) [][]byte {
```
JeremyTheocharis: you could make this function more clear with a slightly revised function signature:

getPendingBundlesFromObservedState(lastSentTimestamp, observed state, logger)

so that you dont send the whole cache in here. could maybve explain better what this function is actually doing

umh-core/pkg/communicator/pkg/generator/topicbrowser.go
```diff
@@ -0,0 +1,329 @@
...
+	// These are incremental updates since the cache was last updated
+	for i, bundle := range tb.UnsBundles {
+		newUnsBundles[i+1] = bundle
+	}
+
+	tb.UnsBundles = newUnsBundles
+	return tb
+}
+
+// generateTbContentForNewSubscriber generates incremental content for new subscribers.
```
JeremyTheocharis: maybe here also think about the function signature and the return value vs the description

it took me soem time to udnerstand that this function generates, AND returns a full TopicBrowser model

maybe it makes sense to restructure it to have two function that fethes the UNSBundle, and then only one conversion from unsbundle to TopicBrowser

umh-core/pkg/communicator/pkg/generator/topicbrowser.go
```diff
@@ -0,0 +1,329 @@
...
+	// Apply bundle count protection
+	return validateAndLimitBundles(newBundles, logger)
+}
+
+// getLastCachedTimestamp extracts the lastCachedTimestamp from the cache.
+//
+// The lastCachedTimestamp represents when the TbCache was last updated with
+// the "one event per topic" snapshot. This timestamp is used to identify
+// which bundles in the observed state are newer than the cached data.
+func getLastCachedTimestamp(cache *topicbrowser.Cache) int64 {
```
JeremyTheocharis: is this function necessary?

umh-core/pkg/communicator/pkg/subscriber/subscribers.go
```diff
@@ -34,7 +35,7 @@ import (
 )
 
 type Handler struct {
-	subscribers                *expiremap.ExpireMap[string, string]
+	subscriberRegistry         *subscribers.Registry
```
JeremyTheocharis, line 38: previously it was an experiemap, which also meant that subscriber could be "expired", e.g., iof they didnt report themselves in for a certain amount. is that still the case? maybe i find out later

umh-core/pkg/communicator/pkg/subscriber/subscribers.go
```diff
@@ -108,40 +109,61 @@ func (s *Handler) notifySubscribers() {
...
-	}
-	if statusMessage == nil {
-		s.logger.Warnf("Failed to generate status message")
-		return
-	}
+	defer cncl()
 
 	notified := 0
-	s.subscribers.Range(func(key string, value string) bool {
+	baseStatusMessage := s.StatusCollector.GenerateStatusMessage(true)
```
JeremyTheocharis, line 121: can you check the performance here and whether it woul be useful to pre-generate the UNS bundles and then only give it here to each subscriber. and in general just have a status message at the beginning, or maybe two different types (new subscriber, existing ones) and then send it out. otherwise we regenerate it here for every subscriber, which could have a measurable impact (because we fetch a lot of system snapshots, deep copy, etc.)

umh-core/pkg/communicator/pkg/subscriber/subscribers.go
```diff
@@ -108,40 +109,61 @@ func (s *Handler) notifySubscribers() {
...
+
+		// Mark subscriber as bootstrapped after first message
+		if !meta.Bootstrapped {
+			s.subscriberRegistry.UpdateMeta(email, func(m *subscribers.Meta) {
+				m.Bootstrapped = true
+			})
+			s.logger.Debugf("Subscriber %s has been bootstrapped", email)
+		} else {
+			// Update last sequence for existing subscribers
+			s.subscriberRegistry.UpdateMeta(email, func(m *subscribers.Meta) {
```
JeremyTheocharis: what happens here?

umh-core/pkg/communicator/subscribers/registry.go
```diff
@@ -0,0 +1,132 @@
...
+import (
+	"sync"
+	"time"
+
+	"github.com/united-manufacturing-hub/expiremap/v2/pkg/expiremap"
+)
+
+// Meta tracks metadata for each subscriber, particularly for topic browser bootstrapping
+type Meta struct {
+	FirstSeen    time.Time // When the subscriber was first added
```
JeremyTheocharis: FirstSeen and LastSeq is not really needed right? only bootstrapped

umh-core/pkg/communicator/subscribers/registry.go
```diff
@@ -0,0 +1,132 @@
...
+	LastSeq      uint64    // Last topic browser sequence delivered
+	Bootstrapped bool      // Did we already send the backlog/full tree?
+}
+
+// Registry wraps expiremap and adds metadata tracking for subscribers
+type Registry struct {
+	subscribers *expiremap.ExpireMap[string, string]
```
JeremyTheocharis: ahh ok, here it is an experimap

but shouldnt then also the metadata expire as well?

umh-core/pkg/communicator/topicbrowser/cache.go
```diff
@@ -0,0 +1,252 @@
...
+		},
+		lastCacheTimestamp: 0,
+		lastSentTimestamp:  0,
+	}
+}
+
+// Buffer represents a compressed data buffer from the FSM
+// This is a placeholder until the actual FSM types are available
+type Buffer struct {
+	Payload   []byte // LZ4 compressed protobuf data
```
JeremyTheocharis: isnt this coming un-LZ4 compressed fromthe FSM?

umh-core/pkg/communicator/topicbrowser/cache.go
```diff
@@ -0,0 +1,252 @@
...
+		sentry.ReportIssueWithContext(err, sentry.IssueTypeError, log, context)
+
+		return nil
+	}
+
+	return encoded
+}
+
+// Snapshot returns a deep copy of all cached event table entries
+func (c *Cache) Snapshot() map[string]*tbproto.EventTableEntry {
```
JeremyTheocharis: can you check where this is sued? this can be potentially quite big and if it iis used regulrly, it might impact performance

umh-core/pkg/communicator/communication_state/communication_state.go
FreiPaul: For the real data, I would use the system snapshot manager that is already accessible in this function to get the real observed snapshot and pass it

umh-core/pkg/communicator/pkg/subscriber/subscribers.go
FreiPaul: This is what the current implementation already does. The baseStatusMessage is generated only once.
I assumed that we do not get a lot of new subscribers per tick so, the generation of the new-sub status message is done in the loop, which in my opinion also makes the code more readable

umh-core/pkg/communicator/communication_state/communication_state.go
```diff
@@ -152,19 +153,14 @@ func (c *CommunicationState) InitialiseAndStartRouter() {
 	c.Router.Start()
 }
 
-func (c *CommunicationState) StartTopicBrowserCacheUpdater(systemSnapshotManager *fsm.SnapshotManager, ctx context.Context) {
+// StartTopicBrowserCacheUpdater starts the topic browser cache updater
+// it is used to update the topic browser cache with the observed state of the topic browser
+// it is also used to run the simulator if the agent is running in simulator mode
+func (c *CommunicationState) StartTopicBrowserCacheUpdater(systemSnapshotManager *fsm.SnapshotManager, ctx context.Context, runSimulator bool) {
```
JeremyTheocharis, line 159: dont want to nitpick too much here, but why is the simulator created if it is only conditionally enabled? couldnt this be all together?

umh-core/pkg/communicator/pkg/generator/topicbrowser.go
```diff
@@ -96,15 +96,25 @@ func GenerateTopicBrowser(cache *topicbrowser.Cache, obs *topicbrowser.ObservedS
...
-		return AddCachedBundleToTbContent(cache, tb)
+		// New subscriber: Get bundles newer than lastCachedTimestamp
+		tb = GenerateTbContent(cache, obs, cache.GetLastCachedTimestamp(), logger)
 	}
+
+	// For new subscribers, prepend the cached bundle (index 0) with complete topic coverage
+	// This ensures new subscribers get both:
+	// 1. Complete topic landscape (from cache at index 0)
+	// 2. Latest incremental updates (from observed state at index 1+)
+	if !isBootstrapped {
```
JeremyTheocharis, line 114: shouldnt this be the other way around? so first check for bootstrapped, and then the rest. also the if / else is not required anymore right?

JeremyTheocharis (approved):

umh-core/pkg/communicator/communication_state/communication_state.go
FreiPaul: This is because we use the (empty) mocked observed in this branch if the simulator is disabled. I can change this in the "implement real observed state" PR

umh-core/pkg/communicator/pkg/subscriber/subscribers.go
Scarjit: The only "in-rush" situation that i can think of if we have many users and an previously offline or new instance connects.

umh-core/pkg/communicator/topicbrowser/simulator.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// HashUNSTableEntry generates an xxHash from the Levels and datacontract.
+// This is used by the frontend to identify which topic an entry belongs to.
+// We use it over full topic names to reduce the amount of data we need to send to the frontend.
+//
+// ✅ FIX: Uses null byte delimiters to prevent hash collisions between different segment combinations.
+// For example, ["ab","c"] vs ["a","bc"] would produce different hashes instead of identical ones.
+// This is a copy of the HashUNSTableEntry function in the benthos topicbrowser plugin.
+func HashUNSTableEntry(info *tbproto.TopicInfo) string {
```
Scarjit, line 142: Just fyi this significantly differs from the old implementation, but should be ok, since we are also using a new frontend.

PR #2111, opened 2025-06-30
ENG-3114 Reconcile inside a errgroup

Scarjit:
This pr changes the base manager to reconcile using multiple goroutines (wrapped in an error group), which should increase the speed.

### DIFF
--- a/umh-core/pkg/fsm/base_manager.go
+++ b/umh-core/pkg/fsm/base_manager.go
@@ -551,7 +551,19 @@ func (m *BaseFSMManager[C]) Reconcile(
 			return nil
 		})
 	}
-	err = errorgroup.Wait()
+
+	waitErrorChannel := make(chan error, 1)
+	go func() {
+		waitErrorChannel <- errorgroup.Wait()
+	}()
+
+	// Wait for either the error group to finish or the context to be done
+	select {
+	case wgErr := <-waitErrorChannel:
+		err = wgErr
+	case <-ctx.Done():
+		err = ctx.Err()
+	}
 
 	instancesToRemoveMutex.Lock()
 	for _, name := range instancesToRemove {


umh-core/pkg/fsm/base_manager.go
```diff
@@ -779,3 +731,76 @@ func (m *BaseFSMManager[C]) maybeEscalateRemoval(ctx context.Context, inst FSMIn
...
+		// For any other error, log and abort reconciliation
+		metrics.IncErrorCount(metrics.ComponentBaseFSMManager, m.managerName)
+		return fmt.Errorf("deadline check error: %w", err), false
+	}
+
+	if !sufficient {
+		m.logger.Warnf("not enough time left to reconcile instance %s (only %v remaining, needed %v), skipping",
+			name, remaining, expectedMaxP95ExecutionTime)
+		return nil, true // return true to indicate that we should not run another manager and instead should wait for the next tick
+	}
```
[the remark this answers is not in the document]
Scarjit: This is unchanged from before and was correct.

umh-core/pkg/fsm/base_manager.go
```diff
@@ -481,83 +483,56 @@ func (m *BaseFSMManager[C]) Reconcile(
...
+			}
+			if shallBeRemoved {
+				instancesToRemoveMutex.Lock()
+				instancesToRemove = append(instancesToRemove, nameCaptured)
+				instancesToRemoveMutex.Unlock()
+			}
+			return nil
+		})
+	}
+	err = errorgroup.Wait()
```
JeremyTheocharis: Can you use the same approach we use everywhere else? the wait is in a seperate go routine, to ensure we can really continue after the context is canceled (if we dont do it, it might wait for a couple of seconds which is what we want to avoid)

JeremyTheocharis: `Use a select statement to wait for either the g.Wait() result or the context's cancellation`

PR #2112, opened 2025-06-30
[ENG-3120] Resubscription Logic

FreiPaul:
This PR also switches to a value type inside the expiremap, which seems more ideomatic, regarding the small footprint of SubscriberData

### DIFF
--- a/umh-core/pkg/communicator/router/router.go
+++ b/umh-core/pkg/communicator/router/router.go
@@ -127,9 +127,6 @@ func (r *Router) handleSub(message *models.UMHMessage, messageContent models.UMH
 		// Try direct type assertion first
 		if payload, ok := messageContent.Payload.(models.SubscribeMessagePayload); ok {
 			subscribePayload = payload
-		} else {
-			r.routerLogger.Warnf("Warning: Could not assert payload to SubscribeMessagePayload or map[string]interface{}. Actual type: %T, Value: %v", messageContent.Payload, messageContent.Payload)
-			return
 		}
 	}
 
--- a/umh-core/pkg/communicator/router/router.go
+++ b/umh-core/pkg/communicator/router/router.go
@@ -133,11 +133,7 @@ func (r *Router) handleSub(message *models.UMHMessage, messageContent models.UMH
 		}
 	}
 
-	if subscribePayload.Resubscribed {
-		r.subHandler.UnexpireSubscriber(message.Email)
-	} else {
-		r.subHandler.AddSubscriber(message.Email)
-	}
+	r.subHandler.AddOrRefreshSubscriber(message.Email, subscribePayload.Resubscribed)
 }
 
 func (r *Router) handleAction(messageContent models.UMHMessageContent, message *models.UMHMessage, watcherUUID uuid.UUID) {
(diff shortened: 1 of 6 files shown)


umh-core/pkg/communicator/subscribers/registry.go
```diff
@@ -75,7 +75,24 @@ func (r *Registry) IsBootstrapped(email string) bool {
...
-	return (*data).Bootstrapped
+	return data.Bootstrapped
+}
+
+// Unexpire resets the TTL for a subscriber by calling Set()
+func (r *Registry) Unexpire(email string) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	if data, exists := r.subscribers.Load(email); exists {
```
JeremyTheocharis: this is now double right? I see two code paths wjere you check iether the subscriber exists and create a new one if it doesnt

umh-core/pkg/communicator/router/router.go
```diff
@@ -109,13 +109,35 @@ func (r *Router) router() {
...
+		if resubscribeValue, exists := payloadMap["resubscribed"]; exists {
+			if resubscribeBool, ok := resubscribeValue.(bool); ok && resubscribeBool {
+				resubscribe = true
+			}
+		}
+	}
+	if resubscribe {
+		r.subHandler.UnexpireSubscriber(message.Email)
+	} else {
+		r.subHandler.AddSubscriber(message.Email)
```
JeremyTheocharis: here for example

umh-core/pkg/communicator/router/router.go
```diff
@@ -109,13 +109,35 @@ func (r *Router) router() {
 	}
 }
 
-func (r *Router) handleSub(message *models.UMHMessage, watcherUUID uuid.UUID) {
+// handleSub handles the subscribe message
+// if the payload contains a "resubscribed" field, it will just "unexpire" the subscriber (reset the TTL)
+// otherwise it will add the subscriber to the registry
+// this is an optimization to avoid sending a "new subscriber" message to the frontend when the user is already subscribed
```
JeremyTheocharis: what do you mean with "new subscriber" message? --> the full UNS with every topic and at least one value per topic. and why should we avoid it --> sending and encoding/decoding requires large amount of network bandwith as well as CPU/memory, so we should prevent it if not necessariely required

umh-core/pkg/communicator/router/router.go
```diff
@@ -109,13 +109,35 @@ func (r *Router) router() {
...
 	}
-	r.subHandler.AddSubscriber(message.Email)
+	resubscribe := false
+	if messageContent.Payload != nil {
+		payloadMap, ok := messageContent.Payload.(map[string]interface{})
+		if !ok {
+			r.routerLogger.Warnf("Warning: Could not assert payload to map[string]interface{}. Actual type: %T, Value: %v", messageContent.Payload, messageContent.Payload)
+			return
+		}
+		if resubscribeValue, exists := payloadMap["resubscribed"]; exists {
```
JeremyTheocharis: is this really the way to do it? is there no schema for this otherwise? it looks quite yolo tbh

umh-core/pkg/communicator/subscribers/registry.go
FreiPaul: At the two places, the exist-check is a safety-mechanism and serves two different purposes. I don't think, we do it more elegant

PR #2114, opened 2025-06-30
Fix LZ4 decode for highly compressible data.

Scarjit:
Fixes decoding by allocation compressed lenght x 255 instead of x4/x8.
This seems to be the upper bound according to https://stackoverflow.com/questions/25740471/lz4-library-decompressed-data-upper-bound-size-estimation


Scarjit:
bugbot run

PR #2115, opened 2025-06-30
fix: resolve Topic Browser documentation inconsistencies

JeremyTheocharis:
- Fix critical GraphQL syntax error in README (array vs object format)
- Add complete configuration paths in consuming-data.md
- Standardize capitalization to 'Topic Browser' throughout
- Ensure consistent path references with ./ prefix

These fixes address broken GraphQL examples and inconsistent documentation that could confuse users attempting to use the Topic Browser functionality.

### DIFF
--- a/umh-core/docs/reference/configuration-reference.md
+++ b/umh-core/docs/reference/configuration-reference.md
@@ -247,6 +247,35 @@ protocolConverter:
               uns: {}
 ```
 
+### GraphQL API - Topic Browser
+
+GraphQL API for querying Unified Namespace topics.
+
+| Field | Type | Default | Description |
+|-------|------|---------|-------------|
+| `enabled` | `bool` | `true` | Enable GraphQL API |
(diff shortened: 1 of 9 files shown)


umh-core/docs/reference/configuration-reference.md
```diff
@@ -259,6 +288,8 @@ internal:
       defaultTopicRetentionMs: 604800000   # 7 days
       maxCores: 1
       memoryPerCoreInBytes: 2147483648
+  topicBrowser:
+    desiredState: active  # Topic Browser service (auto-enabled)
```
Scarjit, line 292: Currently it is not.

Scarjit (approved):

umh-core/docs/reference/configuration-reference.md
JeremyTheocharis: i'd say it is fine like this. we want to enable it in the next days anyway

PR #2120, opened 2025-07-01
[ENG-3135] Parallel managers

Scarjit:
This pr increases performance/reactivity by executing the managers in parallel.
Similar to the base_manager it uses a factor of the real ctx to execute them, ensuring we have some time for cleanup.

### DIFF
--- a/umh-core/pkg/control/loop.go
+++ b/umh-core/pkg/control/loop.go
@@ -316,13 +316,12 @@ func (c *ControlLoop) Reconcile(ctx context.Context, ticker uint64) error {
 	}
 
 	// <factor>% ctx to ensure we finish in time.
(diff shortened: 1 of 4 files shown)


Scarjit:
bugbot run

umh-core/pkg/control/loop.go
```diff
@@ -312,55 +315,73 @@ func (c *ControlLoop) Reconcile(ctx context.Context, ticker uint64) error {
 		}
 	}
 
+	// <factor>% ctx to ensure we finish in time.
+	const factor = 0.95
```
JeremyTheocharis: best into constants package

JeremyTheocharis (approved):

PR #2124, opened 2025-07-02
[ENG-3140] Datamodels Config

FreiPaul:


FreiPaul:
bugbot run

umh-core/pkg/config/config.go
```diff
@@ -40,6 +41,24 @@ type TemplatesConfig struct {
 	ProtocolConverter map[string]interface{} `yaml:"protocolConverter,omitempty"` // Array of protocol converter templates
 }
 
+// DataModelsConfig defines the structure for the data models section
+type DataModelsConfig struct {
+	Name     string                      `yaml:"name"`    // name of the data model
+	Versions map[string]DataModelVersion `yaml:"version"` // version of the data model (v1, v2, etc.)
```
Scarjit: Since this is just counting up, we could just use map[uint64]DataModelVersion

Scarjit (commented):
I would approve the config, but if possible @led0nk should also take a look, since this part is central to all umh-core related changes.

Scarjit (approved):

umh-core/pkg/config/config.go
```diff
@@ -40,6 +41,36 @@ type TemplatesConfig struct {
...
+	Description string                      `yaml:"description,omitempty"` // description of the data model
+	Versions    map[string]DataModelVersion `yaml:"version"`               // version of the data model (1, 2, etc.)
+}
+
+type DataModelVersion struct {
+	Description string           `yaml:"description,omitempty"` // description of the data model version
+	Structure   map[string]Field `yaml:"structure"`             // structure of the data model (fields)
+}
+
+type DataModelVersionRoot struct {
```
led0nk: where will this be used, since the field `Structure` is already used in `DataModelVersion` and this one is currently unused? 
If this will be coming in another PR - feel free to just resolve

FreiPaul: Good point, this is only a leftover from the implementation and can be removed. Just removed it

led0nk (approved):

PR #2125, opened 2025-07-02
Block MacOS's annoying DS_Store

Scarjit:
I don't know why it suddenly tries to insert that everywhere on my machine but it is annoying and provides no value.


led0nk (approved):
don't use macos 🐧

PR #2139, opened 2025-07-09
[ENG-3098] Redpanda schema registry handler

Scarjit:


umh-core/pkg/config/config.go
```diff
@@ -207,6 +207,11 @@ type RedpandaConfig struct {
 
 	// For the Redpanda service
 	RedpandaServiceConfig redpandaserviceconfig.RedpandaServiceConfig `yaml:"redpandaServiceConfig,omitempty"`
+
+	// For the Schema Registry (not marshalled to YAML as these come from the top-level config)
```
JeremyTheocharis: i dont understand this why we have it also here when we already have it in the top level config. is it some intermediary helper? --> some explanation would be nice

umh-core/pkg/fsm/redpanda/manager.go
```diff
@@ -53,6 +53,12 @@ func NewRedpandaManager(name string) *RedpandaManager {
 			redpandaConfig := fullConfig.Internal.Redpanda
 			// Force redpanda name to be "redpanda"
 			redpandaConfig.Name = constants.RedpandaServiceName
+
+			// Store additional config data for schema registry
+			redpandaConfig.DataModels = fullConfig.DataModels
```
JeremyTheocharis: ahh, ok this is why you need it

umh-core/pkg/fsm/redpanda/manager.go
```diff
@@ -85,6 +91,12 @@ func NewRedpandaManager(name string) *RedpandaManager {
 				return fmt.Errorf("instance is not a RedpandaInstance")
 			}
 			RedpandaInstance.config = cfg.RedpandaServiceConfig
+
+			// Set schema registry configuration data
```
JeremyTheocharis: why do we need this?

umh-core/pkg/fsm/redpanda/manager.go
JeremyTheocharis: alternatively you could also fetch it during reconcile time from the observed state snapshot. then you don't have this workaround. fetching it from the observed state is also a workaround, but one that we already used (e.g., in protocol converter templating where we access teh agents location)

umh-core/pkg/service/redpanda/schema_registry.go
```diff
@@ -0,0 +1,1088 @@
...
+	// GetMetrics returns current operational metrics for monitoring and alerting.
+	// All metrics are atomic and thread-safe. Use for health checks and dashboards.
+	GetMetrics() SchemaRegistryMetrics
+}
+
+// SchemaRegistry implements ISchemaRegistry with a multi-phase reconciliation system.
+// It manages schema synchronization between expected state and Redpanda/Kafka registry.
+// All operations are thread-safe and support context cancellation.
+//
+// The reconciliation process follows these phases:
```
JeremyTheocharis, line 145: maybe add a comment why these phases are required in the first place --> all of these steps together take  usually logner than the ticker time, so we need to split it up

JeremyTheocharis (dismissed):

Scarjit:
bugbot run

lefthook.yml
```diff
@@ -29,7 +29,7 @@ pre-push:
       root: "umh-core"
       run: |
         go install go.uber.org/nilaway/cmd/nilaway@latest
-        nilaway -exclude-pkgs="github.com/prometheus/prometheus/util/textparse,github.com/onsi/ginkgo/v2" -exclude-errors-in-files="integration/" ./...
+        nilaway -exclude-pkgs="github.com/prometheus/prometheus/util/textparse,github.com/onsi/ginkgo/v2,go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp,github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/service/redpanda" -exclude-errors-in-files="integration/" ./...
```
Scarjit, line 32: This dependency contains a nilaway violation

FreiPaul (approved):

PR #2140, opened 2025-07-09
[ENG-3168] Stream Processor Config

FreiPaul:


Scarjit (approved):

umh-core/examples/example-config-streamprocessor-templated.yaml
```diff
@@ -0,0 +1,171 @@
...
+          current:
+            _payloadshape: timeseries-number
+          temperature:
+            _payloadshape: timeseries-number
+          status:
+            _payloadshape: timeseries-string
+
+# ---------- Data Contracts ---------- #
+dataContracts:
+  - name: pump-contract
```
JeremyTheocharis: shouldnt this be _pump_v1?

umh-core/examples/example-config-streamprocessor-templated.yaml
```diff
@@ -0,0 +1,171 @@
...
+            _payloadshape: timeseries-number
+          current:
+            _payloadshape: timeseries-number
+          temperature:
+            _payloadshape: timeseries-number
+          status:
+            _payloadshape: timeseries-string
+
+# ---------- Data Contracts ---------- #
+dataContracts:
```
JeremyTheocharis, line 83: maybe also add the _raw example here to check that it is possible to have a data contract wihtout a model?

umh-core/examples/example-config-streamprocessor-templated.yaml
```diff
@@ -0,0 +1,171 @@
...
+    pump-processor:
+      model:
+        name: pump
+        version: v1
+      sources:
+        vibration_sensor: "umh.v1.{{ .location_path }}.vibration"
+        count_sensor: "umh.v1.{{ .location_path }}.count"
+        motor_data: "umh.v1.{{ .location_path }}.motor"
+      mapping:
+        static:
```
JeremyTheocharis: does the difference here between static and dynamic make sense? we could just make it all a dynamic epxression and then just add a static value in there. would fit probably better with the current frontend drafts as well

umh-core/examples/example-config-streamprocessor-templated.yaml
```diff
@@ -0,0 +1,171 @@
...
+          serialNumber: "{{ .SERIAL_NUMBER }}"
+        dynamic:
+          count: "count_sensor.value"
+          vibration:
+            x-axis: "vibration_sensor.x_axis"
```
JeremyTheocharis: the variable here doesnt exist? it only exists vibration_sensor?

umh-core/examples/example-config-streamprocessor-templated.yaml
```diff
@@ -0,0 +1,171 @@
...
+
+# ---------- Templates ---------- #
+templates:
+  streamProcessor:
+    pump-processor:
+      model:
+        name: pump
+        version: v1
+      sources:
+        vibration_sensor: "umh.v1.{{ .location_path }}.vibration"
```
JeremyTheocharis: maybe what needs to be clarified here is that here you can never import a folder, you can only import single variables. or did you discuss it differnetly?

umh-core/examples/example-config-streamprocessor-templated.yaml
```diff
@@ -0,0 +1,171 @@
...
+# ---------- Data Contracts ---------- #
+dataContracts:
+  - name: pump-contract
+    model:
+      name: pump
+      version: v1
+    default_bridges:
+      - type: kafka_to_postgresql
+        config:
+          topic: "{{ .location_path }}/pump"
```
JeremyTheocharis: this is not templatable right?

umh-core/examples/example-config-streamprocessor-templated.yaml
```diff
@@ -0,0 +1,171 @@
+# Copyright 2025 UMH Systems GmbH
```
JeremyTheocharis, line 1: maybe it makes sense here in this PR to also directly update the reference documentation together with the variables section.

FreiPaul: Janik needs it asap to work with it so I'd prefer to ship the doc separately

PR #2141, opened 2025-07-09
[ENG-3152] streamprocessors serviceconfig

led0nk:


led0nk:
closed based on #2140

PR #2144, opened 2025-07-10
[ENG-3184] make storage issue origin clear to redpanda

FreiPaul:


umh-core/pkg/fsm/redpanda/reconcile.go
```diff
@@ -399,15 +399,15 @@ func (r *RedpandaInstance) reconcileTransitionToStopped(ctx context.Context, ser
 			return err, false
 		}
 		// Send event to transition to Stopping
-		r.PreviousObservedState.ServiceInfo.StatusReason = "stopping"
+		r.PreviousObservedState.ServiceInfo.StatusReason = "Redpanda stopping"
```
JeremyTheocharis: i dont think this is the correct place to change it. why not simply change it in the callers? so in protocol converters for example

umh-core/pkg/service/redpanda/redpanda.go
```diff
@@ -82,7 +82,7 @@ type IRedpandaService interface {
 	// It returns:
 	//   ok     – true when metrics are error‑free, false otherwise.
 	//   reason – empty when ok is true; otherwise a short explanation (e.g.
-	//            "storage free space alert: <free>/<total> bytes").
+	//            "Redpanda storage free space alert: <free>/<total> bytes").
```
JeremyTheocharis, line 85: also unsure here, why not add it to the protocol converters?

JeremyTheocharis (changes requested):
Why can't we change it once in the protocol converter? so everytime we are using the redpanda state, just put a redpanda in front of it. should be easier to maintain.

PR #2173, opened 2025-07-21
Fix dlv

FreiPaul:


Scarjit:
This is based on 3196, which is in MQ

PR #2192, opened 2025-07-28
[ENG-3278] increase bundle size limit

FreiPaul:


umh-core/pkg/communicator/topicbrowser/communicator.go
```diff
@@ -32,8 +32,8 @@ const (
 	// MaxTopicCount limits the number of topics to prevent memory exhaustion
 	MaxTopicCount = 1_000_000
 
-	// MaxBundleSize limits individual bundle size to 10MB
-	MaxBundleSize = 10 * 1024 * 1024
+	// MaxBundleSize limits individual bundle size to 50MB
+	MaxBundleSize = 50 * 1024 * 1024
```
JeremyTheocharis, line 36: is this then adjusted with the rest of the limits? For example in benthos or FSM? so that we dont have this issue again?

JeremyTheocharis (changes requested):

PR #2193, opened 2025-07-28
[ENG-3277] only use new logs for error detection

FreiPaul:


umh-core/pkg/communicator/actions/edit-dataflowcomponent.go
```diff
@@ -132,6 +132,8 @@ type EditDataflowComponentAction struct {
 	// Caches for rollback: we hold the freshly generated config and the old one
 	dfc       config.DataFlowComponentConfig // desired (new) config
 	oldConfig config.DataFlowComponentConfig // backup of the original config
+
+	configChangeAt time.Time
```
JeremyTheocharis: Can you document this variable? what it does, why it is important, and also that this should be replaced by archive storage

umh-core/pkg/communicator/actions/edit-dataflowcomponent.go
```diff
@@ -625,9 +628,11 @@ func (a *EditDataflowComponentAction) waitForComponentToBeReady(ctx context.Cont
 							// send the benthos logs to the user
 							logs = dfcSnapshot.ServiceInfo.BenthosObservedState.ServiceInfo.BenthosStatus.BenthosLogs
 
+							// clear the logs that are older than the config change timestamp
```
JeremyTheocharis: can you explain the why? so why do we want to clear logs that are older than the config change timestamp

umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -111,6 +111,8 @@ type EditProtocolConverterAction struct {
 	atomicEditUUID uuid.UUID
 
 	actionLogger *zap.SugaredLogger
+
+	configChangeAt time.Time
```
JeremyTheocharis: can you document this variable?

umh-core/pkg/service/s6/s6.go
```diff
@@ -83,21 +83,22 @@ func (h HealthStatus) String() string {
...
+	ExitCode           int           // Exit code if service is down
+	WantUp             bool          // Whether the service wants to be up (based on existence of down file)
+	IsPaused           bool          // Whether the service is paused
+	IsFinishing        bool          // Whether the service is shutting down
+	IsWantingUp        bool          // Whether the service wants to be up (based on flags)
+	IsReady            bool          // Whether the service is ready
+	ExitHistory        []ExitEvent   // History of exit codes
+	LastChangedAt      time.Time     // Timestamp when the service status last changed
+	LastReadyAt        time.Time     // Timestamp when the service was last ready
+	LastConfigChangeAt time.Time     // Timestamp when the service config last changed
```
JeremyTheocharis: can you again document it here?

umh-core/pkg/service/s6/lifecycle.go
```diff
@@ -127,6 +127,8 @@ func (s *DefaultService) CreateArtifacts(ctx context.Context, servicePath string
 	// Store the created files in artifacts (repository paths)
 	artifacts.CreatedFiles = createdFiles
 
+	s.LastConfigChangeAt = time.Now().UTC()
```
JeremyTheocharis: this need some explanation here, also why is it here UTC time and evertwhere else it is not?

umh-core/pkg/service/s6/s6.go
```diff
@@ -1192,6 +1213,17 @@ func (s *DefaultService) GetLogs(ctx context.Context, servicePath string, fsServ
 		copy(out, st.logs[:st.head])
 	}
 
+	// filter the logs to only include logs since last deployment time
+	if !getLastDeployedTime(servicePath).IsZero() {
+		filtered := out[:0] // reuse slice to avoid allocation
+		for _, entry := range out {
```
JeremyTheocharis: you could add here one optimization: if we once reach a good timestamp, we can simply re-use all the other entries as they are already time sorted and dont have check all the other timestamps

umh-core/pkg/service/s6/s6.go
```diff
@@ -1192,6 +1213,17 @@ func (s *DefaultService) GetLogs(ctx context.Context, servicePath string, fsServ
 		copy(out, st.logs[:st.head])
 	}
 
+	// filter the logs to only include logs since last deployment time
+	if !getLastDeployedTime(servicePath).IsZero() {
+		filtered := out[:0] // reuse slice to avoid allocation
```
JeremyTheocharis: i am not exactly sure this works here as intended

umh-core/pkg/service/s6/status.go
```diff
@@ -161,15 +161,16 @@ func parseS6StatusFile(ctx context.Context, statusFilePath string, fsService fil
...
+		Status:             ServiceUnknown,
+		Pid:                statusData.Pid,
+		Pgid:               statusData.Pgid,
+		IsPaused:           statusData.IsPaused,
+		IsFinishing:        statusData.IsFinishing,
+		IsWantingUp:        statusData.IsWantingUp,
+		IsReady:            statusData.IsReady,
+		LastChangedAt:      statusData.StampTime,
+		LastReadyAt:        statusData.ReadyTime,
+		LastConfigChangeAt: getLastDeployedTime(servicePath),
```
JeremyTheocharis: can we rename the variable to LastDeploymentTime and the function to getLastDeploymentTime?

JeremyTheocharis (changes requested):
looks much cleaner now! will also propose something

JeremyTheocharis (approved):

PR #2195, opened 2025-07-30
Use OS port allocation instead of range-based port management

Scarjit:
- Replace range-based port allocation with OS port allocation (port 0)
- Remove minPort, maxPort, nextPort fields from DefaultPortManager
- Close listeners immediately after port allocation to allow external usage
- Add proper error handling for listener.Close() operations
- Update NewDefaultPortManager() to remove port range parameters
- Update tests to work with OS-allocated ports

### DIFF
--- a/umh-core/pkg/portmanager/portmanager.go
+++ b/umh-core/pkg/portmanager/portmanager.go
@@ -18,6 +18,7 @@ package portmanager
 import (
 	"context"
 	"fmt"
+	"net"
 	"sync"
 )
 
@@ -50,7 +51,7 @@ type PortManager interface {
 }
 
 // DefaultPortManager is a thread-safe implementation of PortManager
-// that keeps track of ports in a simple in-memory store
+// that uses OS port allocation (port 0) to get available ports from the OS
 type DefaultPortManager struct {
 	// mutex to protect concurrent access to maps
 	mutex sync.RWMutex
@@ -58,13 +59,8 @@ type DefaultPortManager struct {
 	// instanceToPorts maps instance names to their allocated ports
 	instanceToPorts map[string]uint16
 
-	// portToInstances maps ports to instance names
+	// portToInstances maps ports to instance names for reverse lookup
 	portToInstances map[uint16]string
-
-	// configuration
-	minPort  uint16
-	maxPort  uint16
-	nextPort uint16
 }
 
 // Global singleton instance of DefaultPortManager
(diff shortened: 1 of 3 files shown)


umh-core/pkg/portmanager/portmanager.go
```diff
@@ -82,100 +93,79 @@ func GetDefaultPortManager() *DefaultPortManager {
...
+					return uint16(min), uint16(max)
+				}
+			}
 		}
-		return existing, nil
 	}
 
-	// Initialize singleton if it doesn't exist
-	return initDefaultPortManager(minPort, maxPort)
+	// Fallback to typical ephemeral port range
```
JeremyTheocharis, line 160: maybe some warning + sentry issue in this case?

umh-core/pkg/portmanager/portmanager.go
```diff
@@ -283,72 +264,47 @@ func (pm *DefaultPortManager) ReservePort(instanceName string, port uint16) erro
 		return nil
 	}
 
-	// Reserve the port
+	// Try to bind to the specific port to verify it's available
+	addr := fmt.Sprintf(":%d", port)
+	listener, err := net.Listen("tcp", addr)
```
JeremyTheocharis: can we check the impact here on the context deadline? could it be that this check takes a lot of time?

umh-core/pkg/portmanager/portmanager.go
```diff
@@ -83,101 +93,83 @@ func GetDefaultPortManager() *DefaultPortManager {
...
-			return existing, fmt.Errorf(
-				"warning: using existing port manager with different range (%d-%d) than requested (%d-%d)",
-				existing.minPort, existing.maxPort, minPort, maxPort,
-			)
+// getEphemeralPortRange returns the OS ephemeral port range.
+// On Linux, it reads from /proc/sys/net/ipv4/ip_local_port_range.
+// Falls back to default range 32768-65535 if unable to read from OS.
+func getEphemeralPortRange() (uint16, uint16) {
+	// Try to read from Linux proc filesystem
+	data, err := os.ReadFile("/proc/sys/net/ipv4/ip_local_port_range")
```
JeremyTheocharis: use our filesystem apckage for context

JeremyTheocharis (changes requested):

umh-core/pkg/portmanager/portmanager.go
Scarjit: done, iv choosen a more then reasonable timeout, as this function is only invoked once at startup.

led0nk (approved):
looks good, was wondering of we should give it more retries to allocate the port

PR #2197, opened 2025-07-30
Let "flows" in bridges only start after the connection is successfully up

JeremyTheocharis:
Fixes ENG-3223


umh-core/pkg/fsm/protocolconverter/reconcile.go
```diff
@@ -265,12 +265,10 @@ func (p *ProtocolConverterInstance) reconcileTransitionToActive(ctx context.Cont
...
-		// Attempt to initiate start
-		if err := p.StartInstance(ctx, services.GetFileSystem()); err != nil {
-			return err, false
-		}
-		p.ObservedState.ServiceInfo.StatusReason = "started"
-		// Send event to transition from Stopped to Starting
+		// Send event to transition from Stopped to StartingConnection
+		// The connection will be started by StartConnectionInstance in the starting_connection state
+		// This prevents Benthos from connecting and sending data when the connection is flaky or filtered
+		p.ObservedState.ServiceInfo.StatusReason = "starting"
```
led0nk: we should maybe then use "starting_connection" here also ?

umh-core/pkg/fsm/protocolconverter/reconcile.go
```diff
@@ -298,7 +296,13 @@ func (p *ProtocolConverterInstance) reconcileStartingStates(ctx context.Context,
 
 	switch currentState {
 	case OperationalStateStartingConnection:
-		// First we need to ensure the S6 service is started
+		// Start the connection component first
+		if err := p.StartConnectionInstance(ctx, services.GetFileSystem()); err != nil {
```
led0nk, line 299: IMO we call this every reconcile-loop whenever the connection is not yet up - meaning this results in ranging through all connections, that are part of the config -> could be later up to `n`. Not sure if we waste too much time then here and should only do this once?

umh-core/pkg/fsm/protocolconverter/reconcile.go
```diff
@@ -345,6 +349,12 @@ func (p *ProtocolConverterInstance) reconcileStartingStates(ctx context.Context,
 			return p.baseFSMInstance.SendEvent(ctx, EventStartFailedDFCMissing), true
 		}
 
+		// Start the DFC components now that prerequisites are met
+		if err := p.StartDFCInstance(ctx, services.GetFileSystem()); err != nil {
```
led0nk, line 358: some kind of same here - we should at least put it into the `running` check of each component (dfc/connection) otherwise we always loop through the configs and reapply it

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -700,12 +715,25 @@ func (p *ProtocolConverterService) EvaluateDFCDesiredStates(protConvName string,
...
-					p.dataflowComponentConfig[i].DesiredFSMState = dfcfsm.OperationalStateActive
+					// CRITICAL: Only set DFC to active when FSM state indicates connection is up
+					// This prevents the issue where Benthos might connect and send data even when 
+					// the connection is flaky or filtered, leading to unreliable data transmission
+					// NOTE: Hardcoded strings to avoid circular import with pkg/fsm/protocolconverter
+					// These correspond to: OperationalStateStartingDFC, OperationalStateIdle, OperationalStateActive
+					if currentFSMState == "starting_dfc" || 
+					   currentFSMState == "idle" || 
+					   currentFSMState == "active" ||
+					   strings.HasPrefix(currentFSMState, "degraded") {
```
led0nk: what if the state is `OperationalStateDegradedConnection`? this wouldn't be a healthy state for dfc-starting

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -718,12 +746,25 @@ func (p *ProtocolConverterService) EvaluateDFCDesiredStates(protConvName string,
...
```
led0nk: same here

umh-core/pkg/service/protocolconverter/mock.go
```diff
@@ -538,7 +542,7 @@ func (m *MockProtocolConverterService) ReconcileManager(ctx context.Context, ser
 // EvaluateDFCDesiredStates mocks the DFC state evaluation logic.
 // This method exists because protocol converters must re-evaluate DFC states
 // when configs change during reconciliation (unlike other FSMs that set states once).
-func (m *MockProtocolConverterService) EvaluateDFCDesiredStates(protConvName string, protocolConverterDesiredState string) error {
+func (m *MockProtocolConverterService) EvaluateDFCDesiredStates(protConvName string, protocolConverterDesiredState string, currentFSMState string) error {
```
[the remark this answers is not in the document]
led0nk: fair point

led0nk (changes requested):

umh-core/pkg/fsm/protocolconverter/reconcile.go
JeremyTheocharis: each function is idempotent, and it is a relatively low performance impact. as soon as it is started, and confirmed to be started, we will move into another state anyway. so i would leave it as it is

umh-core/pkg/fsm/protocolconverter/reconcile.go
JeremyTheocharis: i would leave it as it is

umh-core/pkg/service/protocolconverter/protocolconverter.go
JeremyTheocharis: not a big impact as degraded indicates that the connection was at least once successfully up and I think this code path is never reachable there but implemented it anyway in case I made a mistake in the mental model here.

umh-core/pkg/service/protocolconverter/protocolconverter.go
JeremyTheocharis: see above

led0nk (approved):

PR #2234, opened 2025-09-01
[Bridge Fixes 2.1] Sizing guidelines

Scarjit:
This pr provides a guideline on how many cores of what type are required for a specifc number of bridges.


JeremyTheocharis (changes requested):
Adding performance benchmarks makes it feel like a test report, but not a guideline on what we guarantee to the users and what they should do in what case. Diminishing returns undermines our scalability, we can ignore that entirely. maybe jsut mention bridges up to 30? 

better would be to adjust teh first paragraph of the sizing guidelines and check if this is still the case with the 9 bridges with OPC UA. and then provide explanation on how the users should size their machines.

PR #2242, opened 2025-09-05
fix(force-removal): ensure cleanup succeeds even with expired context

JeremyTheocharis:
## Summary
Fixes force removal operations that were failing when parent context expired, preventing cleanup of failed instances.

## Problem
When instances had permanent failures (after max retries), the system would try to force remove them but fail immediately if the parent context had expired. This prevented cleanup and caused resource leaks.

## Solution
Similar to the FSM transition fix in #2241, this gives force removal operations a fresh context when the parent has expired:
- Check if parent context is expired
- If expired, create fresh 5-second context for cleanup
- Log warning and proceed with force removal

## Changes
- Add fresh context logic to force removal functions:
  - `ForceRemoveProtocolConverter`
  - `ForceRemoveDataFlowComponent`
  - `ForceRemoveConnection`
  - `S6 ForceRemove`
- Centralize timeout constants in `pkg/constants/fsm.go`
- Add proper logging when using fresh context

## Technical Details
The force removal functions were checking `ctx.Err()` at the beginning and immediately returning if context was expired, without performing any cleanup. Now they detect this condition and create a fresh context to ensure cleanup can proceed.

## Linear Issues
- Fixes [ENG-3420](https://linear.app/united-manufacturing-hub/issue/ENG-3420)
- Parent issue: [ENG-3393](https://linear.app/united-manufacturing-hub/issue/ENG-3393)

## Dependencies
This PR is stacked on top of #2241 (FSM context fix) and should be merged after it.

## Test Plan
- [x] Unit tests pass
- [x] Linter passes
- [ ] Deploy and verify force removal succeeds under high load
- [ ] Confirm no more "failed to force remove" errors in logs


JeremyTheocharis:
@coderabbitai review

PR #2259, opened 2025-09-12
fix: implement DirectoryHealth detection for S6 service self-healing

JeremyTheocharis:
## Summary
This PR implements directory health detection for S6 services to automatically recover when service directories are missing or corrupted. This fixes the critical issue where FSMs get stuck in infinite loops showing "stopping: not existing" when S6 service directories are deleted or force-removed.

**Fixes:**
- [[ENG-3468](https://linear.app/united-manufacturing-hub/issue/ENG-3468)](https://linear.app/united-manufacturing-hub/issue/ENG-3468): Parent FSM stuck when S6 service returns empty string "not existing"
- [[ENG-3469](https://linear.app/united-manufacturing-hub/issue/ENG-3469)](https://linear.app/united-manufacturing-hub/issue/ENG-3469): Add S6 directory diagnostic logging for troubleshooting

## Context: The User-Visible Problem

Users are experiencing services stuck in states like:
- `"stopping: S6 is not stopped, current state: not existing"`
- Services showing green/active in UI while processes are actually dead
- Bridge FSMs stuck in "Starting" while underlying services are in "Stopping"

This creates a cascade of issues where services can't recover and require manual intervention.

### Experiment 1: Manual Directory Deletion
- Deleted individual files → Not detected
- Deleted supervise directory → Not detected
- Deleted entire service directory → Detected as "service does not exist" error
- **Finding**: No automatic recovery mechanism exists

### Experiment 2: Directory Deletion + State Change
- Deleted directory, then set desired state to stopped
- Result: UI stays green, but internally shows "fail to update observed state: service does not exist"
- When transitioning states: Gets stuck with "stopping: not existing"
- **Finding**: State transitions fail silently when directories are missing

### Experiment 3: Self-Healing Attempt
- Tried to make FSM self-remove when detecting bad directory
- **Discovery**: FSMs can't self-remove! Manager immediately overrides stop→running
- Created [[ENG-3473](https://linear.app/united-manufacturing-hub/issue/ENG-3473)](https://linear.app/united-manufacturing-hub/issue/ENG-3473) to track this architectural limitation

## Root Cause Analysis

The issue occurs through this sequence:
1. High system load causes status update failures (5x in a row)
2. System triggers force removal to clean up
3. Directory gets deleted but FSM remains in "running" state
4. User deploys new config or system tries to stop service
5. FSM enters "stopping" state but can't verify stop completion
6. Stuck in infinite loop: "not existing" error

### 1. Directory Health Detection (Core Fix)
- Added `DirectoryHealth` field to S6ObservedState with three states:
  - `HealthOK`: All tracked files exist
  - `HealthBad`: Missing/corrupted directory
  - `HealthUnknown`: I/O errors or untracked state (e.g., after restart)
- Check integrity on every tick in `UpdateObservedStateOfInstance`
- Handle unhealthy directories based on current FSM state

### 2. Comprehensive Diagnostic Logging (ENG-3469)
- Added `logS6DirectoryState()` function that captures:
  - Directory structure (exists, supervise, status, run, down files)
  - Process state (PID, running status)
  - S6 service info (uptime, exit history, etc.)
  - FSM state context
- Triggers on: empty state, "not existing" errors, degraded state, stopping failures
- Performance optimized: Only logs during error conditions

#### Config File Path Bug
```go
// BUG: Tracked with relative path
createdFiles = append(createdFiles, filepath.Join("config", originalPath))
// FIX: Track with absolute path
createdFiles = append(createdFiles, path)
```

#### Down File Tracking Issue
```go
// Down files are temporary control files - should NOT be tracked
// They're removed when starting, created when stopping
// After agent restart, missing down file ≠ unhealthy directory
```

#### artifacts==nil Interpretation
```go
// BUG: Returned HealthBad when artifacts==nil
// FIX: Return HealthUnknown (normal during init/restart)
```
### Cold Start Test ✅
- Services reach active in ~14s

### Restart Test ✅
- 5 transient errors (expected)
- Recovery within 2-3 cycles
- No false positives

### Missing File Detection ✅
- Correctly identifies missing files
- Comprehensive diagnostic logs
- ⚠️ Full recovery blocked by manager (ENG-3473)

## Known Limitations

**Manager Override Issue (ENG-3473)**
- FSM detects bad directory → sets state to "stopped"
- Manager sees config says "running" → forces back to running
- Creates ping-pong effect
- **Workaround**: For "stopping" state, we consider bad directory as successfully stopped

## Review Guide

1. **Start with models.go**: New DirectoryHealth field
2. **Then s6/lifecycle.go**: Down file management explanation (line 383+)
3. **Check s6/actions.go**: 
   - CreateInstance verification
   - UpdateObservedStateOfInstance changes
4. **Review s6/reconcile.go**: handleUnhealthyDirectory logic
5. **Finally benthos/**: Diagnostic logging implementation

## Related Issues
- Prerequisite for proper fix: [[ENG-3473](https://linear.app/united-manufacturing-hub/issue/ENG-3473)](https://linear.app/united-manufacturing-hub/issue/ENG-3473) (Manager health awareness)
- Load-related trigger: [[ENG-3417](https://linear.app/united-manufacturing-hub/issue/ENG-3417)](https://linear.app/united-manufacturing-hub/issue/ENG-3417) (Prevent force removal under load)
- Parent issue: [[ENG-3400](https://linear.app/united-manufacturing-hub/issue/ENG-3400)](https://linear.app/united-manufacturing-hub/issue/ENG-3400) (Grey state investigation)

### DIFF
--- a/umh-core/internal/fsm/baseFSM.go
+++ b/umh-core/internal/fsm/baseFSM.go
@@ -446,7 +446,16 @@ func (s *BaseFSMInstance) HandlePermanentError(
 
 		// After successful force removal, set FSM state directly to removed
 		// This will cause ReconcileLifecycleStates to return ErrInstanceRemoved,
-		// signaling the manager to delete and recreate the instance
+		// signaling the manager to delete and recreate the instance.
+		//
+		// NOTE: This is an intentional violation of normal FSM state transition rules.
+		// We directly manipulate the state here because:
+		// 1. This is an emergency recovery path after force removal
+		// 2. Normal state transitions have already failed or are impossible
+		// 3. The instance will be deleted and recreated by the manager anyway
+		// 4. The mutex in SetCurrentFSMState prevents data races
+		// This is safe because force removal only happens when the instance is corrupted
+		// and needs to be completely recreated.
 		s.SetCurrentFSMState(LifecycleStateRemoved)
 		logger.Infof("%s instance %s force removed and transitioned to removed state", s.cfg.ID, instanceID)
 		return err, true
@@ -471,6 +480,7 @@ func (s *BaseFSMInstance) HandlePermanentError(
 				return fmt.Errorf("failed to force remove the %s instance: %s : %w", s.cfg.ID, backoff.PermanentFailureError, forceErr), false
 			}
 			// After successful force removal, set FSM state directly to removed
+			// (See comment above for why this direct state manipulation is safe)
 			s.SetCurrentFSMState(LifecycleStateRemoved)
 			logger.Infof("%s instance %s force removed after normal removal failed and transitioned to removed state", s.cfg.ID, instanceID)
 		}
--- a/umh-core/internal/fsm/baseFSM.go
+++ b/umh-core/internal/fsm/baseFSM.go
@@ -248,12 +248,12 @@ func (s *BaseFSMInstance) SendEvent(ctx context.Context, eventName string, args 
 
 	// Execute the FSM transition with guaranteed time to complete
 	err := s.fsm.Event(fsmCtx, eventName, args...)
-	
+
 	// Check if parent context expired while we were executing
 	if err == nil && ctx.Err() != nil {
 		s.logger.Warnf("FSM transition completed successfully but parent context expired - prevented stuck transition but now outside of cycle time (preventing bigger impact)")
 	}
-	
+
 	if err != nil {
 		// Enhanced error message with state context
 		enhancedErr := fmt.Errorf("FSM %s failed transition: current_state='%s' -> event='%s' (desired_state='%s'): %w",
@@ -458,6 +458,7 @@ func (s *BaseFSMInstance) HandlePermanentError(
 		// and needs to be completely recreated.
 		s.SetCurrentFSMState(LifecycleStateRemoved)
 		logger.Infof("%s instance %s force removed and transitioned to removed state", s.cfg.ID, instanceID)
+
 		return err, true
 	} else {
 		// Not in a shutdown state yet, so try normal removal first
--- a/umh-core/internal/fsm/baseFSM.go
+++ b/umh-core/internal/fsm/baseFSM.go
@@ -444,6 +444,11 @@ func (s *BaseFSMInstance) HandlePermanentError(
 			return fmt.Errorf("failed to force remove the %s instance: %s : %w", s.cfg.ID, backoff.PermanentFailureError, forceErr), false
 		}
 
+		// After successful force removal, set FSM state directly to removed
+		// This will cause ReconcileLifecycleStates to return ErrInstanceRemoved,
+		// signaling the manager to delete and recreate the instance
+		s.SetCurrentFSMState(LifecycleStateRemoved)
+		logger.Infof("%s instance %s force removed and transitioned to removed state", s.cfg.ID, instanceID)
 		return err, true
 	} else {
 		// Not in a shutdown state yet, so try normal removal first
@@ -465,6 +470,9 @@ func (s *BaseFSMInstance) HandlePermanentError(
 
 				return fmt.Errorf("failed to force remove the %s instance: %s : %w", s.cfg.ID, backoff.PermanentFailureError, forceErr), false
 			}
+			// After successful force removal, set FSM state directly to removed
+			s.SetCurrentFSMState(LifecycleStateRemoved)
+			logger.Infof("%s instance %s force removed after normal removal failed and transitioned to removed state", s.cfg.ID, instanceID)
 		}
 
 		return nil, false // Let's try to at least reconcile towards a stopped/removed state
--- a/umh-core/pkg/fsm/benthos/actions.go
+++ b/umh-core/pkg/fsm/benthos/actions.go
@@ -50,6 +50,7 @@ import (
 
 // logS6DirectoryState provides comprehensive diagnostic logging when S6 state is empty or problematic.
 // This helps troubleshoot "not existing" errors (ENG-3468) by capturing the full state of S6 directories.
+// This function spawns a goroutine for async logging to avoid blocking the reconciliation loop.
 func (b *BenthosInstance) logS6DirectoryState(ctx context.Context, trigger string) {
 	// Only log when S6FSMState is empty or shows problems
 	s6State := b.ObservedState.ServiceInfo.S6FSMState
@@ -57,6 +58,14 @@ func (b *BenthosInstance) logS6DirectoryState(ctx context.Context, trigger strin
 		return
 	}
 
+	// Spawn goroutine for async diagnostic logging and Sentry reporting
+	// This prevents blocking the reconciliation loop with I/O operations
+	go b.logS6DirectoryStateAsync(trigger, s6State)
+}
+
+// logS6DirectoryStateAsync performs the actual diagnostic logging in a goroutine.
+// Uses context.Background() since this is fire-and-forget async logging.
+func (b *BenthosInstance) logS6DirectoryStateAsync(trigger string, s6State string) {
 	serviceName := b.baseFSMInstance.GetID()
 	// Convert to S6 service name format (benthos- prefix)
 	s6ServiceName := "benthos-" + serviceName
--- a/umh-core/pkg/fsm/benthos/actions.go
+++ b/umh-core/pkg/fsm/benthos/actions.go
@@ -53,7 +53,7 @@ import (
 
 // logS6DirectoryState provides comprehensive diagnostic logging when S6 state is empty or problematic.
 // This helps troubleshoot "not existing" errors (ENG-3468) by capturing the full state of S6 directories.
-// This function spawns a goroutine for async logging to avoid blocking the reconciliation loop.
+// This function spawns a short-lived goroutine for async logging to avoid blocking the reconciliation loop.
 func (b *BenthosInstance) logS6DirectoryState(ctx context.Context, trigger string) {
 	// Only log when S6FSMState is empty or shows problems
 	s6State := b.ObservedState.ServiceInfo.S6FSMState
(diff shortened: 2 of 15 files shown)


umh-core/pkg/fsm/benthos/actions.go
```diff
@@ -45,6 +47,166 @@ import (
...
+		if _, err := os.Stat(procPath); err == nil {
+			processRunning = true
+		}
+	}
+	
+	// Extract S6 service info for comprehensive logging
+	s6Info := b.ObservedState.ServiceInfo.S6ObservedState.ServiceInfo
+	
+	// Log comprehensive state - this is critical for debugging
+	b.baseFSMInstance.GetLogger().Warnf(
```
FreiPaul: Change log level to debug?

umh-core/pkg/fsm/benthos/reconcile.go
```diff
@@ -473,6 +479,11 @@ func (b *BenthosInstance) reconcileTransitionToStopped(ctx context.Context, serv
 	isStopped, reason := b.IsBenthosS6Stopped()
 	if currentState == OperationalStateStopping {
 		if !isStopped {
+			// Log diagnostic info when stuck in stopping state
+			if strings.Contains(reason, "not existing") {
```
FreiPaul: maybe we can define the not existing state string as a constant

umh-core/pkg/fsm/benthos/actions.go
```diff
@@ -301,6 +463,9 @@ func (b *BenthosInstance) IsBenthosS6Running() (bool, string) {
 
 	currentState := b.ObservedState.ServiceInfo.S6FSMState
 	if currentState == "" {
+		// Log diagnostic info when we encounter empty S6 state
+		b.logS6DirectoryState(context.Background(), "IsBenthosS6Running_empty_state")
```
FreiPaul: should we really give it background context?

umh-core/pkg/fsm/benthos/actions.go
```diff
@@ -323,6 +488,9 @@ func (b *BenthosInstance) IsBenthosS6Stopped() (bool, string) {
 	fsmState := b.ObservedState.ServiceInfo.S6FSMState
 	switch fsmState {
 	case "":
+		// Log diagnostic info when we encounter empty S6 state
+		b.logS6DirectoryState(context.Background(), "IsBenthosS6Stopped_empty_state")
```
FreiPaul: should we really give it background context?

umh-core/internal/fsm/baseFSM.go
```diff
@@ -444,6 +444,11 @@ func (s *BaseFSMInstance) HandlePermanentError(
 			return fmt.Errorf("failed to force remove the %s instance: %s : %w", s.cfg.ID, backoff.PermanentFailureError, forceErr), false
 		}
 
+		// After successful force removal, set FSM state directly to removed
+		// This will cause ReconcileLifecycleStates to return ErrInstanceRemoved,
+		// signaling the manager to delete and recreate the instance
+		s.SetCurrentFSMState(LifecycleStateRemoved)
```
FreiPaul, line 459: Can this direct fsm state assetion cause race conditions when another transition is active?
If you are sure about this, maybe add a comment that we can safely violate FSM rule #9 here

umh-core/pkg/fsm/benthos/actions.go
JeremyTheocharis: Fixed in commit a1f7d2fdb - Changed to DEBUG level as suggested. Also added Sentry reporting to ensure we still capture diagnostics in production (where customers run at INFO level). The Sentry reporting uses the built-in 2hr debounce to prevent spam. Added TODO(ENG-3468) to remove once the directory corruption issues are resolved.

umh-core/pkg/fsm/benthos/reconcile.go
JeremyTheocharis: Fixed in commit 212f774a8 - Created  constant in  with documentation explaining when this state occurs (service removed, directory corrupted, or never created). This eliminates the magic string and provides clarity about what 'not existing' actually represents in the FSM context.

JeremyTheocharis: Correction to my previous comment: Fixed in commit 53f24019b - Created S6StateNotExisting constant in pkg/service/benthos/errors.go with documentation explaining when this state occurs (service removed, directory corrupted, or never created). This eliminates the magic string and provides clarity about what 'not existing' actually represents in the FSM context.

umh-core/pkg/fsm/benthos/actions.go
JeremyTheocharis: Fixed in commit 3690ad1e5 - Made the diagnostic logging async by spawning a goroutine. This way we legitimately use context.Background() for fire-and-forget operations while preventing I/O from blocking the reconciliation loop. The main function still respects the context parameter, but the actual logging happens asynchronously.

umh-core/pkg/fsm/benthos/actions.go
JeremyTheocharis: Same fix as above - commit 3690ad1e5 makes all diagnostic logging async.

umh-core/internal/fsm/baseFSM.go
JeremyTheocharis: Fixed in commit 5d60107ee - Added comprehensive documentation explaining why this direct state manipulation is safe and intentional.

Regarding FSM rule #9: I couldn't find this rule documented in our codebase. Could you clarify what rule #9 refers to? Is this from our team conventions or a general FSM principle?

The SetCurrentFSMState usage here is indeed a deliberate violation of normal state transition rules, but it's safe because:
1. It's only used in the force removal error recovery path
2. Normal transitions have already failed at this point
3. The instance will be deleted and recreated by the manager
4. The method has mutex protection against data races

This is essentially our 'break glass' emergency procedure for corrupted instances. The documentation now makes this explicit.

umh-core/internal/fsm/baseFSM.go
```diff
@@ -444,6 +444,11 @@ func (s *BaseFSMInstance) HandlePermanentError(
...
+		logger.Infof("%s instance %s force removed and transitioned to removed state", s.cfg.ID, instanceID)
 		return err, true
```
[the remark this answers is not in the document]
JeremyTheocharis: This suggestion is not needed. After a successful force removal, the FSM instance is deleted and recreated by the manager (as stated in the comment). When the instance is deleted, all associated state including backoff is cleared. The return of the original error with handled=true signals the manager to perform this deletion. 

Setting the state to LifecycleStateRemoved causes ReconcileLifecycleStates to return ErrInstanceRemoved, which triggers the manager to delete and recreate the entire instance. This is the intended behavior for handling corrupted instances that cannot be recovered through normal means.

umh-core/internal/fsm/baseFSM.go
```diff
@@ -465,6 +470,9 @@
 
 				return fmt.Errorf("failed to force remove the %s instance: %s : %w", s.cfg.ID, backoff.PermanentFailureError, forceErr), false
 			}
+			// After successful force removal, set FSM state directly to removed
+			s.SetCurrentFSMState(LifecycleStateRemoved)
+			logger.Infof("%s instance %s force removed after normal removal failed and transitioned to removed state", s.cfg.ID, instanceID)
 		}
```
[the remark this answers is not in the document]
JeremyTheocharis: Same as above - this suggestion is not needed. The FSM instance is deleted and recreated by the manager after setting state to LifecycleStateRemoved. All state including backoff is cleared when the instance is deleted. This is the intended design for handling corrupted instances.

umh-core/pkg/fsm/s6/actions.go
```diff
@@ -198,39 +212,69 @@ func (s *S6Instance) UpdateObservedStateOfInstance(ctx context.Context, services
...
 
-		err := s.baseFSMInstance.Remove(ctx)
-		if err != nil {
-			s.baseFSMInstance.GetLogger().Errorf("error removing S6 instance %s: %v", s.baseFSMInstance.GetID(), err)
-
-			return err
+				return err
+			}
 		}
 	}
```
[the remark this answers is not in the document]
JeremyTheocharis: Good suggestion for improvement. However, this is in the S6 FSM which currently uses reflect.DeepEqual consistently. For this PR focused on Benthos FSM fixes, I'll maintain consistency with the existing S6 pattern. The improvement to use S6ServiceConfig.Equal() and gate to HealthOK only should be addressed in a separate PR focused on S6 FSM improvements to avoid scope creep.

umh-core/pkg/fsm/s6/reconcile.go
```diff
@@ -233,6 +234,11 @@ func (s *S6Instance) reconcileOperationalStates(ctx context.Context, services se
 		metrics.ObserveReconcileTime(metrics.ComponentS6Instance, s.baseFSMInstance.GetID()+".reconcileOperationalStates", time.Since(start))
 	}()
 
+	// Check for unhealthy directory first - this takes precedence over normal state transitions
+	if s.ObservedState.DirectoryHealth == s6service.HealthBad {
+		return s.handleUnhealthyDirectory(ctx, services)
+	}
+
```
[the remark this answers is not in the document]
JeremyTheocharis: Confirmed - this was addressed in commits cd6abf7 to 5d8b96e. The early return issue was fixed by moving the directory health check to UpdateObservedStateOfInstance, which runs before reconcileOperationalStates. This ensures the stopping workaround can still execute properly.

umh-core/internal/fsm/baseFSM.go
FreiPaul: I meant the FSM rule 9 from fsm.mdc

umh-core/pkg/fsm/benthos/actions.go
```diff
@@ -58,27 +61,75 @@ func (b *BenthosInstance) logS6DirectoryState(ctx context.Context, trigger strin
...
+	var lastLogs []s6service.LogEntry
+	logs := b.ObservedState.ServiceInfo.BenthosStatus.BenthosLogs
+	if len(logs) > 0 {
+		start := len(logs) - 3
+		if start < 0 {
+			start = 0
+		}
+		// Make a copy of just the entries we need
+		lastLogs = make([]s6service.LogEntry, len(logs[start:]))
+		copy(lastLogs, logs[start:])
```
FreiPaul: nit: maybe more readable like that (with import "cmp"):
```
n := min(3, len(logs))
lastLogs := append([]s6service.LogEntry(nil), logs[len(logs)-n:]...)
```

umh-core/pkg/fsm/benthos/actions.go
```diff
@@ -119,8 +170,7 @@ func (b *BenthosInstance) logS6DirectoryStateAsync(trigger string, s6State strin
 		}
 	}
 
-	// Gather FSM state context
-	currentFSMState := b.baseFSMInstance.GetCurrentFSMState()
+	// Note: currentFSMState is now a parameter, already captured
```
FreiPaul: Nit: those notes look good in the diff view here when reviewing but might not be needed in the final code. The info that those are snapshotted parmeters is already in the function docstring

umh-core/internal/fsm/baseFSM.go
JeremyTheocharis: in this case it is fine as it is a "emergency" recovery there

umh-core/pkg/fsm/benthos/actions.go
```diff
@@ -50,13 +50,22 @@ import (
...
 func (b *BenthosInstance) logS6DirectoryState(ctx context.Context, trigger string) {
 	// Only log when S6FSMState is empty or shows problems
 	s6State := b.ObservedState.ServiceInfo.S6FSMState
 	if s6State != "" && s6State != benthos_service.S6StateNotExisting {
 		return
 	}
 
+	// Spawn goroutine for async diagnostic logging and Sentry reporting
+	// This prevents blocking the reconciliation loop with I/O operations
+	go b.logS6DirectoryStateAsync(trigger, s6State)
```
FreiPaul: Nit: Might be worth, noting that this routine is short-lived and terminates quickly. Because we have may other long-running goroutines with loops, I think about goroutines as long-running in this repo

umh-core/pkg/fsm/benthos/actions.go
```diff
@@ -50,13 +50,22 @@ import (
...
+}
+
+// logS6DirectoryStateAsync performs the actual diagnostic logging in a goroutine.
+// Uses context.Background() since this is fire-and-forget async logging.
```
FreiPaul: Nit: I understand, why this is fire-and-forget but is there anything against passing it a context of some seconds just to avoid leaks? This context could also be inherited from the context of logS6DirectoryState:
`logCtx, cancel := context.WithTimeout(ctx, 5*time.Second)`

umh-core/pkg/fsm/benthos/actions.go
```diff
@@ -45,6 +51,250 @@ import (
...
+	healthIsReady := b.ObservedState.ServiceInfo.BenthosStatus.HealthCheck.IsReady
+	metricsState := b.ObservedState.ServiceInfo.BenthosStatus.BenthosMetrics.MetricsState
+	
+	// Copy only the last 3 log entries (not the entire log array)
+	var lastLogs []s6service.LogEntry
+	logs := b.ObservedState.ServiceInfo.BenthosStatus.BenthosLogs
+	if len(logs) > 0 {
+		n := min(3, len(logs))
+		// Make a copy of just the last n entries
+		lastLogs = append([]s6service.LogEntry(nil), logs[len(logs)-n:]...)
```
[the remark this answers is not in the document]
JeremyTheocharis: Starting with Go 1.21, there’s a built-in min function that works with any ordered type (integers, floats, strings, etc.):

PR #2273, opened 2025-09-23
fix: handle empty S6 FSM state during Benthos stopping (ENG-3522)

JeremyTheocharis:
## Summary
- Fixes the "stopping: not existing" stuck state bug that occurs when S6 FSM is removed during rollback
- Treats empty S6 FSM state as "stopped" instead of "not stopped"
- Adds comprehensive unit test coverage

## Problem
When a protocol converter deployment fails and rolls back, the S6 FSM can be removed from the manager while the Benthos FSM is still checking its state. This race condition caused Benthos to receive an empty string, which was incorrectly interpreted as "not stopped", leading to an infinite loop with the state stuck at "stopping: not existing".

## Solution
Modified `IsBenthosS6Stopped()` in `pkg/fsm/benthos/actions.go:579-590` to treat empty FSM state as "stopped" since an empty state means the S6 FSM doesn't exist in the manager (effectively stopped).

### Unit Test
Added test case "should handle empty S6 FSM state during stopping (ENG-3522)" that verifies the fix.

### Integration Test - Full Reproduction and Verification
**Test Time**: 2025-09-23 15:25 - 15:32 UTC

#### Test Scenario 1: Failed S7 Deployment (15:25-15:30)
- Deployed protocol converter with S7 protocol to non-existent PLC (localhost:102)
- Deployment failed and rolled back as expected
- **Result**: NO stuck state occurred

Evidence from logs:
```log
2025-09-23 15:28:56.911553588  [DEBUG]  [dataflow-read-protocolconverter-jeremy-test-5]  
  S6 Directory State Debug [trigger=IsBenthosS6Stopped_empty_state]: 
  FSMState='', currentFSM=stopped
```

Stuck state check:
```bash
grep "jeremy-test-5.*stopping.*not existing" /data/logs/umh-core/current
# Result: (empty - NO stuck state)
```

#### Test Scenario 2: Successful Generate Deployment (15:31-15:32)
- Changed protocol to "generate" and deployed successfully
- Service started and became active
- **Result**: Deployment successful, showing 0.07 msg/sec throughput

Final state confirmation:
```log
2025-09-23 15:31:51.125100044  [INFO]  └─ ✅ jeremy-test-5: active → active
2025-09-23 15:32:01.125999298  [INFO]  └─ ✅ jeremy-test-5: active → active
```

### Key Evidence
1. **Fix activated**: `trigger=IsBenthosS6Stopped_empty_state` logged throughout testing
2. **Proper state transitions**: "not existing" states properly transition without getting stuck
3. **No stuck states**: Multiple searches for "stopping.*not existing" returned empty
4. **Successful recovery**: All failed deployments rolled back cleanly

## Linear Issue
[ENG-3522](https://linear.app/united-manufacturing-hub/issue/ENG-3522)

## Test Plan
- [x] Unit tests pass
- [x] Integration test reproduces bug scenario
- [x] Fix prevents stuck state
- [x] Successful deployments still work
- [x] Failed deployments roll back cleanly

### DIFF
--- a/umh-core/pkg/config/manager.go
+++ b/umh-core/pkg/config/manager.go
@@ -1107,6 +1107,7 @@ func (m *FileConfigManager) WriteYAMLConfigFromString(ctx context.Context, confi
 		m.cacheRawConfig = ""
 		m.cacheError = err
 		m.cacheMu.Unlock()
+
 		return fmt.Errorf("failed to parse new config for cache update: %w", err)
 	}
 
--- a/umh-core/pkg/fsm/benthos/actions.go
+++ b/umh-core/pkg/fsm/benthos/actions.go
@@ -580,8 +580,8 @@ func (b *BenthosInstance) IsBenthosS6Stopped() (bool, string) {
 	case "":
 		// Log diagnostic info when we encounter empty S6 state
 		b.logS6DirectoryState("IsBenthosS6Stopped_empty_state")
-
-		fsmState = benthos_service.S6StateNotExisting
+		// Empty state means S6 FSM doesn't exist in manager - treat as stopped (ENG-3522)
+		return true, "service not existing (effectively stopped)"
 	case s6fsm.OperationalStateStopped:
 		return true, ""
 	}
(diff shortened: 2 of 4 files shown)


led0nk (approved):
looks good, but i'd like to test it under real conditions also (with s7)

JeremyTheocharis:
I already tested it under real conditions :)

PR #2274, opened 2025-09-23
[ENG-3057] Bridge -  inject downsampler

led0nk:
#### Description:
This PR should make sure that the `downsampler` is added by default to the bridges processor-configuration.
It must ensure that the downsampler is always the last processor and every bridge consists in its pipeline of it.

This should be done via the runtime_config package, where we resolve the templating and transform the different configuration specs.

### DIFF
--- a/umh-core/pkg/communicator/actions/get-protocolconverter.go
+++ b/umh-core/pkg/communicator/actions/get-protocolconverter.go
@@ -138,30 +138,7 @@ func determineProcessingMode(readDFC *models.ProtocolConverterDFC) string {
 		return "" // reply with empty string to indicate that no DFC is present
 	}
 
-	processors := readDFC.Pipeline.Processors
-
-	// If more than one processor, return custom
-	if len(processors) > 1 {
-		return "custom"
-	}
-
-	// If exactly one processor, check its type
-	if len(processors) == 1 {
-		// Get the first (and only) processor from the map
-		for _, processor := range processors {
-			switch processor.Type {
-			case "nodered_js":
-				return "nodered_js"
-			case "tag_processor":
-				return "tag_processor"
-			default:
-				return "custom"
-			}
-		}
-	}
-
-	// No processors found, fall back to custom
-	return "custom"
+	return constants.DetermineProcessorType(readDFC.Pipeline.Processors)
 }
 
 // determineProtocol analyzes the input processors to determine the protocol.
--- a/umh-core/pkg/communicator/actions/get-protocolconverter.go
+++ b/umh-core/pkg/communicator/actions/get-protocolconverter.go
@@ -132,13 +132,36 @@ func (a *GetProtocolConverterAction) Validate() error {
 
 // determineProcessingMode analyzes the pipeline processors in readDFC only
 // to determine the appropriate processing mode based on the business rules.
+// NOTE:
+// This is used as a wrapper that converts the communicator format to runtime format,
+// calls the core DetermineDataType function, and maps results to be used for the UI.
 func determineProcessingMode(readDFC *models.ProtocolConverterDFC) string {
 	// Only look at readDFC as requested
 	if readDFC == nil {
 		return "" // reply with empty string to indicate that no DFC is present
 	}
 
-	return constants.DetermineProcessorType(readDFC.Pipeline.Processors)
+	// convert communicator format to runtime format
+	var runtimeProcessors []any
+	for _, processor := range readDFC.Pipeline.Processors {
+		procMap := map[string]any{
+			processor.Type: map[string]any{},
+		}
+		runtimeProcessors = append(runtimeProcessors, procMap)
+	}
+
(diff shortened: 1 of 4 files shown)


umh-core/pkg/service/protocolconverter/runtime_config/runtime_config.go
```diff
@@ -304,3 +309,61 @@ func renderConfig(
...
+	// update the config
+	updatedDFC := dfc
+	if updatedDFC.BenthosConfig.Pipeline == nil {
+		updatedDFC.BenthosConfig.Pipeline = make(map[string]any)
+	}
+
+	updatedDFC.BenthosConfig.Pipeline["processors"] = processors
+
+	return updatedDFC
+}
```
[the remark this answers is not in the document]
led0nk: This shouldn't be an issue here, since the downsampler only gets appended whenever we use timeseries data, so there will only be existing 1 processor, which is by now `tag_processor`.  If there is already an existing downsampler it's either already appended after the existing one or a custom-configuration, where we don't want to add the downsampler. On top of that, if it already exists, we're not appending anything here.

umh-core/pkg/communicator/actions/get-protocolconverter.go
```diff
@@ -138,30 +138,7 @@ func determineProcessingMode(readDFC *models.ProtocolConverterDFC) string {
...
-				return "tag_processor"
-			default:
-				return "custom"
-			}
-		}
-	}
-
-	// No processors found, fall back to custom
-	return "custom"
+	return constants.DetermineProcessorType(readDFC.Pipeline.Processors)
```
JeremyTheocharis: I think this is the wrong abstraction type. What you actually want to show in the UI here is the data type. And the data type can only be relational, time series, or custom advanced, whatever. Now, time series should be, and I have 2 subgroups, which are not exposed to these. The current one is where it's just the tag_processor, but then there's also the type that we're now introducing, which means first the tag_processor and then the downsampler.

umh-core/pkg/service/protocolconverter/runtime_config/runtime_config.go
```diff
@@ -293,6 +295,9 @@ func renderConfig(
 		return protocolconverterserviceconfig.ProtocolConverterServiceConfigRuntime{}, err
 	}
 
+	// the downsampler will be appended only for timeseries data
```
JeremyTheocharis: I think the code here should be: if data type equals time series, then append downsampler

and not put it in the `append downsampler` function, but instead put it here and then `append downsampler` can just always append downsampler. This makes it a little bit more intuitive

umh-core/pkg/service/protocolconverter/runtime_config/runtime_config.go
```diff
@@ -304,3 +309,61 @@ func renderConfig(
...
+// appendDownsampler appends a downsampler as the last processor if one doesn't already exist.
+func appendDownsampler(dfc dataflowcomponentserviceconfig.DataflowComponentServiceConfig) dataflowcomponentserviceconfig.DataflowComponentServiceConfig {
+	processors := getProcessors(dfc.BenthosConfig.Pipeline)
+
+	processorType := constants.DetermineProcessorType(processors)
+	if processorType != constants.TimeseriesProcessor {
+		return dfc
+	}
+
+	if hasDownsampler(processors) {
```
JeremyTheocharis: I don't understand why this function is required, because as far as I understand, we have 3 data types: time series, relational, and custom. 

For custom, where the user can add his own stuff, we will never go into this because Benthos sampler only works if there is a down-sampler there. I think we will not see it from the config, actually. 

So we usually said there are 2 versions of it: with the time series data with down-sampler and without down-sampler. 

Actually, because we will never store it in the config YAML because it is done live here in the runtime_config, it means there's only 1 case of time series, which is without down-sampler. 

Then this logic here can be simplified, the whole append-down-sampler. If it is time series, because then we know it has only 1 processor in there, which is tag_processor. In this case, we can always append the down-sampler. If it's custom, then don't do it. If it's relational, don't do it.

JeremyTheocharis (changes requested):

umh-core/pkg/communicator/actions/get-protocolconverter.go
JeremyTheocharis: so no determineprocessortype --> determineDataType

led0nk: it is now with the abstraction type of datatypes and for the action, we wrapped it and set a note that this is only used as a wrapper for UI purpose

umh-core/pkg/service/protocolconverter/runtime_config/runtime_config.go
led0nk: this was just to make it more failsfae but not needed at all and adjusted towards the approach we want to take

JeremyTheocharis (approved):

PR #2334, opened 2025-11-19
fix: resolve data race in MockFSMManager concurrent access (ENG-3924)

JeremyTheocharis:
## Summary

- Fixed data race in `MockFSMManager` where concurrent writes to `ReconcileDelay` and `ReconcileError` fields could race with reads in `Reconcile()`
- Added thread-safe setter methods (`SetReconcileDelay`, `SetReconcileError`) that acquire mutex before writing
- Updated `WithReconcileDelay` and `WithReconcileError` builder methods to also acquire mutex
- Updated all test usages to use the new thread-safe setters
- Added dedicated race condition tests to verify thread-safety

## Root Cause

The `MockFSMManager` struct had a mutex protecting reads in `Reconcile()`, but the `WithReconcileDelay` and `WithReconcileError` builder methods wrote directly to fields without acquiring the mutex. When tests modified these fields concurrently with `Reconcile()` calls, Go's race detector correctly identified the data race.

## Implementation (TDD Workflow)

1. **RED**: Added failing test that demonstrates the race condition (`mock_race_test.go`)
2. **GREEN**: Added thread-safe setter methods with mutex protection (`mock.go`)
3. **REFACTOR**: Updated `loop_test.go` to use the new setters instead of direct field access

## Test plan

- [x] Race detection tests pass with `-race` flag
- [x] ControlLoop fuzz tests pass (originally failing test suite)
- [x] Full unit test suite passes
- [x] golangci-lint passes with 0 issues

### DIFF
--- a/umh-core/pkg/fsm/mock_race_test.go
+++ b/umh-core/pkg/fsm/mock_race_test.go
@@ -18,136 +18,139 @@ import (
 	"context"
 	"errors"
 	"sync"
-	"testing"
 	"time"
+
+	. "github.com/onsi/ginkgo/v2"
 )
 
-// TestMockFSMManager_RaceCondition verifies that concurrent writes to
-// ReconcileDelay/ReconcileError and reads in Reconcile() are thread-safe.
-//
-// This test should PASS with the race detector enabled (-race flag) because:
-// - Reconcile() acquires mutex and reads ReconcileDelay and ReconcileError
-// - SetReconcileDelay/SetReconcileError and WithReconcileDelay/WithReconcileError
-//   all acquire the mutex before writing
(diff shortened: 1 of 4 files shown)


umh-core/pkg/fsm/mock_race_test.go
```diff
@@ -0,0 +1,156 @@
+// Copyright 2025 UMH Systems GmbH
```
led0nk, line 1: would've loved to have this in the test-directory itself, since it would match the structure of other fsm-tests even though it's the manager, but i'm also fine with that

led0nk (approved):

PR #2344, opened 2025-12-01
fix: propagate DebugLevel through config comparison chain

JeremyTheocharis:
### Problem

Changing `debug_level` in config.yaml did not always trigger bridge restarts. Specifically:
- ✅ Changing `debug_level: false` → `true` worked
- ❌ **Adding** `debug_level: true` when field was not present did NOT trigger restart
- ❌ **Removing** `debug_level` field did NOT trigger restart

### Root Cause

The config comparison chain for "observed" (deployed) config was broken:

```
benthos.yaml → BenthosServiceConfig         ✅ DebugLevel read from file
            → DataflowComponentServiceConfig ❌ DebugLevel NOT copied (Bug #1)
            → ProtocolConverterRuntimeConfig ❌ DebugLevel NOT copied (Bug #2)
```

Because `FromBenthosServiceConfig()` did not copy `DebugLevel`, the observed config always had `DebugLevel=false` (Go zero value). This meant:
- Adding `debug_level: true` → compared `true` vs `false` → should detect change, but...
- The comparison happened at the wrong layer, so the S6 config diff check could not see it

### Fix

Two one-line additions to complete the propagation chain:

**1. `pkg/config/dataflowcomponentserviceconfig/utils.go:60`**
```go
DebugLevel: benthos.DebugLevel,
```

**2. `pkg/config/protocolconverterserviceconfig/config_customized.go:88`**
```go
DebugLevel: dfcRead.DebugLevel,
```

### Testing

**Unit Tests**: Added/updated tests verifying DebugLevel propagation through both conversion functions

**Manual Verification** on test server:
```
# Added debug_level: true to config.yaml (field was not present before)
18:21:27 - Configuration differences detected ✅
18:21:27 - Entering stopping state
18:21:28 - Entering starting state
Result: Benthos logger changed from INFO to DEBUG ✅
```

### Checklist

- [x] Unit tests pass
- [x] Manual verification on test server
- [x] CI checks pass (go vet, nilaway, build)
- [x] No focused tests

Fixes ENG-3960

### DIFF
--- a/umh-core/examples/example-config-protocolconverter.yaml
+++ b/umh-core/examples/example-config-protocolconverter.yaml
@@ -29,7 +29,7 @@ protocolConverter:
       config:
         connection:
           nmap:
-            target:   "{{ .HOST }}"
+            target:   "{{ .IP }}"
             port:     "{{ .PORT }}"    # Now templatable! Resolves to proper uint16 at runtime
         dataflowcomponent_read:
           benthos:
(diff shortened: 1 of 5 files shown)


led0nk (commented):
maybe it's just me, but does it make sense to add this somewhere documentation-wise or at least in an example file? Since i had to look it up again, where we are able to actually set this flag.

Tested back and forth on PC & DFC level, which restarts correctly on each change of the flag.

JeremyTheocharis:
Screenshot 2025-12-04 at 08 20 25

 it is documented right?

led0nk (approved):

PR #2357, opened 2025-12-18
ci: add ARM64 binary architecture verification

JeremyTheocharis:
## Summary

Add CI verification to prevent future ARM64 image regressions like the TARGETARCH issue (ENG-4058, #2356).

## Changes

Add `verify-arm64-binaries` job to `build-umh-core.yml` that:

- **Runs on native ARM64**: Uses `depot-ubuntu-24.04-arm` (no QEMU emulation)
- **Retry logic**: 5 retries with exponential backoff for image pull race conditions
- **Binary verification**: Checks all critical binaries are ARM64:
  - `/usr/local/bin/benthos`
  - `/opt/redpanda/bin/redpanda`
  - `/usr/local/bin/umh-core`
  - `/command/s6-svscan`
  - `/command/s6-supervise`
  - `/command/execlineb`
- **Smoke tests**: Verifies binaries actually execute (catches CGO/linking issues)
- **Job summary**: Adds architecture check results to GitHub Actions summary

## Cost

~$0.008 per run (1-2 minutes on ARM runner) - runs on every build

## Stacked PR

This PR is stacked on top of #2356 (ENG-4058 - TARGETARCH fix). Once #2356 is merged, this PR's base should be updated to `staging`.

## Test Plan

- [ ] Verify workflow syntax is valid
- [ ] After base PR merges, CI should run and verify ARM64 binaries
- [ ] Check job summary shows architecture verification results

ENG-4060

### DIFF
--- a/.github/workflows/build-umh-core.yml
+++ b/.github/workflows/build-umh-core.yml
@@ -97,14 +97,31 @@ jobs:
             VERSION=${{ steps.version.outputs.VERSION }} \
             TAGS="$TAGS"
 
-  verify-arm64-binaries:
-    name: Verify ARM64 Image
+  verify-binaries:
+    name: Verify ${{ matrix.arch }} Image
     needs: build
-    runs-on: depot-ubuntu-24.04-arm
+    runs-on: ${{ matrix.runner }}
     timeout-minutes: 10
     permissions:
       contents: read
       packages: read
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: [arm64, amd64]
+        include:
+          - arch: arm64
+            runner: depot-ubuntu-24.04-arm
+            platform: linux/arm64
+            expected_machine: "b700"
+            expected_uname: "aarch64"
+            arch_name: "ARM aarch64"
+          - arch: amd64
+            runner: depot-ubuntu-24.04
+            platform: linux/amd64
+            expected_machine: "3e00"
+            expected_uname: "x86_64"
+            arch_name: "x86-64"
     steps:
       - name: Login to GitHub Container Registry
         uses: docker/login-action@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/build-umh-core.yml
```diff
@@ -97,6 +97,154 @@ jobs:
...
+        run: |
+          ARCH=$(uname -m)
+          echo "Runner architecture: $ARCH"
+          echo "Expected: ${{ matrix.expected_uname }}"
+          if [ "$ARCH" != "${{ matrix.expected_uname }}" ]; then
+            echo "::error::Not running on expected architecture! Expected: ${{ matrix.expected_uname }}, Got: $ARCH"
+            exit 1
+          fi
+
+      - name: Pull ${{ matrix.arch }} image with retry
```
FalconTube, line 143: I would suggest to use an existing GHA (https://github.com/marketplace/actions/retry-step?version=v3.0.2) for the retry logic. Makes it more readable and is probably better tested. 

```yaml
uses: nick-fields/retry@ce71cc2ab81d554ebbe88c79ab5975992d79ba08 # v3
with:
  timeout_seconds: 10
  max_attempts: 5
  command: |
    SHORT_SHA...
    IMAGE...
    docker pull ...
```

.github/workflows/build-umh-core.yml
```diff
@@ -97,6 +97,154 @@ jobs:
...
+        run: |
+          set -euo pipefail
+          echo "Running smoke tests..."
+
+          # Test benthos
+          docker run --rm --platform ${{ matrix.platform }} "$IMAGE" benthos --version
+          echo "benthos executes"
+
+          # Test umh-core
+          docker run --rm --platform ${{ matrix.platform }} "$IMAGE" umh-core --help || true
```
FalconTube: These tests always succeed, even if there is a `panic` when executing the go binary.
It will also succeed, if a wrong typo is in there e.g. `umh-XXX-core` instead of `umh-core`.
Therefore, I do not think these tests make much sense.
We should be fine by checking the correct architecture, as we do above.

FalconTube (changes requested):
Makes a lot of sense to have this now in CI, nice!

In addition to my comments in the code I would suggest to move the entire `verify-binaries` job into a separate `verify-binaries-in-dockerfiles.yml` workflow, make it callable via:

```yaml
on:
    workflow_call:
```

and then call it from the `build-umh-core.yml`. This separates concerns, also allows for standalone execution and should make it more maintainable :)

led0nk (approved):
tested it locally, looks okay to me. i think it might be a bit bloated but it works properly

FalconTube (approved):
Good to merge now

PR #2366, opened 2026-01-08
fix(ci): Reduce depot runner to 4x for integration tests

FalconTube:
We only need the `4x` depot runner.
Saves a lot of build minutes.


JeremyTheocharis (approved):

led0nk (approved):
strong yes!

PR #2372, opened 2026-01-20
[ENG-4139] DevEx: pipe all umh-core logs to container logs

FreiPaul:
Before, the logs were only written to the frequently rotated log file. This file used to be well accessible when we mounted a data dir to the docker container. Now that we switched over to using Docker Volumes, the log files are not anymore easily accessible on every system. This PR provides better log access by piping the UMH-Core logs to the container output. When the user runs `docker logs umh-core`, the umh-core logs show up. Before this PR, the container logs only contained s6 logs on startup and were not useful for debugging.

### DIFF
--- a/umh-core/docs/usage/instances/access-logs.md
+++ b/umh-core/docs/usage/instances/access-logs.md
@@ -1,0 +1,98 @@
+# Accessing Logs
+
+UMH Core provides multiple ways to access logs depending on your deployment setup.
+
+## Container Logs (Recommended)
+
+The simplest way to access logs is via Docker's built-in logging:
+
+```bash
+# View recent logs
+docker logs umh-core
+
+# Follow logs in real-time
+docker logs -f umh-core
+
+# Show last 100 lines and follow
+docker logs -f --tail 100 umh-core
+```
+
+This works regardless of how your `/data` volume is mounted.
+
+## Log Files
+
+UMH Core also writes logs to files inside the container at `/data/logs/umh-core/`. This is useful when you need:
+- Log persistence across container restarts
+- Access to rotated historical logs
+- Integration with external log collectors
+
+### With Bind Mounts
+
+If you mount `/data` to a host directory:
+
+```bash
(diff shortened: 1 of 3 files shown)


daniel-umh (approved):

umh-core/docs/usage/instances/access-logs.md
```diff
@@ -0,0 +1,98 @@
+# Accessing Logs
+
+UMH Core provides multiple ways to access logs depending on your deployment setup.
+
+## Container Logs (Recommended)
```
JeremyTheocharis, line 5: this is not the recommended way. the recommended way is through the UI and in the future through the CLI. The following things here will confuse a lot of people btw. 

i currently fear that we are building an escape hatch, because something in the current logs is not helpful. maybe we should spend time on improving this then instead? 

i can already see people then piping the logs from docker logs into some log file again. or looking at it, and requesting the bridge logs.

umh-core/docs/usage/instances/access-logs.md
```diff
@@ -0,0 +1,98 @@
...
+# Enter the container
+docker exec -it umh-core /bin/sh
+
+# Follow logs
+tail -F /data/logs/umh-core/current
+```
+
+## Log File Structure
+
+Logs are managed by S6's `s6-log` with automatic rotation:
```
JeremyTheocharis, line 64: isnt this already described in the container layout? https://docs.umh.app/reference/container-layout

umh-core/docs/usage/instances/access-logs.md
```diff
@@ -0,0 +1,98 @@
+# Accessing Logs
+
+UMH Core provides multiple ways to access logs depending on your deployment setup.
```
JeremyTheocharis, line 3: wouldnt it be maybe enough to provide a docker run command that goesi nto the container and then tails the log file? instead of exposign it externally? if it is just for internal use? for developers or LLMs?

PR #2373, opened 2026-01-21
feat: gatekeeper package as communicator interface to umh-core

led0nk:
## Description:

Fixes ENG-4098 and ENG-4086 

more information coming soon...


led0nk:
## Remaining Work

  #### main.go Wiring
  - Creation order — Gatekeeper needs SubHandler (subscriber handler), but subscriber handler is created after gatekeeper. Need to reorder or use a setter.
  - Channel type mismatch — Subscriber handler uses chan<- *transport.UMHMessage, gatekeeper's legacy outbound is chan *models.UMHMessage. Need conversion or unification.
  - Replace channelAdapter — LegacyChannelBridge refs in fsmv2Deps and subscriber handler init need replacing. Only in FSMv2 path.
  - Message flow verification — End-to-end trace: PullWorker -> gatekeeper -> Router -> actionHandler/subscriberHandler -> legacyOutbound -> PushWorker

  #### Follow-up Issues
  - ENG-4558: Refactor actions package to write MessageWithSender instead of pre-encoded UMHMessage (~30+ files)

umh-core/pkg/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,332 @@
...
+	if err != nil {
+		g.logger.Warnw("Failed to decode message", "email", msg.Email, "error", err)
+		return
+	}
+
+	// 4. Validate permissions
+	cert := g.certHandler.Certificate(msg.Email)
+	if cert != nil {
+		rootCA := g.certHandler.RootCA()
+		intermediates := g.certHandler.IntermediateCerts(msg.Email)
```
JeremyTheocharis: **Klaus flagged (HIGH)** When `certHandler.Certificate(email)` returns nil, the message passes to `verifiedInbound` with zero permission checks. This is the default state for up to 60 seconds after startup (`time.NewTicker` doesn't fire immediately).

At minimum add a warn log and a TODO. Better: call the fetch function once immediately in `fetchCertificates` before entering the ticker loop.

_12 agents, DA confirmed_

umh-core/pkg/gatekeeper/encryption/interface.go
```diff
@@ -0,0 +1,22 @@
+// Copyright 2025 UMH Systems GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
```
JeremyTheocharis, line 8: **Klaus flagged (HIGH)** `Encrypt([]byte)/Decrypt([]byte)` is stateless. ECDH needs key material for both sides. With the current signature, CseV1 will require a breaking interface change, which defeats the purpose of this PR establishing the interfaces upfront.

I checked `cryptolib` in MC. The actual ECDH signature there is `EncryptPrivateKeyWithPublicKey(ourPrivKey, theirPubKey, keyToEncrypt)` using raw ed25519 keys (NOT x509 certs). To get the key from a cert, cryptolib uses `certificate.GetPublicKeyFromCertificate(cert)` separately.

Two options:
1. Align the interface with cryptolib's pattern now, pass key material: `Decrypt(data []byte, senderPubKey ed25519.PublicKey) ([]byte, error)`. V0 ignores it.
2. Keep encryption entirely in MC cryptolib (closed source) and import it via build flags, same pattern as the validator. The gatekeeper just gets `encryption.Handler` with a noop for OSS and a cryptolib-backed impl for internal builds. No crypto code in the OSS repo at all.

Option 2 might be cleaner since it matches what the validator already does, and keeps all crypto in one place.

_2 agents + MC codebase check. Upgraded from nit after reviewing the PR's stated goal._

JeremyTheocharis (commented):
Hey Janik, looked through the architecture. Design is good, the pipeline structure and interface separation make sense for where CseV1 is going.

One thing that jumped out: the 4 goroutines in `Start()` manage their own lifecycle with WaitGroup + context + manual channel closing. A few race conditions and shutdown issues came up (see inline comments). I'd think about whether FSMv2 could manage these goroutines instead, it already handles panic recovery, shutdown ordering, and the ticker cold-start. Maybe not this PR, maybe a follow-up, up to you.

The other big one: when no certificate is cached for a sender, `handleInbound` skips all permission validation and passes the message through as trusted. `time.NewTicker` doesn't fire immediately, so every restart has a ~60s window where everything bypasses validation. Is this intentional for V0? If so, needs a comment and a TODO ticket.

I also checked the MC codebase:
- `/v2/instance/user/certificate` does case-sensitive email lookup, no normalization. A case mismatch returns 404.
- Revoked user = 204 (cert fields nulled). Current gatekeeper returns `(nil, nil)` on 204 but doesn't clear the cached cert, so a revoked user stays validated forever.
- `cryptolib`'s actual ECDH signature is `EncryptPrivateKeyWithPublicKey(ourPrivKey, theirPubKey, keyToEncrypt)` with raw ed25519 keys. The `encryption.Handler` interface should align with this, or better yet, keep all crypto in MC cryptolib and import via build flags (same pattern as the validator). See inline comment on encryption/interface.go.

Klaus ran a multi-agent review on the diff. Haven't checked each finding myself, treat the inline comments as flags to consider, not mandates.

umh-core/pkg/gatekeeper/validator/validator_noop.go
```diff
@@ -0,0 +1,50 @@
...
+
+//go:build !cryptolib
+
+package validator
+
+import (
+	"crypto/x509"
+
+	"go.uber.org/zap"
+)
```
JeremyTheocharis, line 24: **Klaus flagged (MEDIUM)** Godoc says "returns an error" but code returns `("", nil)`. Downstream calls `parseCertificate("")` which fails on base64 decode with a confusing unrelated error.

Also: `cryptolib` in MC is a package directory, not a Go build tag. This noop validator concept is new. The comment and error behavior should be clear about what "noop" actually means.

Consider: `return "", fmt.Errorf("DecryptRootCA requires cryptolib package")` and a startup warn log.

_8 agents, DA confirmed_

umh-core/pkg/fsmv2/workers/communicator/transport/transport.go
```diff
@@ -16,13 +16,22 @@ package transport
 
 import "context"
 
+// ProtocolVersion indicates the message protocol version.
+type ProtocolVersion string
+
+const (
+	V0    ProtocolVersion = ""       // legacy, no encryption
+	CseV1 ProtocolVersion = "cse_v1" // client-side encryption v1
```
JeremyTheocharis, line 28: **Klaus flagged (MEDIUM)** `type ProtocolVersion string` defined identically in both this file and `models/action_models.go`. These are distinct Go types, code crossing the boundary needs explicit casts. When CseV1 lands, constants must be updated in two places.

Define once in `transport`, remove from `models` or type alias it.

_11 agents, DA demoted to HIGH, I think it's worth fixing now since it's a 5-minute change_

umh-core/pkg/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,332 @@
...
+	go g.processInbound(ctx)
+	go g.processOutbound(ctx)
+	go g.processLegacyOutbound(ctx)
+	go g.fetchCertificates(ctx)
+}
+
+// Stop gracefully shuts down the Gatekeeper.
+func (g *Gatekeeper) Stop() {
+	g.mu.Lock()
+	if !g.running {
```
JeremyTheocharis, line 126: **Klaus flagged (MEDIUM)** `Stop()` closes `verifiedInbound` but not `verifiedOutbound` or `legacyOutbound`. `actions.go:316` uses bare blocking sends (`outboundChannel <- &umhMessage`, no select on ctx.Done). Once buffer fills, action goroutines block permanently.

Probably moot if you move goroutine lifecycle to FSMv2 (see main comment).

_6 agents, DA challenged as speculative, confirmed after checking actions.go_

umh-core/pkg/gatekeeper/certificatehandler/handler.go
```diff
@@ -0,0 +1,278 @@
...
+	userCerts map[string]*certificateBundle
+
+	log             *zap.SugaredLogger
+	encryptedRootCA string
+	jwt             string
+	apiURL          string
+	authToken       string
+	instanceUUID    string
+	mu              sync.RWMutex
+
```
JeremyTheocharis, line 64: **Klaus flagged (MEDIUM)** Two things here:

1. Cert stored under `resp.UserEmail` (from API) but looked up via `msg.Email` (from transport). I checked MC: the endpoint does case-sensitive SQL lookup with no normalization. A case mismatch returns 404. So the cache key will match as long as both sides use the same casing, but there's no guarantee of that across the system.

2. On HTTP 204 (revoked user, cert fields null), the code returns `(nil, nil)` but doesn't clear the cached cert. A previously-validated user whose cert gets revoked stays in cache indefinitely.

_1-3 agents, enriched with MC codebase check_

umh-core/pkg/gatekeeper/gatekeeper.go
```diff
@@ -0,0 +1,332 @@
+// Copyright 2025 UMH Systems GmbH
```
JeremyTheocharis, line 1: **nit (batch):**
- `sort.Strings(emails)` in `fetchCertificates` does nothing, remove
- `betteralign` not run on new structs (`CertHandler`, `Gatekeeper`), run `make betteralign-fix`
- CseV1 TODOs missing ticket refs, use `// TODO(ENG-XXXX):` pattern
- Import ordering in `communication_state.go`, run `goimports`

PR #2375, opened 2026-01-23
chore: implement renovate config

led0nk:
Fixes ENG-4180

first implementation of renovate.json as configuration for renovate-bot


umh-core/Dockerfile
```diff
@@ -13,9 +13,9 @@
 # limitations under the License.
 
 # --- Build Stage ---
-ARG GOLANG_VERSION=MUST_BE_SET_BY_MAKEFILE
-ARG ALPINE_VERSION=MUST_BE_SET_BY_MAKEFILE
-ARG NMAP_VERSION=MUST_BE_SET_BY_MAKEFILE
+ARG GOLANG_VERSION=1.25.3
+ARG ALPINE_VERSION=3.22.2
+ARG NMAP_VERSION=7.97-r0
```
FalconTube: Set the `NMAP_VERSION` back to Makefile, because we do not plan to auto-update via Renovate

FalconTube (approved):
Fix one comment, then good to go

renovate.json
```diff
@@ -0,0 +1,19 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
```
JeremyTheocharis, line 2: there was an HN article that explains how to configure renovate to be less suspectible to supply chai nattacks, using some cooldown period or somethjing for new versions

led0nk: this is already integrated to only take deps after 1 week

PR #2384, opened 2026-01-28
fix(fsmv2): complete race condition audit and fixes

JeremyTheocharis:
## Summary

This PR addresses the `panic: send on closed channel` error (ENG-4217) and performs a comprehensive race condition audit of the FSMv2 supervisor package.

### HIGH Severity

1. **Send on closed channel** (`action_executor.go`)
   - Added `stopped` flag checked before enqueueing actions
   - Ensures channel is not written to after `Shutdown()` closes it

2. **Unsynchronized timeout map read** (`action_executor.go:266`)
   - Moved timeout map read inside the lock

3. **Unsynchronized callback read/write** (`action_executor.go`)
   - `SetOnActionComplete()` now uses mutex
   - `executeWorkWithRecovery()` captures callback under lock before use

4. **Unsynchronized workerCtx.collector access** (`api.go:315-323`)
   - Callback closure now captures collector under `workerCtx.mu.RLock()`

### MEDIUM Severity

5. **worker() reads ae.ctx without lock** (`action_executor.go`)
   - Worker now captures context under lock at function start

6. **executeWorkWithRecovery receives context unsafely** (`action_executor.go`)
   - Context now passed as parameter from worker (which captures it under lock)

7. **globalVars map reference after lock release** (`reconciliation.go:651`)
   - Changed from reference copy to deep copy of map

8. **WaitGroup race between Start() and Shutdown()** (`action_executor.go`)
   - Moved `wg.Add()` and `metricsWg.Add()` operations inside the lock
   - Ensures Add() completes before Shutdown() can proceed to Wait()

## Verification

- All tests pass with race detector enabled
- Extended testing with `ginkgo --race --repeat=5` (6 iterations) - all 215 specs pass consistently
- No race conditions detected after fixes

## Test plan

- [x] Run `make unit-test` - passes with race detector
- [x] Run FSMv2 supervisor tests with race flag - passes
- [x] Extended race testing (6 iterations) - all pass

---

### DIFF
--- a/umh-core/pkg/fsmv2/supervisor/internal/execution/action_executor.go
+++ b/umh-core/pkg/fsmv2/supervisor/internal/execution/action_executor.go
@@ -46,6 +46,7 @@ type ActionExecutor struct {
 	defaultTimeout   time.Duration
 	mu               sync.RWMutex
 	closeOnce        sync.Once
+	stopped          bool
 }
 
 type actionWork struct {
@@ -90,7 +91,9 @@ func NewActionExecutorWithTimeout(workerCount int, timeouts map[string]time.Dura
 }
 
 func (ae *ActionExecutor) Start(ctx context.Context) {
+	ae.mu.Lock()
 	ae.ctx, ae.cancel = context.WithCancel(ctx)
+	ae.mu.Unlock()
 
 	for range ae.workerCount {
 		ae.wg.Add(1)
@@ -99,7 +102,11 @@ func (ae *ActionExecutor) Start(ctx context.Context) {
 	}
 
 	metricsCtx, metricsCancel := context.WithCancel(ctx)
+
+	ae.mu.Lock()
 	ae.metricsCancel = metricsCancel
+	ae.mu.Unlock()
+
 	ae.metricsWg.Add(1)
 
 	go ae.metricsReporter(metricsCtx)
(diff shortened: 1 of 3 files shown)


umh-core/pkg/fsmv2/supervisor/internal/execution/action_executor.go
```diff
@@ -90,44 +91,58 @@ func NewActionExecutorWithTimeout(workerCount int, timeouts map[string]time.Dura
 }
 
 func (ae *ActionExecutor) Start(ctx context.Context) {
```
FalconTube, line 93: I would opt for a single `ae.mu.Lock()` with a single `defer ae.mu.Unlock()` for the entire function.
Makes it a) easier and less verbose and b) ensures the lock is always released.
I don't see any benefit to the current granularity of the multiple locks.

umh-core/pkg/fsmv2/supervisor/internal/execution/action_executor.go
```diff
@@ -311,13 +341,23 @@ func (ae *ActionExecutor) GetActiveActionCount() int {
 
 // SetOnActionComplete sets a callback invoked after each action execution.
 // Should be called during executor setup, before Start().
+// Thread-safe: can be called concurrently with action execution.
 func (ae *ActionExecutor) SetOnActionComplete(fn func(deps.ActionResult)) {
+	ae.mu.Lock()
 	ae.onActionComplete = fn
+	ae.mu.Unlock()
```
FalconTube, line 348: Prefer `defer` over manual unlock
```suggestion
func (ae *ActionExecutor) SetOnActionComplete(fn func(deps.ActionResult)) {
	ae.mu.Lock()
    defer ae.mu.Unlock()
    ae.onActionComplete = fn

```

FalconTube (commented):
I would oft for `defer Unlock()` in two places, but is not strictly needed.
Let me know, if you want to implement it.
If not, I can give approval with the code as is.

umh-core/pkg/fsmv2/supervisor/internal/execution/action_executor.go
JeremyTheocharis: unsure, because this can cause easy issues down the road if one person adds a go routine or miodifies something in for example ae.worker() that then requires the mutex (which remains locked because of the defer)

umh-core/pkg/fsmv2/supervisor/internal/execution/action_executor.go
JeremyTheocharis: here we could do it with the defer, but unsure in the longer functions

FalconTube (approved):
Can be merged as is without `defer` 👍

PR #2390, opened 2026-01-30
fix(fsmv2): fix supervisor shutdown deadlock causing test timeout

JeremyTheocharis:
## Summary

Fixes a **deadlock** in `Supervisor.Shutdown()` that caused the test `Communicator Scenario > Result structure > closes Done channel when complete` to hang indefinitely.

### Root Cause

The `Shutdown()` method had a circular dependency deadlock:
- Phase 1 waited for child's done channels (`<-done`)
- Done channels only close when `tickLoop` exits
- `tickLoop` only exits when context is cancelled (`<-ctx.Done()`)
- Context was cancelled in Phase 3, **AFTER** waiting in Phase 1
- **Result: Deadlock** - Phase 3 never reached

### Fix

Reorder shutdown phases to cancel context **FIRST**:
1. **Phase 1**: Cancel context (allows tickLoops to exit via `ctx.Done()`)
2. **Phase 2**: Wait for children (now unblocked since context is cancelled)
3. **Phase 3**: Request worker graceful shutdown
4. **Phase 4**: Cleanup executors and collectors

### Additional Changes

- Add configurable `MetricsReportInterval` to allow tests to use shorter intervals
- Increase test timeout from 5s to 15s to account for cascading graceful shutdown timeouts

## Test plan

- [x] Run specific failing test multiple times: `go test -v ./pkg/fsmv2/examples/... -ginkgo.focus="closes Done channel when complete"`
- [x] Run full examples test suite: `go test -v ./pkg/fsmv2/examples/...` (30/30 pass)
- [x] Run supervisor tests: `go test -v ./pkg/fsmv2/supervisor/...` (all pass)
- [x] Run with race detector: `go test -race ./pkg/fsmv2/supervisor/...` (no races)

## Linear Issue

Fixes [ENG-4234](https://linear.app/united-manufacturing-hub/issue/ENG-4234)

### DIFF
--- a/umh-core/pkg/fsmv2/examples/communicator_scenario_test.go
+++ b/umh-core/pkg/fsmv2/examples/communicator_scenario_test.go
@@ -207,10 +207,7 @@ var _ = Describe("Communicator Scenario", func() {
 			result := examples.RunCommunicatorScenario(ctx, examples.CommunicatorRunConfig{
 				Duration: 500 * time.Millisecond,
 			})
-			// Allow extra time for graceful shutdown.
-			// Shutdown may take up to 5s (GracefulShutdownTimeout) per supervisor level.
-			// With parent + child supervisors = up to 10s+ for shutdown.
-			// 15s timeout provides safety margin.
+			// Allow extra time for graceful shutdown (duration + cascading child shutdown timeouts)
 			Eventually(result.Done, 15*time.Second).Should(BeClosed())
 		})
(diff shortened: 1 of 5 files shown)


led0nk (approved):
lg from my perspective but might makes sense that @FalconTube also ahve a look ehere

FalconTube (approved):
looks good to me as well

PR #2400, opened 2026-02-04
fix(fsmv2): initialize lastLifecyclePhase and implement single-tick architecture

JeremyTheocharis:
## Summary

- **Fixes flaky test** `scenarios_test.go:591` ("child count mismatch") caused by `lastLifecyclePhase` being read before initialization
- **Implements single-tick architecture** - children no longer have their own tick loops, eliminating double-ticking and unnecessary goroutines

### Issue 1: Use Before Initialization
`lastLifecyclePhase` starts as zero value (`PhaseUnknown`) and isn't set until the first tick. But the collector runs async and can read it immediately after `AddWorker()` via `ChildrenCountsProvider`. If read before first tick → wrong phase → incorrect child health counts.

### Issue 2: Double-Ticking
Children had their own tick loops (spawned by `Start()`), meaning they were ticked both by their own goroutine AND by the parent's synchronous `child.tick()` call. This created unnecessary goroutines and potential timing issues.

## Changes

| File | Change |
|------|--------|
| `supervisor/api.go` | Initialize `lastLifecyclePhase` from `initialState.LifecyclePhase()` in `AddWorker()` |
| `supervisor/lifecycle.go` | Add `StartAsChild()` method (starts collectors/executors but no tick loop) |
| `supervisor/types.go` | Add `StartAsChild()` to `SupervisorInterface` |
| `supervisor/reconciliation.go` | Use `StartAsChild()` instead of `Start()` for child supervisors |
| `integration/scenarios_test.go` | Update test to check for both `tick_loop_started` and `supervisor_started_as_child` |

## Single-Tick Architecture

```
Root Supervisor
├─ Tick Loop (only goroutine doing ticks)
│   └─ tick()
│       └─ tickWorker()
│       └─ reconcileChildren()
│       └─ for each child:
│           └─ child.tick()  ← SYNCHRONOUS
```

**Before**: Children had their own tick loops → double-ticking, unnecessary goroutines
**After**: Only root has tick loop → deterministic, no concurrent tick loops

## Test Plan

- [x] All 208 supervisor tests pass
- [x] All 18 integration tests pass
- [x] Ran flaky test 5 times consecutively - all passed
- [x] `go vet` passes
- [x] `golangci-lint` passes (no new issues)

Fixes: ENG-4267
Parent: ENG-4231 (Rollout of FSMv2)


umh-core/pkg/fsmv2/supervisor/lifecycle.go
```diff
@@ -73,6 +73,42 @@ func (s *Supervisor[TObserved, TDesired]) Start(ctx context.Context) <-chan stru
...
+func (s *Supervisor[TObserved, TDesired]) StartAsChild(ctx context.Context) {
+	s.ctxMu.Lock()
+	s.ctx, s.ctxCancel = context.WithCancel(ctx)
+	s.ctxMu.Unlock()
+	s.started.Store(true)
+
+	s.logger.Debugw("supervisor_started_as_child")
+
+	s.ctxMu.RLock()
+	supervisorCtx := s.ctx
```
led0nk: we could put this to the first lock to simplify, because it doesn't get modified afterwards

JeremyTheocharis: unsure, because we are accessing another function "Store" there, and this could lead to potential deadlocks in the future (as people dont expect it)

JeremyTheocharis: todo: move thisc ctx to above lock of crtx

led0nk (approved):
Good catch for second method

PR #2402, opened 2026-02-04
[PR 1 of ENG-4292] ENG-4293: Add memory leak characterization test for FSMv2 persistence

JeremyTheocharis:
## Summary

Adds a characterization test that demonstrates the memory leak in FSMv2's in-memory persistence layer. This test passes by proving the leak exists and will serve as a regression test after cleanup is implemented.

### What the test demonstrates:

1. **Delta Entry Accumulation**: After 50 worker create/remove cycles with 5 ticks each, 150 delta entries remain (should be 0)

2. **Worker Document Accumulation**: After removing 20 workers, 60 documents remain (20 identity + 20 desired + 20 observed)

### Test Labels

- `memory-leak`: Tests that demonstrate the current leak behavior
- `memory-cleanup`: Tests that will pass after ENG-4295 implements cleanup (currently skipped)

## Context

Part of the FSMv2 Memory Cleanup stack:
- **ENG-4293** (this PR): Add characterization test ✅
- ENG-4295: Implement storage layer cleanup methods
- ENG-4296: Implement PersistenceWorker
- ENG-4294: Implement FSMv2-native metrics
- ENG-4297: Integration testing and validation

**Root Issue**: ENG-4270 - Sonnländer reports 13GB RAM usage with only 4 msg/sec throughput

## Test plan

- [x] Run memory leak tests: `go test ./pkg/fsmv2/supervisor/... -ginkgo.label-filter="memory-leak" -v`
- [x] Verify no focused tests: `go test ./pkg/fsmv2/supervisor/... --ginkgo.fail-on-pending=false`
- [ ] CI passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2403, opened 2026-02-04
[PR 2 of ENG-4292] ENG-4295: Implement Storage Layer Cleanup Methods

JeremyTheocharis:
## Summary

Implements `CompactDeltas()` and `Maintenance()` methods on TriangularStore to prevent unbounded delta growth in FSMv2's in-memory persistence layer.

### Changes

- **`CompactDeltas(retentionWindow)`** - Deletes deltas older than the specified retention window. Clients who fall behind get a full bootstrap instead of incremental deltas.
- **`Maintenance()`** - Clears the snapshot cache (preparation for future SQLite VACUUM operations).
- **Clock injection** - Added `NewTriangularStoreWithClock()` constructor and propagated clock throughout the storage layer for deterministic testing.
- **Tests enabled** - Removed `Skip()` calls from memory leak characterization tests; all 6 tests now pass.

### Key Design Decisions

1. **Compaction = DELETE, not aggregate**: The triangular collections already store full state. `buildBootstrap()` handles sync clients who fall behind.
2. **Millisecond timestamp precision**: Deltas store timestamps as `UnixMilli()`. Tests must advance clock by at least 1ms for compaction to work.
3. **Non-blocking deletion**: Uses individual `Delete()` calls to avoid holding locks during bulk operations.

### Test Results

```
Ran 90 of 90 Specs in 0.015 seconds
SUCCESS! -- 90 Passed | 0 Failed | 0 Pending | 0 Skipped
```

## Test plan

- [x] All CSE storage tests pass (90/90)
- [x] All FSMv2 tests pass
- [x] `go vet` clean
- [x] `golangci-lint` clean on modified files
- [x] CI/CD passes

## Related

- Part of ENG-4270: FSMv2 Memory Leak
- Stacked on: PR #2402 (ENG-4293)


umh-core/pkg/cse/storage/interfaces.go
```diff
@@ -109,6 +110,12 @@ type TriangularStoreInterface interface {
 
 	// GetDeltas returns incremental deltas or full bootstrap data if client is too far behind.
 	GetDeltas(ctx context.Context, sub Subscription) (DeltasResponse, error)
+
+	// CompactDeltas removes delta entries older than the retention window.
+	CompactDeltas(ctx context.Context, retentionWindow time.Duration) (int, error)
```
JeremyTheocharis, line 116: NOTE: is this only deleting deltas, or also old snapshots?

umh-core/pkg/cse/storage/triangular.go
```diff
@@ -1309,17 +1332,107 @@ func (ts *TriangularStore) GetLatestSyncID(ctx context.Context) (int64, error) {
...
+					"doc", doc,
+					"reason", "missing or invalid id field")
+
+				skipped++
+
+				continue
+			}
+
+			if err := ts.store.Delete(ctx, DeltaCollectionName, id); err != nil {
+				ts.logger.Warnw("failed_to_delete_delta",
```
led0nk: why are these warnings, whenever the `Find` above is an error? Also i'm thinking of this being not able to delete the delta -> being skipped and then although it's logged how much was skipped/failed one doesn't really pay attention to this

umh-core/pkg/cse/storage/triangular.go
```diff
@@ -1309,17 +1332,107 @@ func (ts *TriangularStore) GetLatestSyncID(ctx context.Context) (int64, error) {
...
+	if deleted > 0 || skipped > 0 || failed > 0 {
+		ts.logger.Infow("compaction_complete",
+			"deleted", deleted,
+			"skipped", skipped,
+			"failed", failed,
+			"retention_window", retentionWindow)
+	}
+
+	if failed > 0 {
+		return deleted, fmt.Errorf("compaction completed with %d failures (deleted %d, skipped %d)", failed, deleted, skipped)
```
led0nk: Also here i get the point, first logging something as an "overview" but then we again check on this and return with an error > which will also lead to a logging. It would be better if something failed to do an early return there, otherwise print out the ones that are skipped/deleted and the retentionwindow and thats it

umh-core/pkg/cse/storage/triangular.go
```diff
@@ -1309,17 +1332,107 @@ func (ts *TriangularStore) GetLatestSyncID(ctx context.Context) (int64, error) {
...
+		}
+
+		return 0, fmt.Errorf("failed to query deltas for compaction: %w", err)
+	}
+
+	deleted := 0
+	skipped := 0
+	failed := 0
+
+	for _, doc := range docs {
```
led0nk, line 1383: what happens if the context gets cancelled here? i guess the loop will still run, do we want this behaviour?

umh-core/pkg/cse/storage/triangular.go
```diff
@@ -1309,17 +1332,107 @@ func (ts *TriangularStore) GetLatestSyncID(ctx context.Context) (int64, error) {
...
+		ts.logger.Warnw("compaction_skipped",
+			"reason", "store is nil")
+
+		return 0, nil
+	}
+
+	cutoffTime := ts.clock.Now().Add(-retentionWindow)
+	cutoffMs := cutoffTime.UnixMilli()
+
+	docs, err := ts.store.Find(ctx, DeltaCollectionName, persistence.Query{})
```
led0nk, line 1365: will this then load all deltas into memory?

led0nk (changes requested):

JeremyTheocharis:
Addressed all of @led0nk's feedback in commit 9474af1:

**Memory concern (loading all deltas):** Yes, but for the in-memory backend this is fine — the data is already in memory. Added a TODO for when we switch to SQLite (that's when a filtered query like `WHERE timestamp < ?` will actually matter). Once compaction runs every 5 min with 10-min retention, the delta collection stays bounded anyway.

**Context cancellation:** Added `select { case <-ctx.Done() }` at the top of the loop. Follows the pattern from `collector.go`, `action_executor.go`, and all `Execute()` methods.

**Warn noise + double-logging:** Switched to fail-fast error handling. For the in-memory backend, delete failures are essentially impossible:
- `ErrNotFound` requires a concurrent deleter — but `RemoveWorker()` doesn't clean deltas, and PersistenceWorker is single-threaded
- Malformed timestamps require data corruption — but `DeltaStore.Append()` always sets the field
- Context nil is never passed

If the impossible happens, we want to know immediately — so the function now returns an error on the first anomaly. No counters, no partial-success tracking, no summary-then-error double-logging.

For SQLite: the error handling strategy should be revisited when that backend is implemented (I/O errors become real).

---

**Re Jeremy's question on `interfaces.go:115`:** `CompactDeltas` only handles deltas. `Maintenance()` handles snapshot cache clearing. Compaction is frequent/lightweight (every 5 min), maintenance is infrequent/heavier (weekly). Snapshots are overwritten in-place (not accumulated), so they don't need time-based cleanup.

JeremyTheocharis:
**Error handling for malformed deltas:**

Switching to skip malformed docs (missing timestamp/id) instead of failing the whole compaction. One bad delta shouldn't block cleanup of everything else.

**The change:**
- Count skipped docs, log once at the end with a sentinel error (`ErrMalformedDelta`)
- Sentry gets alerted and can fingerprint properly
- No per-doc spam, operators see the aggregate count
- Delete errors still fail (monitoring will catch it)

Fixes the log spam issue you flagged.

```go
// Before:
if !hasTimestamp {
    return deleted, fmt.Errorf("delta missing timestamp: doc_id=%v", doc["id"])
}

// After:
if !hasTimestamp {
    skipped++
    continue
}
// at end:
if skipped > 0 {
    err := fmt.Errorf("%w: skipped %d malformed documents", ErrMalformedDelta, skipped)
    ts.logger.Warnw("compaction_skipped_documents", "error", err, "skipped", skipped, ...)
}
```

umh-core/pkg/cse/storage/triangular.go
```diff
@@ -1307,19 +1334,129 @@ func (ts *TriangularStore) GetLatestSyncID(ctx context.Context) (int64, error) {
...
 //   - error: Any error that occurred during compaction
-//
-// TODO(ENG-4295): Implement actual compaction logic using timestamp-based deletion.
 func (ts *TriangularStore) CompactDeltas(ctx context.Context, retentionWindow time.Duration) (int, error) {
-	return 0, errors.New("CompactDeltas not implemented - see ENG-4295")
+	if ts.store == nil {
+		return 0, errors.New("store is nil, cannot compact deltas")
+	}
+
+	cutoffTime := ts.clock.Now().Add(-retentionWindow)
```
led0nk, line 1358: maybe just use `Sub()` ?

umh-core/pkg/cse/storage/triangular.go
```diff
@@ -1307,19 +1334,129 @@ func (ts *TriangularStore) GetLatestSyncID(ctx context.Context) (int64, error) {
...
+
+// extractDocKeys returns the keys present in a document for diagnostic logging.
+// Helps troubleshoot malformed documents by showing what fields ARE present.
+func extractDocKeys(doc persistence.Document) []string {
+	keys := make([]string, 0, len(doc))
+	for k := range doc {
+		keys = append(keys, k)
+	}
+
+	return keys
```
led0nk, line 1462: not sure if this has impact, but slice sorting might make sense

led0nk (approved):
looks good, the 2 things are more on readability-side of things

umh-core/pkg/cse/storage/triangular.go
JeremyTheocharis: `time.Time.Sub()` takes a `time.Time` argument (not a `time.Duration`), so it returns the duration between two points in time. To subtract a duration from a time, `Add(-duration)` is the standard Go idiom — there is no `Sub(duration)` on `time.Time`.

umh-core/pkg/cse/storage/triangular.go
JeremyTheocharis: Good catch — added `sort.Strings(keys)` in 0f7988f08.

umh-core/pkg/cse/storage/triangular.go
```diff
@@ -1307,19 +1335,130 @@ func (ts *TriangularStore) GetLatestSyncID(ctx context.Context) (int64, error) {
...
+}
+
+// extractDocKeys returns the keys present in a document for diagnostic logging.
+// Helps troubleshoot malformed documents by showing what fields ARE present.
+func extractDocKeys(doc persistence.Document) []string {
+	keys := make([]string, 0, len(doc))
+	for k := range doc {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
led0nk: just uses `slices.Sort` directly as this already does

led0nk (changes requested):

umh-core/pkg/cse/storage/triangular.go
JeremyTheocharis: Switched to `slices.Sort` and dropped the `sort` import.

led0nk (approved):

PR #2412, opened 2026-02-05
[PR 6 of ENG-4292] ENG-4320: Add startup/shutdown maintenance states

JeremyTheocharis:
## Summary

- Add `TryingToStartState` and `ShuttingDownState` to run maintenance explicitly during startup and graceful shutdown, preparing for SQLite migration where `Maintenance()` = `VACUUM`
- Introduce shared `deps.Scheduler` interface for nighttime/weekend preference scheduling with three-tier maintenance logic (preferred/acceptable/overdue)
- Add scheduling booleans (`IsPreferredMaintenanceWindow`, `IsAcceptableMaintenanceWindow`) to observed state, computed via injected Scheduler dependency

State machine change: `Stopped → TryingToStart → Running ↔ RunningDegraded → ShuttingDown → Stopped`

## Cross-Stack Impact

The `deps.Scheduler` interface introduced here will be reused by the TransportWorker (PR chain #2399) to replace inline scheduling logic for JWT proactive reauth. Follow-up ticket: ENG-4322.

## Test plan

- [x] 14 Scheduler unit tests (boundaries, timezone independence)
- [x] 5 TryingToStartState specs (shutdown, success, retry, priority)
- [x] 4 ShuttingDownState specs (success, retry, multi-tick)
- [x] 11 preferential scheduling specs (short/long intervals, windows, overdue, loop prevention)
- [x] Updated existing state tests for new transitions
- [x] 60 total persistence worker tests passing with `-race`
- [x] 45 example/scenario integration tests passing with `-race`
- [x] `go vet` clean, no focused tests


umh-core/pkg/fsmv2/deps/scheduler.go
```diff
@@ -0,0 +1,75 @@
...
+// Methods are timezone-agnostic — they operate on the timezone of the
+// provided time.Time. Containers typically run in UTC, so time.Now()
+// produces UTC. The transport worker uses time.Now().Local() explicitly.
+// Both approaches work — the caller decides.
+//
+// Maintenance window hierarchy (subset relationship):
+//   - IsPreferredMaintenanceWindow ⊂ IsAcceptableMaintenanceWindow
+//   - Preferred = weekend AND low-traffic (best time for heavy ops like VACUUM)
+//   - Acceptable = weekend OR low-traffic (good-enough fallback after target interval)
+type Scheduler interface {
```
led0nk, line 27: The Scheduler's design is intentional here — methods take `time.Time` as a parameter rather than calling `time.Now()` internally. This makes them pure functions: the caller decides what time to pass (`time.Now()`, `time.Now().Local()`, or a fixed time in tests). This achieves the same testability as a clock interface but with simpler API surface. The comment at scheduler.go:19-22 documents this choice explicitly.

So no clock abstraction needed — passing `time.Time` as parameter IS the clock pattern, just expressed differently.

umh-core/pkg/fsmv2/deps/scheduler_test.go
```diff
@@ -0,0 +1,171 @@
...
+	var s deps.DefaultScheduler
+
+	BeforeEach(func() {
+		s = deps.DefaultScheduler{}
+	})
+
+	Describe("IsWeekend", func() {
+		It("returns true for Saturday", func() {
+			sat := time.Date(2025, 2, 1, 12, 0, 0, 0, time.UTC)
+			Expect(sat.Weekday()).To(Equal(time.Saturday))
```
led0nk: this line is just self-confirmation imo

on top of that, those tests would be great to be in table-driven tests of ginkgo

umh-core/pkg/fsmv2/workers/persistence/state/trying_to_start.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsmv2/internal/helpers"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsmv2/workers/persistence/action"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsmv2/workers/persistence/snapshot"
+)
+
+type TryingToStartState struct {
+	helpers.StartingBase
+}
+
+func (s *TryingToStartState) Next(snapAny any) fsmv2.NextResult[any, any] {
```
led0nk, line 28: will this be shutdown whenever we fail multiple times here for a certain amount of time ?

umh-core/pkg/fsmv2/workers/persistence/state/shutting_down.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsmv2/internal/helpers"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsmv2/workers/persistence/action"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsmv2/workers/persistence/snapshot"
+)
+
+type ShuttingDownState struct {
+	helpers.StoppingBase
+}
+
+func (s *ShuttingDownState) Next(snapAny any) fsmv2.NextResult[any, any] {
```
led0nk, line 28: same question for me, what happens if we're never successful, we will run maintancen again and again, right?

umh-core/pkg/fsmv2/deps/scheduler.go
led0nk: btw we only use `IsPreferredMaintenanceWindow` and `IsAcceptableMaintenanceWindow` currently, are those other ones planned to use or what's the purpose?

umh-core/pkg/fsmv2/workers/persistence/state/running_test.go
```diff
@@ -122,7 +122,7 @@ var _ = Describe("RunningState", func() {
 					Observed: snapshot.PersistenceObservedState{
 						CollectedAt:       now,
 						LastCompactionAt:  now.Add(-1 * time.Minute),
-						LastMaintenanceAt: now.Add(-8 * 24 * time.Hour),
+						LastMaintenanceAt: now.Add(-10 * 24 * time.Hour),
```
led0nk, line 125: should always use `now.Sub()` which os more readable imo

led0nk (changes requested):

umh-core/pkg/fsmv2/workers/persistence/state/running_test.go
led0nk: ignore this then

umh-core/pkg/fsmv2/workers/persistence/state/trying_to_start.go
JeremyTheocharis: Good catch — this is indeed unbounded today.

After surveying all 9 FSMv2 workers, we confirmed this is a **universal pattern**: every transitional state (TryingToStart, TryingToStop, TryingToConnect, ShuttingDown) retries indefinitely across the entire codebase. The persistence worker is not unique here.

The good news is the framework already has all the building blocks needed: `FrameworkMetrics.TimeInCurrentStateMs` for time-based bounds, `ConsecutiveActionErrors` for attempt counting, and `SignalNeedsRestart` for escalation to the supervisor (which has a 30s force-reset backstop).

Created **ENG-4365** to add reusable `StartupBudgetExceeded()` / `ShutdownBudgetExceeded()` helper functions that lifecycle states can call. Default budgets: 2 min / 30 attempts for startup, 5 min / 60 attempts for shutdown. Startup is safe to abandon (no resources held), shutdown gets more patience because of data loss risk.

This is scoped as post-rollout work since it's a cross-cutting concern affecting all workers, not just persistence.

umh-core/pkg/fsmv2/workers/persistence/state/shutting_down.go
JeremyTheocharis: Same situation as TryingToStart above — today this loops indefinitely. See the reply on the TryingToStart comment for the full analysis.

The key insight for ShuttingDown specifically: this state holds resources (the worker acquired them during startup), so we can't just give up as easily. The proposed solution in **ENG-4365** gives shutdown a longer budget (5 min / 60 attempts vs 2 min / 30 for startup). When exhausted, it signals `SignalNeedsRestart` and the supervisor's existing 30-second force-reset becomes the hard backstop.

umh-core/pkg/fsmv2/deps/scheduler_test.go
JeremyTheocharis: Deleted test blocks for the removed methods (self-confirming assertions gone). Remaining tests converted to table-driven DescribeTable/Entry format. See c00117c0d.

umh-core/pkg/fsmv2/deps/scheduler.go
JeremyTheocharis: Removed `IsWeekend`, `IsLowTrafficHours`, `IsBusinessHours` from the Scheduler interface — never called outside DefaultScheduler. Kept `IsPreferredMaintenanceWindow` and `IsAcceptableMaintenanceWindow` (used by the 3-tier maintenance scheduling in the persistence worker). See c00117c0d.

led0nk (approved):

umh-core/pkg/fsmv2/examples/persistence_scenario_test.go
```diff
@@ -0,0 +1,209 @@
...
+			Expect(result.Error).NotTo(HaveOccurred())
+			Expect(result.Done).NotTo(BeNil())
+			Expect(result.Shutdown).NotTo(BeNil())
+			<-result.Done
+			Expect(result.CompactionCycles).To(BeNumerically(">=", 0))
+			Expect(result.MaintenanceCycles).To(BeNumerically(">=", 0))
+			Expect(result.Healthy).To(BeTrue())
+			Expect(result.LastCompactionAt).NotTo(BeZero())
+			Expect(result.LastMaintenanceAt).NotTo(BeZero())
+		})
```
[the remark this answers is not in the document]
JeremyTheocharis: These tests have a 30-second context timeout (line 33) that prevents infinite hangs -- the supervisor is forced to shut down if it doesn't complete naturally. The `<-result.Done` calls are synchronization barriers before asserting result fields, not the test assertion itself. The tests that DO assert channel closure (lines 119, 130, 194) already use `Eventually`. This is intentional: `Eventually` where closure is the assertion, bare receive where it's a sync step.

umh-core/pkg/fsmv2/examples/scenario.go
```diff
@@ -249,6 +250,55 @@ var CommunicatorScenarioEntry = Scenario{
...
+//
+//	go run pkg/fsmv2/cmd/runner/main.go --scenario persistence --duration 5s --dump-store
+//
+// # What Gets Tested
+//
+//   - FSMv2 persistence worker state machine (Stopped -> Running)
+//   - CompactDeltas action triggered on first tick (LastCompactionAt=zero fires immediately)
+//   - RunMaintenance action (independent of compaction; compaction takes priority per tick)
+//   - Metrics collection (compaction cycles, maintenance cycles, duration gauges)
+var PersistenceScenarioEntry = Scenario{
```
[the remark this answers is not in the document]
JeremyTheocharis: The `CommunicatorScenarioEntry` at line 193 has the same comment structure and was not flagged. These are standard godoc comments on exported variables documenting CLI usage and what gets tested -- they follow the established pattern in this file. The CLAUDE.md style guide targets generated/auto-added comments, not developer-authored Go doc comments.

umh-core/pkg/fsmv2/workers/persistence/snapshot/snapshot.go
```diff
@@ -58,6 +59,9 @@ type PersistenceObservedState struct {
 	deps.MetricsEmbedder `json:",inline"`
 
 	ConsecutiveActionErrors int `json:"consecutive_action_errors"`
+
+	IsPreferredMaintenanceWindow  bool `json:"is_preferred_maintenance_window"`
+	IsAcceptableMaintenanceWindow bool `json:"is_acceptable_maintenance_window"`
```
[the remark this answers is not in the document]
JeremyTheocharis: Verified -- `make betteralign-fix` produces no changes. The bool fields are already at the end of the struct.

umh-core/pkg/fsmv2/workers/persistence/state/shutting_down.go
```diff
@@ -0,0 +1,43 @@
...
+	snap := helpers.ConvertSnapshot[snapshot.PersistenceObservedState, *snapshot.PersistenceDesiredState](snapAny)
+
+	for _, result := range snap.Observed.LastActionResults {
+		if result.Success {
+			return fsmv2.Result[any, any](&StoppedState{}, fsmv2.SignalNone, nil, "Shutdown maintenance completed")
+		}
+	}
+
+	return fsmv2.Result[any, any](s, fsmv2.SignalNone,
+		action.NewRunMaintenanceAction(), "Running shutdown maintenance")
```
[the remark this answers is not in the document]
JeremyTheocharis: Valid point. Added ActionType filtering in both ShuttingDownState and TryingToStartState -- the state now only accepts results matching the action it emitted.

umh-core/pkg/fsmv2/workers/persistence/state/trying_to_start.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	for _, result := range snap.Observed.LastActionResults {
+		if result.Success {
+			return fsmv2.Result[any, any](&RunningState{}, fsmv2.SignalNone, nil, "Startup maintenance completed")
+		}
+	}
+
+	return fsmv2.Result[any, any](s, fsmv2.SignalNone,
+		action.NewRunMaintenanceAction(), "Running startup maintenance")
```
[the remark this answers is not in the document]
JeremyTheocharis: Same fix applied here. Both states now filter by ActionType.

umh-core/pkg/fsmv2/examples/persistence_scenario.go
```diff
@@ -0,0 +1,154 @@
...
+
+	logger := cfg.Logger
+	if logger == nil {
+		logger = deps.NewNopFSMLogger()
+	}
+
+	tickInterval := cfg.TickInterval
+	if tickInterval == 0 {
+		tickInterval = 100 * time.Millisecond
+	}
```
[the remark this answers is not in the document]
JeremyTheocharis: Added negative TickInterval validation with early return and error, matching the existing Duration validation pattern.

PR #2413, opened 2026-02-05
Make FSMv2 feature flags env-var-only

JeremyTheocharis:
## Summary

- Moves `USE_FSMV2_TRANSPORT` out of the config manager pipeline (no longer persisted to config.yaml)
- Adds `USE_FSMV2_MEMORY_CLEANUP` and `USE_FSMV2_PROTOCOL_CONVERTER` as env-var-only flags
- All three FSMv2 flags are now read directly from `os.Getenv()` in `cmd/main.go` after config loading

This makes the temporary migration flags consistent: runtime toggles passed via `docker -e`, never written to disk, removed when the config manager becomes an FSMv2 worker.

**Breaking change**: Users with `useFSMv2Transport: true` in config.yaml must switch to `-e USE_FSMV2_TRANSPORT=true`.

## Test plan

- [x] `go build ./cmd/...` passes
- [x] `go test ./pkg/config/... -count=1 -race` passes
- [x] CI green
- [ ] Verify env vars work in Docker: `docker run -e USE_FSMV2_TRANSPORT=true -e USE_FSMV2_MEMORY_CLEANUP=true ...`

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>


umh-core/cmd/main.go
```diff
@@ -101,6 +102,19 @@ func main() {
 		return
 	}
 
+	// FSMv2 feature flags: read directly from env vars, not persisted to config.yaml.
+	// These bypass the config manager intentionally — they are temporary migration flags
+	// that will be replaced when the config manager becomes an FSMv2 worker.
+	if v, _ := env.GetAsBool("USE_FSMV2_TRANSPORT", false, false); v {
```
led0nk: we should do errorhandling here imo

umh-core/cmd/main.go
```diff
@@ -101,6 +102,19 @@ func main() {
...
+		configData.Agent.UseFSMv2Transport = true
```
led0nk: also what if its already set in the config as true, this then won't be turned off again?

umh-core/pkg/config/env.go
```diff
@@ -30,15 +30,15 @@ import (
...
+// 1. Environment variables (AUTH_TOKEN, API_URL, RELEASE_CHANNEL, ALLOW_INSECURE_TLS, LOCATION_*)
 // 2. Existing config file values
 // 3. Default values
 //
 // In Docker environments, this enables runtime configuration through environment variables,
 // which is particularly useful for CI/CD pipelines, testing, and containerized deployments.
 // For example, in the Makefile:
 //
-//	docker run -e AUTH_TOKEN=xyz -e LOCATION_0=factory1 -e ALLOW_INSECURE_TLS=true -e USE_FSMV2_TRANSPORT=true $(IMAGE_NAME):$(TAG)
+//	docker run -e AUTH_TOKEN=xyz -e LOCATION_0=factory1 -e ALLOW_INSECURE_TLS=true $(IMAGE_NAME):$(TAG)
```
led0nk, line 41: why is this removed? its in the config package yes, but it needs to be set then as envvar, doesn't it?

led0nk (changes requested):

umh-core/cmd/main.go
JeremyTheocharis: Now reports parse errors via `sentry.ReportIssuef(IssueTypeWarning, ...)` instead of silently discarding them.

umh-core/cmd/main.go
JeremyTheocharis: Changed to unconditional assignment — `configData.Agent.UseFSMv2Transport = v` — so the env var can also disable the flag, not just enable it.

umh-core/pkg/config/env.go
JeremyTheocharis: The flag is still read from env vars, just moved to `cmd/main.go` as a runtime-only override. It's no longer part of the config manager pipeline since these are temporary migration flags.

led0nk: @JeremyTheocharis yeah i got that,sorry i meant why is the comment removed with the envvar here, since it looks like now you don't need it anymore as you still need to set them as envvar via run-command

umh-core/cmd/main.go
```diff
@@ -105,16 +105,27 @@ func main() {
 	// FSMv2 feature flags: read directly from env vars, not persisted to config.yaml.
 	// These bypass the config manager intentionally — they are temporary migration flags
 	// that will be replaced when the config manager becomes an FSMv2 worker.
-	if v, _ := env.GetAsBool("USE_FSMV2_TRANSPORT", false, false); v {
-		configData.Agent.UseFSMv2Transport = true
+	v, err := env.GetAsBool("USE_FSMV2_TRANSPORT", false, false)
+	if err != nil {
+		sentry.ReportIssuef(sentry.IssueTypeWarning, log, "Failed to parse USE_FSMV2_TRANSPORT: %v", err)
```
led0nk: Do you think it makes sense to report a sentry issue for this? 
I'm thinking of what does it provide to us, which would basically just be, that someone used a wrong type for the flag here. Personally i would expect to just have it logged via startup to have feedback (warning) on this.

led0nk (approved):

umh-core/pkg/config/env.go
JeremyTheocharis: ahh

JeremyTheocharis: The comment was updated, not removed — env.go now has a note that USE_FSMV2_* flags are handled in cmd/main.go. They still exist as env vars, just moved out of LoadConfigWithEnvOverrides since they're runtime-only migration flags.

umh-core/cmd/main.go
JeremyTheocharis: Agreed — changed to log.Warnf. A misconfigured env var is an operator error, not something we need Sentry alerts for. Startup log warning is sufficient feedback.

PR #2427, opened 2026-02-10
chore(deps): update Alpine version for openssl

led0nk:
# Description

Fixes ENG-4366

We need to update the current Alpine-Image to a newer version due to a security issue within openssl version on the older Alpine-Image.

Upgrade:
`3.22.2` -> `3.23.3` (from Dockerfile)


FalconTube (approved):
Nice

PR #2438, opened 2026-02-18
feat: add config.yaml backup versioning to prevent data loss

JeremyTheocharis:
Add automatic timestamped backups of config.yaml before every write operation. Backups are stored in /data/config-backups/ with unix timestamp filenames. Keeps last 100 backups to bound storage usage.

Key changes:
- Add createConfigBackup() to create timestamped backup before writes
- Add cleanupOldBackups() to remove old backups beyond 100 limit
- Modify writeConfig() and WriteYAMLConfigFromString() to call backup
- Add constants ConfigBackupDir and ConfigBackupMaxEntries
- Add unit tests for backup and cleanup functionality

Backup is best-effort: failures are logged but don't block main writes. This addresses customer data loss incidents where config.yaml was overwritten with no recovery option.

Slack: https://umh-app.slack.com/archives/C07D3S5ESHM/p1771415761017189?thread_ts=1771398686.818119&cid=C07D3S5ESHM

https://claude.ai/code/session_018sb7inQyVWfZtPhEK4JVsK

Fixes
VE-235
ENG-4398 (sub-issue)
ENG-4399 (sub-issue)

PR #2439, opened 2026-02-18
feat: config backup versioning before writes

JeremyTheocharis:
## Summary

- **Config backup on write**: Before every config.yaml write, creates a timestamped backup in `/data/config-backups/` to prevent data loss during upgrades or crash loops
- **Diff-check**: Compares current config against latest backup before creating a new one — identical content is skipped, preventing backup spam during repeated restarts (crash loop scenario)
- **Automatic cleanup**: Retains the last 100 distinct config versions (~1MB max), removing oldest backups when the limit is exceeded

## Implementation

Three new unexported methods on `FileConfigManager`:
- `createConfigBackup(ctx)` — best-effort backup before writes (never blocks config writes on failure)
- `getLatestBackupContent(ctx)` — reads latest backup for diff-check
- `cleanupOldBackups(ctx)` — enforces the 100-backup retention limit

Wired into both write paths (`writeConfig` and `WriteYAMLConfigFromString`), inside the write lock, before the actual write.

## Test plan

- [x] 18 unit tests covering all backup scenarios (create, skip, diff-check, cleanup, error handling)
- [x] 2 integration tests verifying call-site wiring for both write paths
- [x] `go test ./pkg/config/... -race -count=1` — all 98 specs pass
- [x] `golangci-lint run` — 0 issues
- [x] `go vet ./...` — clean
- [x] No focused specs
- [x] OSS hygiene verified (no customer names, no internal links)

Closes ENG-4398, ENG-4399
Supersedes #2438


umh-core/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## [0.44.9]
 
+### New Features
+
+- **Config backup versioning** - Automatic timestamped backups of config.yaml are created before every write operation, providing recovery options when config is accidentally overwritten. Identical configs are deduplicated to prevent backup churn during restarts. Last 100 backups are retained in /data/config-backups/.
```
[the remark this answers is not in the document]
JeremyTheocharis: Fixed in bb4e761f7. Also added a verification hint for users.

umh-core/cmd/main.go
```diff
@@ -128,6 +128,14 @@ func main() {
 
 	configData.Agent.UseFSMv2ProtocolConverter = v
 
+	// Config backup feature flag: read from env var, not persisted to config.yaml.
+	v, err = env.GetAsBool("ENABLE_CONFIG_BACKUP", false, false)
+	if err != nil {
+		log.Warnf("Failed to parse ENABLE_CONFIG_BACKUP: %v", err)
+	}
+
+	configManager.SetConfigBackupEnabled(v)
```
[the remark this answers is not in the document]
JeremyTheocharis: Good catch — fixed in bb4e761f7. Moved `ENABLE_CONFIG_BACKUP` parsing before `LoadConfigWithEnvOverrides` so the first startup write backs up the pre-existing config.

umh-core/pkg/config/manager.go
```diff
@@ -150,6 +152,7 @@ type FileConfigManager struct {
 	// ---------- background refresh state ----------
 	refreshMu sync.Mutex // prevents concurrent background refreshes
 
+	backupEnabled bool // Feature flag: create config backups before writes
```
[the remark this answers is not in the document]
JeremyTheocharis: Fixed in bb4e761f7. Converted to standalone comment above the field.

umh-core/pkg/config/manager.go
```diff
@@ -1134,3 +1147,131 @@ func (m *FileConfigManagerWithBackoff) WriteYAMLConfigFromString(ctx context.Con
...
+		m.logger.Warnf("config backup: failed to create backup directory: %v", err)
+
+		return
+	}
+
+	filename := time.Now().UTC().Format("20060102T150405.000000000Z") + ".yaml"
+	backupPath := filepath.Join(constants.ConfigBackupDir, filename)
+
+	if err := m.fsService.WriteFile(ctx, backupPath, content, 0666); err != nil {
+		m.logger.Warnf("config backup: failed to write backup file: %v", err)
```
led0nk, line 1190: was just wondering if it makes sense to have sentry in here, that we are able to see if at least this works? (just as an idea, not needed for this quickfix)

umh-core/pkg/constants/config.go
```diff
@@ -28,4 +28,10 @@ const (
 	// It is more a safety net to prevent a single reader from blocking the config file
 	// The actual number does not really matter, it should be "high enough".
 	AmountReadersForConfigFile = 100
+
+	// ConfigBackupDir is the directory where config backups are stored.
+	ConfigBackupDir = "/data/config-backups"
+
+	// ConfigBackupMaxEntries is the maximum number of backup files to retain.
+	ConfigBackupMaxEntries = 100
```
led0nk, line 36: maybe we reduce this to e.g. 50 for now? as this might increase and then can get quite big fast

led0nk (approved):
looks good, works, but maybe just reduce the amount of backups as this might grow and grow

umh-core/pkg/config/manager.go
JeremyTheocharis: also thought about it, but it will probably hide in the current noise of fsmv1 sentyr anyway :/

umh-core/pkg/constants/config.go
JeremyTheocharis: i was unsure how often it would actually triggert, i think it can trigger a couple of times for a single bridge deplioyment, so i wanted to keep the number high as disk space is usually available plenty

umh-core/pkg/config/manager.go
JeremyTheocharis: will create follow up PR for this to use the new fsmv2 sentry package, might still be good insights

PR #2440, opened 2026-02-18
feat: update edit-instance action to use generic location format

ricci2511:
## Summary

Fixes ENG-4336.

- Replaces the old hardcoded location fields in the `edit-instance` action with the generic `map[int]string` location format already used throughout the rest of the config
- Ensures the Management Console can update instance location to any arbitrary hierarchy depth without requiring manual `config.yaml` edits
- Adds location hierarchy validation to prevent gaps in user input (see `validation.go`)
- Updates `AtomicSetLocation` to a simpler function definition

## Bug Fix: Stale Location Level Cleanup

Fixed = bug identified in `AtomicSetLocation` was not properly removing stale location levels from `ProtocolConverter` and `StreamProcessor` configurations when the agent's location hierarchy shrunk.

**Problem**: When an agent location changed from `{0:"ACME", 1:"Factory-1", 2:"Area-A"}` to `{0:"ACME", 1:"Factory-2"}`, the old level 2 ("Area-A") remained in bridge and stream processor configurations, causing inconsistent topic paths.

**Solution**: Implemented a cleanup approach that:
1. Captures the old agent location before applying changes
2. Identifies which specific levels were removed from the agent configuration
3. Selectively removes only those stale agent levels from child component configurations
4. Preserves any bridge/stream processor-specific location levels that don't conflict

**Gap Preservation**: The implementation intentionally preserves gaps in location levels (e.g., `{0:"ACME", 1:"Factory-1", 3:"Line-B"}`) rather than reordering to avoid potential breaking changes to user configurations. Gaps can occur when:
- Users manually edit location hierarchies
- Location restructuring removes intermediate levels

This approach prioritizes configuration stability over level ordering.

**Testing**: See test cases in `edit_instance_test.go`, specifically:
- `"should properly handle location shrinking"` - Basic shrinking scenario
- `"should handle multiple levels being removed simultaneously"` - Complex multi-level removal with bridge-specific level preservation

## Location Validation

Added defensive validation for location hierarchies to catch common input errors:

- **New validation helper**: `config.ValidateLocation()` checks for required level 0, consecutive levels, and non-empty values
- **Integration**: Used in `edit-instance` action's `Validate()` method for early error detection
- **Clear error**: Specific feedback about which level has issues (e.g., "location hierarchy has a gap at level 2")

**Validation scope**: Only validates new user input through Management Console actions. Existing configurations with gaps are preserved for backward compatibility.


JeremyTheocharis (commented):
## Review — PR #2440 (edit-instance location format)

Reviewed this together with MC PR united-manufacturing-hub/ManagementConsole#2939. Overall the direction is right — the generic `map[int]string` format is a good move and the feature flag coordination with MC is clean. Two things to address:

### 1. Bug: Stale location levels in ProtocolConverter/StreamProcessor after shrink

`manager.go:716-726` (and mirrored in `mock.go:371-382`): The loop that propagates location to ProtocolConverters and StreamProcessors is additive-only — it never removes old keys.

**Concrete failure:** User has `{0:"ACME", 1:"Factory-1", 2:"Area-A"}`, edits to `{0:"ACME", 1:"Factory-2"}` (removing level 2). After `AtomicSetLocation`:
- `Agent.Location` → `{0:"ACME", 1:"Factory-2"}` ✓
- `ProtocolConverter[i].Location` → `{"0":"ACME", "1":"Factory-2", "2":"Area-A"}` ✗ stale level 2!

The bridge computes a topic path with the deleted level. Agent and bridges become inconsistent.

**Fix approach:** Selective removal — delete keys matching old agent-level indices from ProtocolConverter/StreamProcessor maps before copying new ones. A full map replacement would risk wiping bridge-specific location keys. Both `manager.go` and `mock.go` need the fix.

Also add a test that verifies ProtocolConverter/StreamProcessor locations after a level-removal operation — current tests only check `Agent.Location`.

### 2. Changelog: reframe for cross-repo coordination

The current entry says "You can now update the instance location directly from the Management Console UI" — but this is only true once MC PR #2939 also ships. Since `changelog.umh.app` requires separate entries per component, I'd suggest reframing the umh-core entry to describe the backend capability:

```
### Improvements

- **Edit-instance action now supports generic location format** - The edit-instance
  action now accepts any number of location hierarchy levels instead of the previous
  five hardcoded fields (enterprise, site, area, line, workCell). This enables the
  Management Console to offer flexible location editing once updated.
```

The MC PR should then carry its own changelog entry describing the user-facing outcome ("you can now update location from the UI") with a note "Requires umh-core v0.44.9 or later."

### Minor items (non-blocking, can be tickets)
- No validation for gaps in location levels (e.g., `{0:"ACME", 2:"Assembly"}` skipping level 1) — MC enforces sequential levels in the UI so low risk, but defensive validation would be nice
- Mock lacks `AtomicSetLocation` call-tracking flag (pattern inconsistency with other Atomic methods)

ricci2511:
@JeremyTheocharis 

Thanks for the review! I've implemented the fix for the stale location level cleanup bug in `AtomicSetLocation`.

### Details

The solution works as follows:

1. **Capture old state**: Before updating the agent location, we capture the existing `config.Agent.Location` 
2. **Identify removed levels**: Compare old vs new to determine which specific agent levels were removed
3. **Selective cleanup**: Remove only the identified stale agent levels from `ProtocolConverter` and `StreamProcessor` configurations
4. **Preserve bridge-specific levels**: Any levels that weren't part of the original agent location remain untouched

### Gap Preservation Decision

The implementation intentionally preserves gaps in location levels rather than reordering them. For example, if a configuration has `{0:"ACME", 1:"Factory-1", 3:"Line-B"}`, we maintain the gap at level 2.

**Rationale**:
- **Avoid unforeseen breaking changes**: Reordering could break existing user configurations or integrations that depend on specific level indices
- **Preserve semantic meaning or user intent**: Users might assign semantic significance to specific level numbers

While gaps are uncommon in typical usage (Management Console UI enforces sequential levels), they can still occur via:
- Manual `config.yaml` edits
- Instance location hierarchy restructuring

### Testing

Added two tests in `edit_instance_test.go`:
- **Basic shrinking**: `"should properly handle location shrinking"` 
- **Complex scenarios**: `"should handle multiple levels being removed simultaneously"` - tests removing multiple agent levels (2,3,4) while preserving a bridge-specific level (5)

Both tests verify that stale agent levels are removed while bridge-specific levels are preserved.

Let me know if this is what you expected or whether we should re-order bridge and stream processor levels to avoid gaps.

ricci2511:
@JeremyTheocharis I've also added proper location validation, see `validation.go`

led0nk:
let's bring this topic up in the product meeting before proceeding here

ricci2511:
@led0nk @JeremyTheocharis As discussed, the ability to edit the instance location brings in more cons than pros.
Should we close this pr or consider removing everything related to instance location edits?

JeremyTheocharis:
## Product Meeting Decisions (2026-02-25)

Same decisions as MC #2939 — lock location after initial setup, auto-lowercase, installation warnings.

For the backend: the `edit-instance` action's location editing capability may become restricted once MC #2939 locks the frontend. The validation and gap-handling code in this PR is still valuable for migration scenarios and edge cases.

Full meeting discussion: [ENG-3762](https://linear.app/united-manufacturing-hub/issue/ENG-3762/)

ricci2511:
Closing as it isn’t relevant anymore.
We won’t provide an action to edit the instance’s location due to the many cascading effects it presents.

PR #2452, opened 2026-02-26
fix(ci): label live-endpoint tests to unblock unit-test CI

JeremyTheocharis:
## Summary

- Add `Label("live")` to `Context("Requester")` in `requester_test.go` — tests hit live `management.umh.app/api`
- Add `Label("live")` to `Describe("Checking API reachability")` in `reachability_test.go` — calls `CheckIfAPIIsReachable()` against live endpoint with 5-minute timeout
- Add `!live` to `unit-test` and `unit-test-fail-slow` Makefile label filters so labeled tests are excluded from CI

CI has been blocked across all branches since Feb 24 because `GET /v2/instance/login` now returns 404. These tests are unlabeled, so the `unit-test` blocklist (`!integration && !redpanda-extended && !tls`) doesn't exclude them.

The Keep-Alive Header tests (using `httptest.NewTLSServer`) and TLS Security tests (already `Label("tls")`) are unaffected.

Ref: ENG-4475 (parent: ENG-4474)

## Test plan

- [x] `ginkgo --dry-run --label-filter='!integration && !redpanda-extended && !tls && !live'` — live tests excluded, Keep-Alive tests still run
- [x] `ginkgo --dry-run --label-filter='live'` — selects only the 3 live specs
- [x] `go vet` passes
- [x] CI `unit-test` job passes (no live endpoint calls)


rpaland (approved):
LGTM

PR #2453, opened 2026-02-26
Parallelize unit test suite execution in CI

JeremyTheocharis:
## Summary

Switches the `unit-test` Makefile target from sequential `ginkgo -r` to parallel `go test ./...` + `gotestsum` (for JUnit XML in CI).

**Result: ~23 min → ~16 min (~30% faster)**

### Changes

- Add `gotestsum` as a dev tool (standard in Kubernetes, Terraform)
- Extract `UNIT_LABEL_FILTER` variable to deduplicate label filter across targets
- Replace `unit-test` with `go test` (packages run in parallel via GOMAXPROCS)
- In CI: use `gotestsum` for JUnit XML output; locally: plain `go test`
- Pass label filter via `-args -ginkgo.label-filter=...` (correct Ginkgo v2 approach for `go test`)
- Exclude `integration/` from unit test packages (needs Docker, has own CI job)
- Keep `unit-test-fail-slow` on ginkgo (sequential, easier to debug) with shared variable

### Key learnings

- `GINKGO_LABEL_FILTER` env var does NOT exist in Ginkgo v2 — must use `-ginkgo.label-filter` flag
- `-ginkgo.label-filter` must be passed via `-args` (after packages) for `go test` to forward it correctly
- The bottleneck is the Examples suite (6.9 min) — wall time can't be less than the longest package

### CI results

| Package | Time | Notes |
|---------|------|-------|
| pkg/fsmv2/examples | 414s | Bottleneck |
| pkg/fsmv2/supervisor | 283s | |
| pkg/fsmv2/config | 275s | |
| pkg/fsmv2/integration | 137s | (fsmv2 unit tests, not top-level) |
| 120 other packages | <33s each | |
| **Total** | **124 packages, 0 failures** | **~16 min wall time** |

## Test plan

- [x] `make -n unit-test` shows correct command (non-CI path)
- [x] `CI=true make -n unit-test` shows gotestsum with JUnit output
- [x] `make -n unit-test-fail-slow` uses shared `UNIT_LABEL_FILTER` variable
- [x] CI unit tests pass (124 packages, 0 failures)
- [x] Label filter works: HTTP package runs 2/36 specs (34 live/tls skipped)
- [x] Top-level `integration/` excluded from test packages
- [x] JUnit XML uploaded as artifact

### DIFF
--- a/umh-core/Makefile
+++ b/umh-core/Makefile
@@ -317,10 +317,12 @@ test: vet-nilaway-lint-betteralign
 unit-test: vet-nilaway-lint-betteralign
 ifeq ($(CI), true)
 	gotestsum --junitfile unit-test-report.xml -- -race -tags=test -count=1 -failfast \
-		-ginkgo.label-filter='$(UNIT_LABEL_FILTER)' ./cmd/... ./internal/... ./pkg/... ./test/...
+		./cmd/... ./internal/... ./pkg/... ./test/... \
+		-args -ginkgo.label-filter='$(UNIT_LABEL_FILTER)'
 else
 	go test -race -tags=test -count=1 -failfast \
-		-ginkgo.label-filter='$(UNIT_LABEL_FILTER)' ./cmd/... ./internal/... ./pkg/... ./test/...
+		./cmd/... ./internal/... ./pkg/... ./test/... \
+		-args -ginkgo.label-filter='$(UNIT_LABEL_FILTER)'
 endif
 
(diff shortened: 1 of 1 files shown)


FalconTube (approved):
- Looks good, but only uses 2 parallel workers (because of 2 CPU on the depot VM) to execute the tasks
- Can probably be improved. A quick test on my side however showed one specific test, that does not like scaling up the parallel workers

tl;dr: Geat for now, but probably room for even further improvement in the future

PR #2460, opened 2026-03-02
ci: reject PRs that modify released CHANGELOG sections

JeremyTheocharis:
## Summary

Three times on PR #2447 alone: CHANGELOG.md merge conflicts during rebase silently moved new entries under the already-released `## [0.44.10]` section. The existing `validate-changelog` job only checks format, not content integrity. This adds an automated guard that catches it.

### What happened

When rebasing onto staging, git's conflict markers land inside CHANGELOG.md. Resolving them by hand (or via AI) can accidentally place new entries below the released version header instead of above it. The TypeScript format validator has no concept of "this section was already shipped."

### How it works

A single new bash step ("tail match") in the `validate-changelog` job:

1. Finds the latest stable tag (`v0.44.10`)
2. Extracts everything from `## [0.44.10]` to EOF in the PR's CHANGELOG ("released tail")
3. Extracts the same tail from `git show v0.44.10:umh-core/CHANGELOG.md`
4. Diffs them — any difference = CI failure with an actionable error message and collapsible diff

This catches edits anywhere in released sections, including older versions like 0.44.9, since the tail includes everything below the latest tagged header.

### Why inline bash (not extending the TS validator)

The check needs `git show TAG:path` which requires repo-local tag access. The TS validator is format-focused and shared via changelog.umh.app. Keeping this as inline bash matches the existing pattern (`update-github-release.yml` already does shell-based CHANGELOG parsing).

### Reuse in other repos

Portable to ManagementConsole and benthos-umh — change `CHANGELOG="umh-core/CHANGELOG.md"` to the respective path. Could be extracted into a reusable composite action later if needed.

### Known limitations

**Merge queue**: The guard is intentionally skipped on `merge_group` events because `github.event.pull_request` is null there (label bypass would silently break). This means the guard runs on every PR push — which is when the human is actively looking at CI — but not on the final merge queue run. A bad rebase in the merge queue itself would not be caught. In practice, the merge queue uses the same commit that already passed the guard on the PR branch, making a new violation at that stage very unlikely.

**Link references at file bottom**: If the changelog format ever adds reference-style links at the bottom (standard keep-a-changelog practice), those would trigger the guard. Add the `skip-changelog-guard` label in that case, or let a maintainer bypass.

**`fetch-depth: 0`**: Full history is fetched on every changelog PR so `git tag --list --sort=-version:refname` and `git show TAG:path` work. A targeted single-tag fetch would be cheaper but more complex to implement correctly. Accepted tradeoff.

**Bypass (maintainers only)**: The `skip-changelog-guard` label bypasses the check. Only repository members with triage access can add labels — external contributors who hit a false positive should ask a maintainer to apply it.

### Changes

One file, two edits to `.github/workflows/pull-request.yml`:

1. `fetch-depth: 0` + `permissions: contents: read` on `validate-changelog`
2. New step "Check released sections are unmodified" between `setup-node` and "Validate CHANGELOG.md"

## Test plan

- [x] Guard passes on unmodified CHANGELOG
- [x] Guard fails with clear diff when bad entry inserted under released section
- [x] No trailing-newline false positive (`printf '%s'` strips trailing newline)
- [x] Scenario A (no tags): `::warning::` + exit 0 ✓
- [x] Scenario C (header not in tag): `::warning::` + exit 0 ✓
- [x] Scenario D (diff exit 2): `::error::` + exit 1 ✓
- [x] Merge-group fix: guard correctly skips on `merge_group`/`workflow_dispatch`
- [x] `|| true` on all grep pipelines prevents pipefail-triggered abort
- [ ] CI runs the new step on this PR (happy path, guard passes trivially)

Closes ENG-4525

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -129,8 +129,9 @@ jobs:
 
           CHANGELOG="umh-core/CHANGELOG.md"
 
-          # Find the latest stable tag (vX.Y.Z, no pre-release suffixes)
-          LATEST_TAG=$(git tag --list 'v[0-9]*.[0-9]*.[0-9]*' --sort=-version:refname | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | head -1)
+          # Find the latest stable tag (vX.Y.Z, no pre-release suffixes).
+          # "|| true" prevents set -e from aborting when grep finds no matches.
+          LATEST_TAG=$(git tag --list 'v[0-9]*.[0-9]*.[0-9]*' --sort=-version:refname | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | head -1 || true)
 
           if [[ -z "$LATEST_TAG" ]]; then
             echo "::warning::No stable tags found — skipping released-section guard"
@@ -141,8 +142,8 @@ jobs:
 
           HEADER="## [${LATEST_TAG#v}]"
 
-          # Find the header line in the current CHANGELOG (fixed-string match)
-          LINE_NUM=$(grep -Fn "$HEADER" "$CHANGELOG" | head -1 | cut -d: -f1)
+          # Find the header line in the current CHANGELOG (fixed-string match; || true for set -e safety)
+          LINE_NUM=$(grep -Fn "$HEADER" "$CHANGELOG" | head -1 | cut -d: -f1 || true)
 
           if [[ -z "$LINE_NUM" ]]; then
             echo "::warning::Header '$HEADER' not found in current $CHANGELOG — skipping guard"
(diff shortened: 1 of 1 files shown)


led0nk (approved):
> 🧹 Nitpick comments (1)
>
>    .github/workflows/pull-request.yml (1)
>
>        129-212: Consider extracting to a standalone script for testability.
>
>        The inline bash is well-commented and correct, but at ~80 lines it's nearing the threshold where a separate script file (e.g., .github/scripts/changelog-guard.sh) would enable:
>
>            Direct unit testing with various changelog fixtures
>            Easier local debugging without workflow dispatch
>            Shellcheck linting in CI

we should think about this as this gets longer and longer with each change we make and an additional shellcheck would help indeed then

PR #2243, opened 2025-09-05
feat: implement resource-based bridge creation blocking (ENG-3421)

JeremyTheocharis:
## Summary
- Implements proactive resource management to prevent system overload by blocking protocol converter bridge creation when resource limits are reached
- Blocks creation when CPU/Memory/Disk degraded or theoretical bridge limit exceeded
- Allows stuck bridges to be removed even when at resource limits
- **Intentionally exposes CPU throttling metrics via public APIs** to provide valuable user insights (deviation from original spec, see rationale below)

## API Surface Additions (Intentional)
While the original Linear issue specified "no new APIs", we've decided to expose throttling information for significant user value:

### New Public APIs:
1. **Interface Method**: `IProtocolConverterService.IsResourceLimited(snapshot) (bool, string)`
2. **CPU Status Fields**: 
   - `CPU.CgroupCores`: Container CPU quota
   - `CPU.ThrottleRatio`: Ratio of time being throttled  
   - `CPU.IsThrottled`: Boolean throttling indicator
3. **Config Flag**: `Agent.EnableResourceLimitBlocking` (feature flag, defaults to false)

### Rationale:
- Users need to understand WHY their system is degraded (throttling is counterintuitive)
- Enables proactive capacity planning before hitting limits
- Essential for troubleshooting performance issues
- Provides actionable insights ("Container limited to 2 cores, needs more during peaks")

## Test Plan
- [x] Unit tests added (16 tests covering all scenarios)
- [x] Manual testing with 11 bridges on 2-core system
- [x] Verified CPU throttling detection works
- [x] Confirmed stuck bridges can be removed
- [ ] Integration test with production workload
- [ ] Performance testing under load

### Resource Limiting Logic
- Added `IsResourceLimited()` to check system health before bridge creation
- Blocks when ANY resource (CPU/Memory/Disk) is degraded
- Enforces theoretical limit: (CPU cores - 1 for Redpanda) × 5 bridges
- Detects CPU throttling via cgroups v2

### Thresholds Lowered
- CPU: 90% → 70% (prevent throttling)
- Memory: 90% → 80% (safety margin)  
- Disk: 90% → 85% (prevent full disk)

### Edge Case Handling
- Bridges stuck in `to_be_created` can transition to removal states
- Known bypass: User can toggle stopped→active to bypass limits (acceptable, requires deliberate action)

### Clear Error Messages
Before: `To_be_created: Protocol converter state: to_be_created: 11/10 per 2.0 allocated cores`
[...]

PR #2463, opened 2026-03-05
fix(mem): cgroup memory reading

led0nk:
Fixes ENG-4471

# Description
This PR fixes memory metrics reporting host-level values instead of container cgroup
limits.

### Previous Issue
`getMemoryMetrics()` used `gopsutil`'s `mem.VirtualMemoryWithContext()`, which reads `/proc/meminfo` — host-level memry. The struct fields `CGroupUsedBytes` / `CGroupTotalBytes` were populated with host values. On a 32 GB VM with an 8 GB k3s pod imit, it showed 31.4 GiB total, causing the 80% memory threshold to trigger against host RAM instead of the pod limt, wrongly blocking bridge creation.

### Solution
Added `cgroup_memory.go` (same pattern as existing for cpu) that reads:
- `/sys/fs/cgroup/memory.max` → cgroup memory limit (`"max"` = unlimited)
- `/sys/fs/cgroup/memory.current` → current container memory usage

`getMemoryMetrics()` now prefers cgroup values when available and falls back to `gopsutil` host values in non-container environments (macOS, bare metal).

### DIFF
--- a/umh-core/pkg/service/container_monitor/cgroup_memory.go
+++ b/umh-core/pkg/service/container_monitor/cgroup_memory.go
@@ -46,6 +46,9 @@ func ParseMemoryMax(data []byte) (limitBytes int64, unlimited bool, err error) {
 	if err != nil {
 		return 0, false, fmt.Errorf("failed to parse memory.max value %q: %w", s, err)
 	}
+	if limitBytes < 0 {
+		return 0, false, fmt.Errorf("negative memory.max value %q", s)
+	}
 
 	return limitBytes, false, nil
 }
@@ -61,6 +64,9 @@ func ParseMemoryCurrent(data []byte) (currentBytes int64, err error) {
 	currentBytes, err = strconv.ParseInt(s, 10, 64)
 	if err != nil {
 		return 0, fmt.Errorf("failed to parse memory.current value %q: %w", s, err)
+	}
+	if currentBytes < 0 {
+		return 0, fmt.Errorf("negative memory.current value %q", s)
 	}
 
 	return currentBytes, nil
--- a/umh-core/pkg/service/container_monitor/cgroup_memory.go
+++ b/umh-core/pkg/service/container_monitor/cgroup_memory.go
@@ -16,6 +16,7 @@ package container_monitor
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"strconv"
@@ -34,7 +35,7 @@ type MemoryCgroupInfo struct {
 func ParseMemoryMax(data []byte) (limitBytes int64, unlimited bool, err error) {
 	s := strings.TrimSpace(string(data))
 	if s == "" {
-		return 0, false, fmt.Errorf("empty memory.max data")
+		return 0, false, errors.New("empty memory.max data")
 	}
 
 	if s == "max" {
@@ -54,7 +55,7 @@ func ParseMemoryMax(data []byte) (limitBytes int64, unlimited bool, err error) {
 func ParseMemoryCurrent(data []byte) (currentBytes int64, err error) {
 	s := strings.TrimSpace(string(data))
 	if s == "" {
-		return 0, fmt.Errorf("empty memory.current data")
+		return 0, errors.New("empty memory.current data")
 	}
 
 	currentBytes, err = strconv.ParseInt(s, 10, 64)
--- a/umh-core/pkg/service/container_monitor/cgroup_memory.go
+++ b/umh-core/pkg/service/container_monitor/cgroup_memory.go
@@ -1,0 +1,98 @@
+// Copyright 2025 UMH Systems GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


umh-core/pkg/service/container_monitor/cgroup_memory.go
```diff
@@ -0,0 +1,105 @@
...
+
+	currentBytes, err := ParseMemoryCurrent(memCurrentData)
+	if err != nil {
+		return nil, err
+	}
+
+	info.CurrentBytes = currentBytes
+
+	return info, nil
+}
```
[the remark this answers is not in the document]
led0nk: will be tackled via eng-4555 same as cpu

JeremyTheocharis (commented):
FYI this closes a 6-month gap where CPU had cgroup-aware reading (PR #2243, ENG-3421) but memory was still on host-level gopsutil since the original container monitoring was added in March 2025.

Also: missing changelog entry. This is user-visible (correct memory metrics in containers), needs an entry under `### Fixes`.

umh-core/pkg/service/container_monitor/container_monitor.go
```diff
@@ -349,6 +349,15 @@ func (c *ContainerMonitorService) getMemoryMetrics(ctx context.Context) (*models
...
 	totalBytes := vmStat.Total
 
+	// Try cgroup values — prefer container-aware limits over host values
+	cgroupInfo, cgroupErr := c.getCgroupMemoryInfo(ctx)
+	if cgroupErr == nil {
+		usedBytes = uint64(cgroupInfo.CurrentBytes)
+		if !cgroupInfo.Unlimited && cgroupInfo.LimitBytes > 0 {
+			totalBytes = uint64(cgroupInfo.LimitBytes)
+		}
+	}
```
JeremyTheocharis, line 363: When `getCgroupMemoryInfo` fails, this silently falls back to host values with no log. On a Linux container where cgroup should work but doesn't (seccomp, permissions), this is invisible. Can you add a debug log in the else branch?

```go
} else {
    c.logger.Debugf("cgroup memory info unavailable, using host values: %v", cgroupErr)
}
```

umh-core/pkg/service/container_monitor/container_monitor.go
```diff
@@ -349,6 +349,15 @@ func (c *ContainerMonitorService) getMemoryMetrics(ctx context.Context) (*models
 	usedBytes := vmStat.Used
 	totalBytes := vmStat.Total
 
+	// Try cgroup values — prefer container-aware limits over host values
+	cgroupInfo, cgroupErr := c.getCgroupMemoryInfo(ctx)
+	if cgroupErr == nil {
+		usedBytes = uint64(cgroupInfo.CurrentBytes)
+		if !cgroupInfo.Unlimited && cgroupInfo.LimitBytes > 0 {
```
JeremyTheocharis, line 356: When unlimited, `usedBytes` comes from cgroup but `totalBytes` stays as the host value. This is correct (matches how CPU handles unlimited quota in `cgroup_cpu.go`), but worth a comment so the next reader doesn't wonder:

```go
if !cgroupInfo.Unlimited && cgroupInfo.LimitBytes > 0 {
    // Only override totalBytes when a cgroup limit is set.
    // When unlimited, keep host total (same approach as CPU with unlimited quota).
    totalBytes = uint64(cgroupInfo.LimitBytes)
}
```

umh-core/pkg/service/container_monitor/container_monitor.go
```diff
@@ -349,6 +349,15 @@ func (c *ContainerMonitorService) getMemoryMetrics(ctx context.Context) (*models
 	usedBytes := vmStat.Used
 	totalBytes := vmStat.Total
 
+	// Try cgroup values — prefer container-aware limits over host values
```
JeremyTheocharis: Nit: our remove-AI-slop skill flags em-dashes as an AI writing pattern. Not a hard rule, but easy swap:
```go
// Try cgroup values: prefer container-aware limits over host values
```

led0nk: Fixed, replaced em-dash with colon.

umh-core/pkg/service/container_monitor/container_monitor.go
led0nk: Added the clarifying comment about unlimited keeping host total.

umh-core/pkg/service/container_monitor/container_monitor.go
led0nk: Added `else` branch with `Debugf` for cgroup fallback.

led0nk:
Changelog entry added under Fixes in 0.44.12.

JeremyTheocharis (approved):
Re-review with 6 agents + devil's advocate.

Previous comments: 2/3 confirmed fixed (debug log, em-dash colon swap). The clarifying comment for unlimited keeping host total, can you check if it's committed locally but not pushed?

Looks good. The unlimited-mode behavior (`container_used / host_total`) is correct for a container monitor. It answers "what fraction of available memory is my container using?" and the 80% alert threshold fires when it should.

Side finding: CPU has the same pre-PR problem (`cpu.PercentWithContext` reads host-wide `/proc/stat`). Already tracked in ENG-3423.

Nits (non-blocking):
- `getMemoryMetrics` godoc (line 341-342) still says "collects memory metrics using gopsutil / host-level usage", now stale
- `ParseMemoryMax`/`ParseMemoryCurrent` are exported but `cgroup_cpu.go` equivalents (`parseCPUMax`/`parseCPUStats`) are unexported

_6 agents, DA confirmed behavior correct for container use case_

PR #2474, opened 2026-03-16
fix(fsmv2): make StoppingState transition unconditional (ENG-4608)

JeremyTheocharis:
## The bug

An instance appeared degraded in the Management Console even though the pod was running and healthy. The communicator's pull worker was stuck in `Stopping` state indefinitely. Only a pod restart fixed it.

## Reproduction test (written before the fix)

We wrote a test before touching any code to confirm this is a state machine bug, not a configuration issue.

**Before the fix:** The test creates a pull worker in `StoppingState` and gives it a snapshot where the parent has recovered to `Running`. The worker stays stuck — `StoppingState.Next()` returns itself forever. The test passed on the first run, confirming the dead end.

**After the fix:** The same test verifies recovery: `Stopping → Stopped → Running`. The worker passes through `StoppedState`, which checks `ShouldBeRunning()` and transitions back to `Running`.

The push worker has the identical bug — same test, same result.

## Root cause

FSMv2 uses a parent-child worker hierarchy: the **transport** worker manages **push** (outbound) and **pull** (inbound) as children. Children only run when the parent is in certain states — configured via `ChildStartStates: ["Running", "Degraded"]`.

The pull and push workers have a `StoppingState` that checks `IsStopRequired()` each tick:
- If true: transition to `Stopped`
- If false: stay in `Stopping`

There is no other exit.

This breaks during **token re-authentication**. The transport parent enters `Starting` to refresh its JWT. The supervisor sees the parent leave `Running` and tells children to stop — because `"Starting"` is not in their allowed parent states. Children enter `Stopping`. But re-auth completes in a single tick. By the next tick the parent is back in `Running`, children get told to run again, `IsStopRequired()` flips to false — and the child is trapped. It cannot go forward to `Stopped` (requires true) and has no path back to `Running`.

The cascade:

```
pull-001 stuck in Stopping (supervisor counts this as unhealthy)
  → transport-001 Degraded ("children unhealthy (1)")
    → communicator-001 Recovering ("children unhealthy: healthy=0, unhealthy=1")
      → Management Console shows instance degraded/offline
```

A second problem made this invisible in logs: reason strings read `ParentMappedState` from the wrong struct field — one that is never populated. Every log line showed `parentState=` regardless of the actual state.

### 1. Make StoppingState transition unconditional

`StoppingState` now always transitions to `Stopped`. The decision to stop was already made when entering `Stopping`. `StoppedState` handles recovery — it checks `ShouldBeRunning()` and transitions back to `Running` when the parent is healthy.

`StoppingState` is preserved as a cleanup lifecycle hook — the structure and comments are in place for future resource cleanup (close connections, drain buffers). Self-return during cleanup is valid but must carry an action, never nil.

### 2. AST validator: prevent this bug class at CI time

New architecture rule `ValidateStoppingStateNoCatchAllSelfReturn`: states embedding `StoppingBase` must not have a catch-all self-return with nil action. Self-return WITH an action (cleanup in progress) is allowed. This catches the exact pattern that caused this bug and prevents it in any future worker.

### 3. Fix ParentMappedState logging

Changed `snap.Desired.ParentMappedState` → `snap.Observed.ParentMappedState` in reason strings across all state files. Added `parentState(observed)=` label so operators immediately know which field is being logged.

## Full audit: all FSMv2 stopping states

| Worker | State | Verdict |
|--------|-------|---------|
| **Transport** | `StoppingState` | Fixed in this PR |
| **Push** | `StoppingState` | Fixed in this PR |
| **Pull** | `StoppingState` | Fixed in this PR |
| Persistence | `ShuttingDownState` | Safe — always carries `RunMaintenanceAction` |
| ExampleChild | `TryingToStopState` | Safe — carries `DisconnectAction` |
| ExampleParent | `TryingToStopState` | Safe — waits for children (supervisor-controlled) |
| ExampleFailing | `TryingToStopState` | Safe — carries action |
| ExamplePanic | `TryingToStopState` | Safe — carries action |
| ExampleSlow | `TryingToStopState` | Safe — carries action |

## Test plan

- [x] Reproduction test written before the fix: asserts `StoppingState.Next()` returns `StoppingState` (stuck) when parent is Running — confirms the dead end
- [x] After fix: same test asserts `StoppingState.Next()` returns `StoppedState`, then `StoppedState.Next()` returns `RunningState` (full recovery)
- [x] Scenario tests: shutdown cancelled mid-stop, parent flapping between Running and Starting
- [x] Architecture tests pass (including new `ValidateStoppingStateNoCatchAllSelfReturn`)
- [x] All transport worker tests pass
- [ ] Confirm no recurrence on affected instance after deploy

### Preview: FSMv2 Communicator

- Instances could appear permanently offline in the Management Console even though the pod was running and healthy -- only a restart would fix it. This happened because token re-authentication briefly caused child workers to enter a stopping state from which they could not recover, leaving the communicator stuck indefinitely. Workers now always complete the stop and automatically recover when the parent is healthy again. Requires `USE_FSMV2_TRANSPORT=true`

### DIFF
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -7,6 +7,10 @@
 - The container previously restarted hundreds of times per minute when config.yaml was missing or invalid and now waits 60 seconds before retrying, giving you time to fix the configuration
 - Previously, memory monitoring read host-level values even inside containers, while CPU monitoring already used cgroup-aware values. Memory metrics now read cgroup v2 limits and usage, so dashboards show correct container memory utilization
 
+### Preview: FSMv2 Communicator
+
+- Instances could appear permanently offline in the Management Console even though the pod was running and healthy -- only a restart would fix it. This happened because token re-authentication briefly caused child workers to enter a stopping state from which they could not recover, leaving the communicator stuck indefinitely. Workers now always complete the stop and automatically recover when the parent is healthy again. Requires `USE_FSMV2_TRANSPORT=true`
+
 ## [0.44.11]
 
 This release reduces steady-state container memory by roughly 30% on busy systems with many bridges and data flows.
--- a/umh-core/pkg/fsmv2/CLAUDE.md
+++ b/umh-core/pkg/fsmv2/CLAUDE.md
@@ -64,7 +64,7 @@ Each state file follows this pattern:
 
 ```go
 type RunningState struct {
-    helpers.BaseRunningState  // Embed exactly one base type
+    helpers.RunningHealthyBase  // Embed exactly one base type
 }
 
 func (s *RunningState) Next(snapAny any) fsmv2.NextResult[any, any] {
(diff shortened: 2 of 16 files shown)


JeremyTheocharis:
may the klaus roasting come in

led0nk (approved):
Other than the klaus-findings it looks good to me

PR #2479, opened 2026-03-18
feat(fsmv2): pass DesiredState to CollectObservedState (ENG-4638)

JeremyTheocharis:
## Summary

- Adds `desired DesiredState` parameter to the `CollectObservedState` interface
- Updates all 12 worker implementations (accept and ignore the new parameter)
- 37 files, 140 insertions, 101 deletions. One-time migration.

## The problem

`CollectObservedState` can't access configuration. For a worker that replaces the FSMv1 nmap/connection stack with a TCP reachability check, the implementation looks like this:

```go
func (w *Worker) CollectObservedState(ctx context.Context) (ObservedState, error) {
    // I need to dial target:port... but where do I get target and port?
    // DesiredState has them, but I can't see DesiredState here.
}
```

Actions can read `snap.Desired` in `State.Next()` and carry config on their struct. The sensor can't. So config-dependent I/O gets routed through Actions instead: emit a `ConfigureAction` that copies IP/port to Dependencies, wait a tick, then read from Dependencies. That's startup latency and boilerplate for every config field.

Most new workers need config-dependent observation. The FSMv1-to-FSMv2 migration alone has several (the nmap/connection stack, protocol converter health checks), and each one hits this same workaround.

## The fix

```go
// Before
CollectObservedState(ctx context.Context) (ObservedState, error)

// After
CollectObservedState(ctx context.Context, desired DesiredState) (ObservedState, error)
```

The supervisor loads DesiredState from the store and passes it before each collection call. Existing workers accept the parameter as `_`. Zero behavior change.

## Validation

In the documentation stress test described in the base PR (#2482), 5 implementation attempts on the old interface all placed `net.DialTimeout` inside an Action. After this interface change, the agent placed it in `CollectObservedState` and read host/port from `desired`. Full methodology and artifacts are in the base PR.

N=1, so take it accordingly. More runs planned.

## Design concerns

Three things that came up:

1. **Observation/intent boundary**: Already porous. Every `ObservedState` embeds `DesiredState` via `json:",inline"`, and `GetObservedDesiredState()` pairs them at the snapshot level. The coupling is in the data model already, this just makes the code honest about it.

2. **Goroutine safety**: Passing desired to the sensor means each observation records which config was in effect when it ran. Same pattern as `ShutdownRequestedProvider`.

3. **Slippery slope** (developers using `desired` for policy decisions in collectors): We can guard against this with architecture tests that detect calls to `desired.GetState()` or `desired.IsShutdownRequested()` inside `CollectObservedState`.

Outstanding work:
- [ ] Architecture test guard against policy methods in `CollectObservedState`
- [ ] `doc.go` update documenting permitted vs prohibited uses
- [ ] Migration guide for updating worker implementations

## Stacking

This PR sits on top of #2482, which rewrites the FSMv2 README to teach the sensor/controller/actuator framing that motivates this interface change. The base PR fixes the docs, this one fixes the interface.

## Test plan

- [x] `go build ./pkg/fsmv2/...` -- clean
- [x] `go test ./pkg/fsmv2/ -v -count=1` -- 39/39 architecture tests pass
- [x] Full test suite: 54/54 pass
- [ ] Address outstanding design concern items above

### DIFF
--- a/umh-core/pkg/fsmv2/api.go
+++ b/umh-core/pkg/fsmv2/api.go
@@ -186,7 +186,11 @@ func Result[TSnapshot any, TDeps any](
 // not by a method on this interface.
 type Worker interface {
 	// CollectObservedState monitors the actual system state.
-	CollectObservedState(ctx context.Context) (ObservedState, error)
+	// The desired parameter provides the current desired state so observation-based
+	// workers can access configuration (target IP, port, etc.) without workarounds.
+	// The supervisor guarantees desired is always non-nil; collection is skipped
+	// until a desired state exists in the store.
+	CollectObservedState(ctx context.Context, desired DesiredState) (ObservedState, error)
 
 	// DeriveDesiredState derives the target state from user configuration (spec).
 	DeriveDesiredState(spec interface{}) (DesiredState, error)
--- a/umh-core/pkg/fsmv2/examples/helloworld.go
+++ b/umh-core/pkg/fsmv2/examples/helloworld.go
@@ -18,10 +18,10 @@ package examples
 //
 // Demonstrates both I/O patterns:
 //   - Action: SayHelloAction writes to in-memory deps
-//   - Observation: CollectObservedState reads the mood file path from disk
+//   - Observation: CollectObservedState reads the mood file from disk
 //
-// The mood file path is configured via DesiredState (moodFilePath).
-// When omitted, mood checking is skipped entirely (backwards compatible).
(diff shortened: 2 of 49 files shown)


JeremyTheocharis:
Superseded by Worker API v2 refactor (PR1: feat/worker-api-v2-foundation). All commits from this PR (including the DesiredState interface change) are incorporated in the refactor branch.

PR #2480, opened 2026-03-19
Fix: cpu throttle ratio

led0nk:
# Description

- CPU throttle ratio was calculated from cumulative cgroup counters, causing permanent Degraded status after any brief throttle event
- Now uses a 5-second sliding window reflecting recent behavior, allowing recovery to Active when throttling stops

Fixes ENG-4470

### DIFF
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Fixes
 
+- CPU health stayed Degraded permanently after any brief throttle event because the ratio used cumulative counters since pod start. Now uses a sliding window, so health recovers when throttling stops
 - ADS symbol downloads failed in certain configurations -- bumped ADS plugin to v1.0.8 which fixes the issue
 
 ### Preview: FSMv2 Communicator
--- a/umh-core/pkg/constants/container.go
+++ b/umh-core/pkg/constants/container.go
@@ -49,4 +49,9 @@ const (
 	// CPUThrottleRatioThreshold defines when CPU throttling is considered significant.
 	// If more than 5% of periods are throttled, the CPU is considered throttled.
 	CPUThrottleRatioThreshold = 0.05
+
+	// CPUThrottleWindow defines the sliding window duration for throttle ratio calculation.
+	// At the 5% threshold, a 5-second window requires ~250ms of sustained throttling
+	// (2-3 CFS periods) before flagging degraded, filtering out transient spikes.
+	CPUThrottleWindow = 5 * time.Second
 )
--- a/umh-core/pkg/constants/container.go
+++ b/umh-core/pkg/constants/container.go
@@ -51,7 +51,7 @@ const (
 	CPUThrottleRatioThreshold = 0.05
 
 	// CPUThrottleWindow defines the sliding window duration for throttle ratio calculation.
-	// At the 5% threshold, a 5-second window requires ~250ms of sustained throttling
-	// (2-3 CFS periods) before flagging degraded, filtering out transient spikes.
-	CPUThrottleWindow = 5 * time.Second
+	// 60 seconds aligns with Prometheus/K8s rate windows for counter-derived metrics
+	// and avoids rapid Degraded/Active flips during bursty throttle events.
+	CPUThrottleWindow = 60 * time.Second
 )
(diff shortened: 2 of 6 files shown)


umh-core/pkg/constants/container.go
```diff
@@ -49,4 +49,9 @@ const (
 	// CPUThrottleRatioThreshold defines when CPU throttling is considered significant.
 	// If more than 5% of periods are throttled, the CPU is considered throttled.
 	CPUThrottleRatioThreshold = 0.05
+
+	// CPUThrottleWindow defines the sliding window duration for throttle ratio calculation.
+	// At the 5% threshold, a 5-second window requires ~250ms of sustained throttling
+	// (2-3 CFS periods) before flagging degraded, filtering out transient spikes.
+	CPUThrottleWindow = 5 * time.Second
 )
```
JeremyTheocharis, line 57: Klaus flagged (MEDIUM): Why 5 seconds? This signal drives bridge creation blocking via `IsResourceLimited()`. At 100ms reconcile ticks, a brief throttle burst flips health to Degraded for ~5s, clears, flips again on the next burst. The old bug was "permanently stuck Degraded" - 5s swings to the opposite extreme. I'd say 60 seconds. That's long enough to smooth out transient spikes, short enough to recover within a minute. Also lines up with how Prometheus/K8s rate windows work for counter-derived metrics.

_I found this, and then all of the Klaus horde reported this independently as well_

Also nit on the comment above: "~250ms of sustained throttling (2-3 CFS periods)" conflates throttled *periods* with throttled *time*. `nr_throttled` counts periods where any throttling occurred (could be 1us of throttling in a 100ms period). "2-3 CFS periods" is correct, "250ms" is not.

umh-core/pkg/service/container_monitor/container_monitor.go
```diff
@@ -259,8 +267,15 @@ func (c *ContainerMonitorService) getCPUMetrics(ctx context.Context) (*models.CP
...
-	// Check for throttling - only if cgroupInfo is not nil
-	isThrottled := cgroupErr == nil && cgroupInfo != nil && cgroupInfo.IsThrottled
+	// Compute windowed throttle ratio (replaces cumulative ratio from cgroup)
+	var windowedRatio float64
+	var isThrottled bool
+	if cgroupErr == nil && cgroupInfo != nil {
+		windowedRatio, isThrottled = c.updateThrottleWindow(cgroupInfo)
+		// Override the cumulative values with windowed values
+		cgroupInfo.ThrottleRatio = windowedRatio
+		cgroupInfo.IsThrottled = isThrottled
```
JeremyTheocharis, line 278: Klaus flagged (MEDIUM): The `Warnf` at line 295 fires every tick (~10/s) while throttled. The container monitor uses the global logger (no SentryHook), so it won't flood Sentry, but it will flood stdout. With the new windowed approach it fires in bursts rather than continuously. Either downgrade to `Debugf` (throttling is already visible via health status) or only log on the false->true transition. If you want actual Sentry alerts for throttle events, wire up the FSMv2 SentryHook, but decide whether you actually want that notification volume.
_3 agents · DA promoted from LOW_

JeremyTheocharis (commented):
Good fix, two things to address + a few comment corrections.

Not in the diff but worth fixing:

**nit — Dead cumulative calc (`cgroup_cpu.go:142-146`):**
These lines still compute `ThrottleRatio`/`IsThrottled` from cumulative counters, but `getCPUMetrics()` overrides both at lines 276-277. Add a comment, or better: delete the calculation and have `CPUCgroupInfo` carry only raw counters.
_6 agents · DA confirmed at LOW_

**nit — `status_message.go:129`:**
Comment says "Ratio of time throttled" but the value is ratio of throttled *periods*, not time.
_1 agent · DA confirmed_

umh-core/pkg/constants/container.go
JeremyTheocharis: correction to be clear: i found it, and then klaus found it independently

led0nk: changed to 60s

umh-core/pkg/service/container_monitor/container_monitor.go
led0nk: changed the logging to only occur on positive transition `false -> true`

umh-core/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Fixes
 
+- CPU health stayed Degraded permanently after any brief throttle event because the ratio used cumulative counters since pod start. Now uses a 5-second sliding window, so health recovers when throttling stops
```
JeremyTheocharis: nit: still says "5-second sliding window" but `CPUThrottleWindow` is now 60s. Update to "60-second" or just drop the duration.

JeremyTheocharis (approved):
All review items addressed correctly — 60s window, transition-only logging, dead code removed, comment fix. One changelog nit inline.

PR #2482, opened 2026-03-20
docs(fsmv2): rewrite README with control loop framing (ENG-4634)

JeremyTheocharis:
## Summary

- Adds "Modeling your worker" section that teaches when to use observations vs actions (sensor/controller/actuator framing)
- Updates all code examples to match the actual API (`fsmv2.Result[]`, pointer receivers, reason strings)
- Extends helloworld with DegradedState and a standalone scenario
- Fixes stale type names and jargon in CLAUDE.md and README

## The problem

The FSMv1 nmap/connection stack is two layers (Nmap FSM + Connection FSM, ~800 lines) that should collapse into one FSMv2 worker (~200 lines) using `net.DialTimeout`. The workshop spec even asks the right question: "Is checking whether a TCP port is open an action or an observation?"

We stress-tested the docs by giving AI agents this migration task. Every agent knew that `CollectObservedState` is the right place for `net.DialTimeout`. When asked directly, they said so. But when implementing, all of them put the TCP dial inside an Action.

The docs weren't factually wrong. Every API was correctly documented. But the README never told you how to decide which component to use. Without that, developers default to whatever pattern has the most examples, which is Actions.

This is half the problem. The other half, that `CollectObservedState` can't access configuration like target IP/port, is in the stacked PR #2479 which adds `desired DesiredState` as a parameter to `CollectObservedState`.

## The experiment

We ran 5 implementation attempts on the old interface with varying levels of guidance (full spec, outline only, direct question). All 5 placed `net.DialTimeout` inside an Action.

After both changes (this README rewrite + the interface change in PR #2479), one agent immediately placed the dial in `CollectObservedState` and read host/port from the `desired` parameter. So: 5 failures on the old setup, 1 success on the new. N=1, so take it accordingly. We plan more runs.

## What changed

New "Modeling your worker" section:
- `CollectObservedState` = sensor (read the world)
- `State.Next()` = controller (pure logic, no I/O)
- Actions = actuator (change the world)

The control loop was previously described in three places with different terminology. Now there's one description and the other two reference it.

Helloworld is now a teaching example:
- DegradedState shows observation-driven transitions (Running to Degraded based on mood file)
- Standalone `--scenario=helloworld` for quick experimentation
- Control loop mapping table in the helloworld README

Code examples updated to the actual API: pointer receivers, `fsmv2.Result[]`, `helpers.ConvertSnapshot`, `helpers.DeriveStateName` (PascalCase). The I/O isolation table now says READ/WRITE instead of YES/NO, which is clearer.

README: "safety system" changed to "watchdog" (avoids SIS regulatory connotation), "error signal" changed to "health checks". CLAUDE.md: `BaseRunningState` changed to `RunningHealthyBase`.

## Stacking

Base of a 2-PR stack. PR #2479 sits on top and adds `desired DesiredState` to `CollectObservedState`, giving the sensor access to configuration. This PR fixes the docs, PR #2479 fixes the interface.

## Test plan

- [x] `go build ./pkg/fsmv2/...`
- [x] `go test ./pkg/fsmv2/ -v -count=1` -- 40 of 41 architecture tests pass (1 skipped)
- [x] `make betteralign` -- clean
- [x] `--scenario=helloworld` runs: Stopped -> TryingToStart -> Running
- [ ] Re-run agent experiment with updated README only (no interface change) to isolate the documentation improvement

### DIFF
--- a/umh-core/pkg/fsmv2/README.md
+++ b/umh-core/pkg/fsmv2/README.md
@@ -487,7 +487,7 @@ go run pkg/fsmv2/cmd/runner/main.go --scenario=simple --duration=10s
 go run pkg/fsmv2/cmd/runner/main.go --scenario=communicator --log-level=debug
 ```
 
-Built-in scenarios: `simple`, `failing`, `panic`, `cascade`, `timeout`, `communicator`
+Built-in scenarios: `helloworld`, `simple`, `failing`, `panic`, `slow`, `cascade`, `timeout`, `configerror`, `inheritance`, `communicator`, `concurrent`, `persistence`
 
 ### Unit tests
 
--- a/umh-core/pkg/fsmv2/README.md
+++ b/umh-core/pkg/fsmv2/README.md
@@ -11,7 +11,8 @@ Type-safe state machine framework for managing worker lifecycles with compile-ti
 ```bash
 # Run the existing examples
 go run pkg/fsmv2/cmd/runner/main.go --list          # List all scenarios
-go run pkg/fsmv2/cmd/runner/main.go --scenario=simple --duration=5s
+go run pkg/fsmv2/cmd/runner/main.go --scenario=helloworld --duration=5s  # Minimal single worker
+go run pkg/fsmv2/cmd/runner/main.go --scenario=simple --duration=5s      # Parent with 2 children
 ```
 
 ### Creating a New Worker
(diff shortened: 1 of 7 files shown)


JeremyTheocharis:
Superseded by Worker API v2 refactor (PR1: feat/worker-api-v2-foundation). All commits from this PR are incorporated in the refactor branch.

PR #2492, opened 2026-03-23
test(fsmv2): add failing reproduction test for parent health accounting bug (ENG-4662)

JeremyTheocharis:
## Problem

Parent supervisor health accounting returns healthy=0, unhealthy=0 for all child phases. The communicator stays in "Recovering" permanently because it can't count its children's health.

Found during ENG-4576 container experiment: communicator-001 showed `healthy=0, unhealthy=0` for 190 seconds while transport-001 was in AuthFailed. The counting logic (ChildrenManager.Counts) works correctly when called directly. The bug is in the collector pipeline that connects the counting logic to the parent's observed state.

## What this PR adds

A reproduction test file with 6 test cases:

**5 passing** (prove the counting logic works):
- PhaseStarting child counted as unhealthy=1
- PhaseRunningHealthy child counted as healthy=1
- PhaseStopped child counted as neither
- PhaseRunningDegraded child counted as unhealthy=1
- AddWorker seeds lastLifecyclePhase correctly

**1 failing** (reproduces the bug):
- Full collector pipeline: TestTick on parent with PhaseStarting child, assert unhealthy=1
- Gets (0,0) because the collector pipeline doesn't deliver counts to the observed state

The failing test turns green when the pipeline bug is fixed.

Resolves: ENG-4662. Parent: ENG-4231.

## Test plan

- [x] 5 passing tests prove counting logic correct
- [x] 1 failing test reproduces the production bug
- [x] 4 rounds of DA review (converged at round 4: forced to hallucinate)
- [x] Cross-checked against container experiment evidence
- [x] Existing supervisor tests unaffected (248 specs pass)

### DIFF
--- a/umh-core/pkg/fsmv2/supervisor/children_health_accounting_test.go
+++ b/umh-core/pkg/fsmv2/supervisor/children_health_accounting_test.go
@@ -1,0 +1,325 @@
+// Copyright 2025 UMH Systems GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package supervisor_test
+
+import (
+	"context"
+	"sync"
+	"time"
+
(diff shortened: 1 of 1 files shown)


JeremyTheocharis:
Closing — cannot reproduce.

A full-pipeline Ginkgo scenario test (RunCommunicatorScenario with permanent 401 auth rejection) confirms the collector goroutine correctly enriches children counts: `healthy=0, unhealthy=1`. The collector pipeline works end-to-end.

The FAILING unit test here fails because TestTick() doesn't start the collector goroutine — that's by design (TestTick is synchronous, the collector is async). It's a test infrastructure limitation, not a production bug.

The 190-second (0,0) observation from the ENG-4576 container experiment (`sha-91ddda5`) was likely specific to that build or experiment setup. No customer has reported this.

PR #2495, opened 2026-03-25
feat: read and write flows for protocol converter

FalconTube:
- Adds separate `write flow` to protocol converters (bridge)
- Check benthos for `read flow` and `write flow` separately
- Has ability to only deploy one of them or both
- From my testing, it still works with `old` and `new` frontend i.e. with or without the FF being set

### DIFF
--- a/umh-core/pkg/communicator/actions/edit-protocolconverter.go
+++ b/umh-core/pkg/communicator/actions/edit-protocolconverter.go
@@ -41,6 +41,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/tiendc/go-deepcopy"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/config"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/config/benthosserviceconfig"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/config/connectionserviceconfig"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/config/dataflowcomponentserviceconfig"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/config/protocolconverterserviceconfig"
@@ -62,6 +63,8 @@ const (
 	DFCTypeRead DFCType = "read"
 	// DFCTypeWrite represents a write dataflow component.
 	DFCTypeWrite DFCType = "write"
+	// DFCTypeBoth represents both read and write dataflow components being updated simultaneously.
+	DFCTypeBoth DFCType = "both"
 	// DFCTypeEmpty represents no dataflow component (connection/location update only).
 	DFCTypeEmpty DFCType = "empty"
 )
@@ -74,7 +77,7 @@ func (d DFCType) String() string {
 // IsValid checks if the DFCType has a valid value.
 func (d DFCType) IsValid() bool {
 	switch d {
-	case DFCTypeRead, DFCTypeWrite, DFCTypeEmpty:
+	case DFCTypeRead, DFCTypeWrite, DFCTypeBoth, DFCTypeEmpty:
 		return true
 	default:
 		return false
@@ -98,14 +101,17 @@ type EditProtocolConverterAction struct {
 	dfcType        DFCType
 	connectionPort string
 	connectionIP   string
-	state          string
+	readDFCState   string // desired state for the read DFC ("active" or "stopped"; empty = default active)
+	writeDFCState  string // desired state for the write DFC ("active" or "stopped"; empty = default active)
 
-	dfcPayload models.CDFCPayload
+	readDFCPayload  *models.CDFCPayload
+	writeDFCPayload *models.CDFCPayload
 
 	vb []models.ProtocolConverterVariable
 
-	// Desired DFC config for comparison during health checks
-	desiredDFCConfig dataflowcomponentserviceconfig.DataflowComponentServiceConfig
+	// Desired DFC configs for comparison during health checks
+	desiredReadDFCConfig  dataflowcomponentserviceconfig.DataflowComponentServiceConfig
+	desiredWriteDFCConfig dataflowcomponentserviceconfig.DataflowComponentServiceConfig
 
 	actionUUID   uuid.UUID
 	instanceUUID uuid.UUID
@@ -145,25 +151,44 @@ func (a *EditProtocolConverterAction) Parse(payload interface{}) error {
 	if pcPayload.UUID == nil {
 		return errors.New("missing required field UUID")
 	}
-
 	a.protocolConverterUUID = *pcPayload.UUID
 	a.name = pcPayload.Name
 
-	// Determine which DFC is being updated and convert it to CDFCPayload
-	var dfcToUpdate *models.ProtocolConverterDFC
-
+	// Determine which DFC(s) are being updated and convert to CDFCPayload.
 	switch {
+	case pcPayload.ReadDFC != nil && pcPayload.WriteDFC != nil:
+		a.dfcType = DFCTypeBoth
+		readPayload := dfcToPayload(pcPayload.ReadDFC)
+		writePayload := dfcToPayload(pcPayload.WriteDFC)
+		a.readDFCPayload = &readPayload
+		a.writeDFCPayload = &writePayload
+		if pcPayload.ReadDFC.IgnoreErrors != nil {
+			a.ignoreHealthCheck = *pcPayload.ReadDFC.IgnoreErrors
+		}
+		if pcPayload.WriteDFC.IgnoreErrors != nil {
+			a.ignoreHealthCheck = a.ignoreHealthCheck || *pcPayload.WriteDFC.IgnoreErrors
+		}
+		a.readDFCState = pcPayload.ReadDFC.State
+		a.writeDFCState = pcPayload.WriteDFC.State
 	case pcPayload.ReadDFC != nil:
 		a.dfcType = DFCTypeRead
-		dfcToUpdate = pcPayload.ReadDFC
+		readPayload := dfcToPayload(pcPayload.ReadDFC)
+		a.readDFCPayload = &readPayload
+		if pcPayload.ReadDFC.IgnoreErrors != nil {
+			a.ignoreHealthCheck = *pcPayload.ReadDFC.IgnoreErrors
+		}
+		a.readDFCState = pcPayload.ReadDFC.State
 	case pcPayload.WriteDFC != nil:
 		a.dfcType = DFCTypeWrite
-		dfcToUpdate = pcPayload.WriteDFC
+		writePayload := dfcToPayload(pcPayload.WriteDFC)
+		a.writeDFCPayload = &writePayload
+		if pcPayload.WriteDFC.IgnoreErrors != nil {
+			a.ignoreHealthCheck = *pcPayload.WriteDFC.IgnoreErrors
+		}
+		a.writeDFCState = pcPayload.WriteDFC.State
 	default:
 		a.dfcType = DFCTypeEmpty
-		dfcToUpdate = &models.ProtocolConverterDFC{}
 	}
-
 	if pcPayload.TemplateInfo != nil {
 		a.vb = pcPayload.TemplateInfo.Variables
 	} else {
@@ -178,30 +203,8 @@ func (a *EditProtocolConverterAction) Parse(payload interface{}) error {
 	a.connectionPort = strconv.Itoa(int(pcPayload.Connection.Port))
 	a.connectionIP = pcPayload.Connection.IP
 
-	// Convert ProtocolConverterDFC to CDFCPayload for internal processing
-	if a.dfcType != DFCTypeEmpty {
-		a.dfcPayload = models.CDFCPayload{
-			Inputs:   models.DfcDataConfig{Data: dfcToUpdate.Inputs.Data, Type: dfcToUpdate.Inputs.Type},
-			Pipeline: convertPipelineToMap(dfcToUpdate.Pipeline),
-			// Set default outputs since ProtocolConverterDFC doesn't have outputs
-			Outputs: models.DfcDataConfig{Data: "", Type: ""},
-			// Extract Inject data from RawYAML to support CacheResources, RateLimitResources, and Buffer
-			Inject: extractInjectFromRawYAML(dfcToUpdate.RawYAML),
-		}
-		// Handle optional fields
-		if dfcToUpdate.IgnoreErrors != nil {
-			a.ignoreHealthCheck = *dfcToUpdate.IgnoreErrors
-		}
-	}
-
-	// Extract the desired state, defaulting to "active" if not provided
-	a.state = pcPayload.State
-	if a.state == "" {
-		a.state = protocolconverter.OperationalStateActive
-	}
-
-	a.actionLogger.Debugf("Parsed EditProtocolConverter action payload: uuid=%s, name=%s, dfcType=%s, state=%s",
-		a.protocolConverterUUID, a.name, a.dfcType, a.state)
+	a.actionLogger.Debugf("Parsed EditProtocolConverter action payload: uuid=%s, name=%s, dfcType=%s, readDFCState=%s, writeDFCState=%s",
+		a.protocolConverterUUID, a.name, a.dfcType, a.readDFCState, a.writeDFCState)
 
 	return nil
 }
@@ -217,13 +220,10 @@ func (a *EditProtocolConverterAction) Validate() error {
 		return err
 	}
 
-	if a.dfcType != DFCTypeEmpty {
-		if err := ValidateCustomDataFlowComponentPayload(a.dfcPayload, false); err != nil {
-			return fmt.Errorf("invalid dataflow component configuration: %w", err)
-		}
+	if err := validateDFCPayloadAndState(a.readDFCPayload, a.readDFCState, "read"); err != nil {
+		return err
 	}
-
-	if err := ValidateDataFlowComponentState(a.state); err != nil {
+	if err := validateDFCPayloadAndState(a.writeDFCPayload, a.writeDFCState, "write"); err != nil {
 		return err
 	}
 
@@ -247,23 +247,36 @@ func (a *EditProtocolConverterAction) Execute() (interface{}, map[string]interfa
 		confirmationMessage, a.outboundChannel, models.EditProtocolConverter)
 
 	var (
-		benthosConfig dataflowcomponentserviceconfig.BenthosConfig
-		err           error
+		readBenthosConfig  *dataflowcomponentserviceconfig.BenthosConfig
+		writeBenthosConfig *dataflowcomponentserviceconfig.BenthosConfig
+		err                error
 	)
 
-	// Only create Benthos config if we have a DFC to configure
-
-	if a.dfcType != DFCTypeEmpty {
-		// Convert the DFC payload to BenthosConfig
-		benthosConfig, err = CreateBenthosConfigFromCDFCPayload(a.dfcPayload, a.name)
-		if err != nil {
-			errorMsg := fmt.Sprintf("Failed to create Benthos configuration: %v", err)
+	if a.readDFCPayload != nil {
+		cfg, buildErr := CreateBenthosConfigFromCDFCPayload(*a.readDFCPayload, a.name)
+		if buildErr != nil {
+			errorMsg := fmt.Sprintf("Failed to create read DFC Benthos configuration: %v", buildErr)
 			SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionFinishedWithFailure,
 				errorMsg, a.outboundChannel, models.EditProtocolConverter)
 
 			return nil, nil, fmt.Errorf("%s", errorMsg)
 		}
+		readBenthosConfig = &cfg
+	}
 
+	if a.writeDFCPayload != nil {
+		cfg, buildErr := CreateBenthosConfigFromCDFCPayload(*a.writeDFCPayload, a.name)
+		if buildErr != nil {
+			errorMsg := fmt.Sprintf("Failed to create write DFC Benthos configuration: %v", buildErr)
+			SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionFinishedWithFailure,
+				errorMsg, a.outboundChannel, models.EditProtocolConverter)
+
+			return nil, nil, fmt.Errorf("%s", errorMsg)
+		}
+		writeBenthosConfig = &cfg
+	}
+
+	if a.dfcType != DFCTypeEmpty {
 		SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionExecuting,
 			fmt.Sprintf("Updating protocol converter configuration with %s DFC...", a.dfcType.String()),
 			a.outboundChannel, models.EditProtocolConverter)
@@ -274,7 +287,7 @@ func (a *EditProtocolConverterAction) Execute() (interface{}, map[string]interfa
 	}
 
 	// Apply mutations to create new spec
-	newSpec, atomicEditUUID, err := a.applyMutation(benthosConfig)
+	newSpec, atomicEditUUID, desiredPCState, err := a.applyMutation(readBenthosConfig, writeBenthosConfig)
 	if err != nil {
 		errorMsg := fmt.Sprintf("Failed to apply configuration mutation: %v", err)
 		SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionFinishedWithFailure,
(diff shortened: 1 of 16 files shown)


umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -449,21 +470,21 @@ func (a *EditProtocolConverterAction) applyMutation(benthosConfig dataflowcompon
...
+
+	// Derive the PC-level FSM state from the per-DFC states.
+	// For DFCTypeEmpty (connection/location-only edits) the new states are empty,
+	// so we fall back to the existing spec values via derivePCDesiredState.
+	instanceToModify.DesiredFSMState = derivePCDesiredState(
+		a.readDFCState,
+		a.writeDFCState,
+		instanceToModify.ProtocolConverterServiceConfig.ReadDFCDesiredState,
+		instanceToModify.ProtocolConverterServiceConfig.WriteDFCDesiredState,
+	)
```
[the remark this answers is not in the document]
FalconTube: Not an issue, because state is always read out in the frontend and then sent along with the payload, if it was already present

umh-core/pkg/fsm/protocolconverter/actions.go
```diff
@@ -435,24 +437,49 @@ func (p *ProtocolConverterInstance) IsRedpandaHealthy() (bool, string) {
...
+	}
+
+	statusReason := p.ObservedState.ServiceInfo.DataflowComponentWriteObservedState.ServiceInfo.StatusReason
 	if statusReason == "" {
 		statusReason = "flow health status unknown"
 	}
 
-	// TODO: check the write DFC as well
 	return false, statusReason
 }
```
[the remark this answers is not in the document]
FalconTube: Same as Claude comment above, will be fixed

umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -449,21 +470,21 @@ func (a *EditProtocolConverterAction) applyMutation(benthosConfig dataflowcompon
...
-	case DFCTypeEmpty:
-		// For empty dfcType, we only update connection, location, and name - no DFC configuration
-		// The connection, location, and name updates are already handled above
-	default:
-		return config.ProtocolConverterConfig{}, uuid.Nil, fmt.Errorf("invalid DFC type: %s", a.dfcType.String())
-	}
-
-	instanceToModify.DesiredFSMState = a.state
+	// Store per-DFC desired state overrides in the spec so the FSM can propagate them.
+	instanceToModify.ProtocolConverterServiceConfig.ReadDFCDesiredState = a.readDFCState
```
JeremyTheocharis: **Klaus flagged** — `applyMutation()` overwrites `ReadDFCDesiredState` / `WriteDFCDesiredState` on the spec before passing them as the "existing" fallback args to `derivePCDesiredState`. The 3rd and 4th arguments always equal the 1st and 2nd — the fallback is structurally dead.

**Example: user changes the bridge IP while read flow is stopped**

Before this PR, the bridge state was a single field. Editing the IP didn't touch it — the stopped state was preserved.

After this PR:
1. Read flow is stopped (`ReadDFCDesiredState = "stopped"` in the spec)
2. User edits connection IP (a `DFCTypeEmpty` edit — no DFC fields in the payload)
3. `applyMutation` sets `ReadDFCDesiredState = ""` (line ~486, because the edit payload has no state)
4. Then calls `derivePCDesiredState("", "", "", "")` — the "existing" args are already overwritten
5. Returns `"active"` — the stopped read flow silently restarts

The code comment says "fall back to the existing spec values via derivePCDesiredState" — the implementation doesn't achieve this intent. Same issue applies to single-DFC edits (e.g., `DFCTypeRead` clears the write DFC's stored state).

Suggested fix — only overwrite when the action actually carries a state:
```go
origRead  := instanceToModify.ProtocolConverterServiceConfig.ReadDFCDesiredState
origWrite := instanceToModify.ProtocolConverterServiceConfig.WriteDFCDesiredState

if a.readDFCState != "" {
    instanceToModify.ProtocolConverterServiceConfig.ReadDFCDesiredState = a.readDFCState
}
if a.writeDFCState != "" {
    instanceToModify.ProtocolConverterServiceConfig.WriteDFCDesiredState = a.writeDFCState
}

instanceToModify.DesiredFSMState = derivePCDesiredState(
    a.readDFCState, a.writeDFCState, origRead, origWrite,
)
```

The conditional overwrites preserve stored per-DFC states across `DFCTypeEmpty` and single-DFC edits. Without them, the spec is clobbered to `""` and the next FSM tick reads the empty value, restarting the DFC.

_4 reviewers + 8 code quality agents · DA confirmed as P1_

JeremyTheocharis (commented):
### How I reviewed this

I started by reading the diff on GitHub. The FSMv1 protocol converter code is hard to follow from the diff alone -- the business logic (when does a bridge degrade? what happens if only one flow is stopped? does the connection stay alive?) is not documented in the code. So I traced it myself, scenario by scenario.

That's where most of the findings below come from. I identified spots where I needed a comment to understand what's happening (the `nit: suggestion` comments). Writing those comments forced me to check whether the code actually does what I assumed. It doesn't always.

Three categories of issues:

**1. Business logic gaps** (Comments 10, 11, 12): The per-DFC state split is the right idea, but the health check functions (`IsOtherDegraded`, `IsDataflowComponentWithProcessingActivity`, `IsProtocolConverterStopped`) still only look at the read flow. Stopping a read flow while the write flow runs causes false degradation, wrong idle state, and orphaned S6 processes. Also, `derivePCDesiredState` ties the bridge lifecycle to the DFC lifecycle -- stopping both flows tears down the connection, which shouldn't happen. These aren't edge cases: they're the core scenarios that per-DFC state enables.

**2. UNS enforcement** (Comment 8): Read DFC output is always overwritten with UNS (enforced). This PR adds a guard that makes write DFC input conditional -- a user-provided input can bypass UNS. The read and write side should be symmetric. UNS is the bus; both sides should enforce it.

**3. Code quality** (Comments 1-7, 9a-9d): I also ran Klaus on this (automated multi-agent review + cross-check pass). Those findings are below. I haven't traced every Klaus finding personally -- treat them as flagged for your attention, not as required changes. The `applyMutation` fallback bug (Comment 1) is the most important one: editing a bridge's IP silently restarts a stopped flow.

One more thing: the business logic for this code is not documented anywhere. I had to trace through 5 files to answer "what happens when I stop one flow?" That should be a one-line comment, not a 30-minute investigation. The `nit: suggestion` comments are a start, but the overall flow (deploy → start connection → start DFCs → health checks → stop) should be documented somewhere, even if just as a comment block in `actions.go`.

| Category | Count |
|----------|-------|
| Business logic / design | 3 |
| UNS enforcement | 1 |
| Klaus flagged (should fix) | 5 |
| Klaus flagged (MEDIUM) | 2 |
| nit (code comment suggestions) | 3 |

---

**Comment 5 — Missing CHANGELOG entry**

Klaus flagged — The `skip-changelog-guard` label is applied, but this enables a feature that future MC versions will expose. Suggested entry under `## Unreleased` → `### Improvements`:

> Starting with this version, read and write flows within a bridge can be deployed and controlled independently. Previously, a bridge always ran both directions as a single unit. A Management Console update is required to expose this in the UI

_1 code quality agent · DA confirmed as P1_

---

**Comment 10 — Health functions only check read flow**

Three functions have `// TODO: check the write DFC as well` comments that become active bugs when per-DFC stop is used:

- `IsOtherDegraded` (line 517): only checks read flow metrics. Stopping the read flow zeroes those metrics, triggering "Redpanda active but no output." The bridge shows degraded even though the write flow is healthy.
- `IsDataflowComponentWithProcessingActivity` (line 554): only checks read flow. When only the write flow is active, the bridge shows idle.
- `IsProtocolConverterStopped` (line 575): only checks read flow and connection. It declares the bridge stopped before the write benthos process has shut down, leaving an orphaned S6 process.

These will become active bugs when the per-DFC stop UI ships and must be fixed before that. Consider addressing them in this PR since it introduces the per-DFC state plumbing.

---

**Comment 11 — `derivePCDesiredState` should not tie bridge lifecycle to DFC lifecycle**

The PC desired state should not be derived from per-DFC states. Add a bridge-level desired state field that the user sets directly, separate from the per-DFC fields.

The current design has a concrete consequence: stopping both DFCs also tears down the connection monitor. A user who stops both flows loses visibility into whether the device is reachable. The bridge connection should stay alive until the user explicitly stops the bridge, not as a side effect of stopping both flows.

---

**Comment 12 — Per-flow failure visibility: how does the user know which flow failed?**

With both read and write flows deployed, if the read flow fails to start but the write flow is healthy: what does the user see?

Today there's one state badge per bridge. `starting_failed_dfc` doesn't distinguish read vs write. The user sees "failed" but can't tell which flow caused it -- or that the other flow is actually fine.

The status reason field could carry "read flow failed: config error" vs "write flow failed: connection timeout" without a UI redesign. But it needs to be implemented -- today the status reason doesn't include per-flow detail.

Worth solving before the MC adds write flow deployment.

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -793,6 +801,9 @@ func (p *ProtocolConverterService) EvaluateDFCDesiredStates(protConvName string,
 		if config.Name == dfcWriteName {
 			if protocolConverterDesiredState == "stopped" { // NOTE: Hardcoded to avoid circular import with pkg/fsm/protocolconverter
 				p.dataflowComponentConfig[i].DesiredFSMState = dfcfsm.OperationalStateStopped
+			} else if writeDFCDesiredState == "stopped" {
+				// Per-DFC override: user explicitly wants write DFC stopped
+				p.dataflowComponentConfig[i].DesiredFSMState = dfcfsm.OperationalStateStopped
 			} else {
 				// Only start the DFC if it has been configured AND connection is confirmed up
 				if len(p.dataflowComponentConfig[i].DataFlowComponentServiceConfig.BenthosConfig.Input) > 0 {
```
JeremyTheocharis: **Klaus flagged** — Write DFC "has config" check uses `.BenthosConfig.Input` but write DFC user config lives in `.BenthosConfig.Output`. The Input is auto-generated at runtime by `GetDFCWriteServiceConfig()`.

Not permanently broken — write DFCs start after one reconciliation cycle (~1s) once the runtime injection populates Input. But semantically wrong and fragile: a refactor that changes injection order would silently break write DFC startup.

Note: the mock at `mock.go` already uses `.Output` (correct), but the production code uses `.Input` — a pre-existing divergence exposed by this PR.

Suggested fix: Change `.BenthosConfig.Input` to `.BenthosConfig.Output` for the write DFC check.

_6/8 code quality agents + service reviewer · DA confirmed as P1 (demoted from P0 — not permanent, but fragile)_

umh-core/pkg/communicator/actions/get-protocolconverter.go
```diff
@@ -311,6 +312,10 @@ func (a *GetProtocolConverterAction) Execute() (interface{}, map[string]interfac
 							fmt.Sprintf("Warning: Failed to build read DFC for protocol converter '%s': %v", instance.ID, err),
 							a.outboundChannel, models.GetProtocolConverter)
 					}
+
+					if readDFC != nil {
+						readDFC.State = specConfig.ReadDFCDesiredState
```
JeremyTheocharis, line 317: **Klaus flagged** — Two issues in the Get handler for write DFCs:

1. `buildProtocolConverterDFCFromConfig` (line ~181): `if len(dfcConfig.BenthosConfig.Input) == 0 { return nil, nil }` — always true for write DFCs (Input is auto-generated at runtime, not stored in spec). Write DFCs are invisible in the Get response.

2. The `len(writeDFCConfig.BenthosConfig.Input) > 0` check further down has the same bug. Write DFC user config lives in `.BenthosConfig.Output`. Use `.Output` there.

The function needs either a `dfcType` parameter or separate read/write variants, since the "has config" signal is Input for read and Output for write.

_1 reviewer · DA confirmed as P1, amplified by MI-3 (helper function needs the same fix)_

umh-core/pkg/communicator/actions/edit-protocolconverter_test.go
```diff
@@ -232,40 +232,44 @@ var _ = Describe("EditProtocolConverter", func() {
 			Expect(err.Error()).To(ContainSubstring("failed to parse protocol converter payload"))
 		})
 
-		It("should default state to 'active' when not provided", func() {
+		It("should default DFC type to empty when no DFCs provided", func() {
```
JeremyTheocharis, line 235: **Klaus flagged** — The PR adds several new functions (`derivePCDesiredState`, `compareSingleDFCConfig`, `DFCTypeBoth` paths, per-DFC state handling) but zero new test cases. The existing test changes only rename tests and update assertions from `GetState()` to `GetDFCType()`.

At minimum, add:
- A truth table test for `derivePCDesiredState` (pure function, no setup needed)
- A test for `DFCTypeEmpty` edit on a bridge where `ReadDFCDesiredState` is "stopped": the stopped state should be preserved. This test currently fails -- it would have caught the fallback clobber in `applyMutation`.
- A test for write DFC state after a read-only DFC edit

_6/8 code quality agents + service reviewer · DA promoted to P1 (would have caught Comment 1)_

umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -768,8 +789,14 @@ func (a *EditProtocolConverterAction) awaitRollout(pcConfig config.ProtocolConve
 					currentStateReason = pcSnapshot.ServiceInfo.StatusReason
 				}
 
-				// send the benthos logs to the user
-				logs = pcSnapshot.ServiceInfo.DataflowComponentReadObservedState.ServiceInfo.BenthosObservedState.ServiceInfo.BenthosStatus.BenthosLogs
+				// send the benthos logs to the user; for DFCTypeBoth, prefer write logs since
```
JeremyTheocharis, line 791: **Klaus flagged (MEDIUM)** — For `DFCTypeBoth`, logs are set to write DFC only. The comment here claims "read logs are also included below via the merged slice" but no merge code exists. Read DFC config errors would cause a 60-second timeout instead of immediate rollback detection.

Not triggerable from the current frontend (always sends readDFC only), but worth fixing before DFCTypeBoth is used.

_1 reviewer + 5/8 code quality agents · DA confirmed as P2_

umh-core/pkg/service/protocolconverter/runtime_config/runtime_config.go
```diff
@@ -194,6 +194,7 @@ func BuildRuntimeConfig(
 	}
 
 	vb.Internal["bridged_by"] = config.GenerateBridgedBy(config.ComponentTypeProtocolConverter, nodeName, pcName)
+	vb.Internal["umh_topic"] = "umh.v1." + locationPath + ".*"
```
JeremyTheocharis: **Klaus flagged (MEDIUM)** — `"umh.v1." + "" + ".*"` produces `"umh.v1..*"` when locationPath is empty. Quick guard:

```go
if locationPath != "" {
    vb.Internal["umh_topic"] = "umh.v1." + locationPath + ".*"
} else {
    vb.Internal["umh_topic"] = "umh.v1.*"
}
```

_1 reviewer + 5/8 code quality agents + coderabbitai · DA confirmed as P2_

umh-core/pkg/config/protocolconverterserviceconfig/config_customized.go
```diff
@@ -50,8 +50,8 @@ func (c ProtocolConverterServiceConfigSpec) GetDFCReadServiceConfig() dataflowco
 func (c ProtocolConverterServiceConfigSpec) GetDFCWriteServiceConfig() dataflowcomponentserviceconfig.DataflowComponentServiceConfig {
 	dfcWriteConfig := c.Config.DataflowComponentWriteServiceConfig
 
-	// Only append UNS input if there's an output config
-	if len(dfcWriteConfig.BenthosConfig.Output) > 0 {
+	// Only append UNS input if there's an output config and no input has been set by the user
+	if len(dfcWriteConfig.BenthosConfig.Output) > 0 && len(dfcWriteConfig.BenthosConfig.Input) == 0 {
```
JeremyTheocharis, line 54: The `&& len(dfcWriteConfig.BenthosConfig.Input) == 0` guard added here breaks UNS enforcement on the write side. Read DFC output is always overwritten with UNS (line ~38, unconditional). Write DFC input should follow the same pattern. UNS is the bus between read and write flows, and both sides should enforce it.

The old behavior (unconditional UNS input injection) was correct. This guard introduces a way for a write DFC to bypass UNS, which should not be possible.

Suggested fix: remove the `&& len(dfcWriteConfig.BenthosConfig.Input) == 0` condition, matching the read DFC pattern.

umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -987,7 +1076,55 @@ func (a *EditProtocolConverterAction) GetDFCType() string {
...
+		return fmt.Errorf("invalid %s DFC configuration: %w", label, err)
+	}
+	return nil
+}
+
+// derivePCDesiredState computes the protocol converter desired state from per-DFC states.
+// newRead/newWrite are the states from the current action (may be empty).
+// existingRead/existingWrite are the states already stored in the spec (fallback).
+// The PC is stopped only when both effective DFC states are "stopped".
+func derivePCDesiredState(newRead, newWrite, existingRead, existingWrite string) string {
```
JeremyTheocharis: nit: Reading just the diff, it is not obvious why this function exists or what rule it encodes.

```suggestion
// derivePCDesiredState returns "stopped" only when both DFCs are explicitly
// stopped. A single running DFC keeps the bridge and its connection alive.
func derivePCDesiredState(newRead, newWrite, existingRead, existingWrite string) string {
```

umh-core/pkg/service/protocolconverter/protocolconverter.go
```diff
@@ -753,10 +756,20 @@ func (p *ProtocolConverterService) EvaluateDFCDesiredStates(protConvName string,
...
+		currentFSMState == "active" ||
+		currentFSMState == "degraded_redpanda" ||
+		currentFSMState == "degraded_dfc" ||
+		currentFSMState == "degraded_other"
+
 	for i, config := range p.dataflowComponentConfig {
 		if config.Name == dfcReadName {
 			if protocolConverterDesiredState == "stopped" { // NOTE: Hardcoded to avoid circular import with pkg/fsm/protocolconverter
 				p.dataflowComponentConfig[i].DesiredFSMState = dfcfsm.OperationalStateStopped
+			} else if readDFCDesiredState == "stopped" {
```
JeremyTheocharis: nit: The ordering here matters but is not obvious from the diff.

```suggestion
			} else if readDFCDesiredState == "stopped" { // NOTE: Hardcoded to avoid circular import
				// Per-DFC overrides take priority: a DFC marked "stopped" stays stopped
				// regardless of valid config or connection state.
```

umh-core/pkg/fsm/protocolconverter/actions.go
```diff
@@ -435,24 +437,49 @@ func (p *ProtocolConverterInstance) IsRedpandaHealthy() (bool, string) {
...
+// IsDFCHealthy checks whether the underlying DFCs are healthy.
+// A DFC is considered healthy when it is active/idle, intentionally stopped
+// (ReadDFCDesiredState/WriteDFCDesiredState == "stopped"), or has no configuration
+// (nothing deployed means nothing to check).
 //
 // It returns:
 //
-//	ok     – true when the DFC is healthy, false otherwise.
-//	reason – empty when ok is true; otherwise a explanation
+//	ok     – true when all relevant DFCs are healthy, false otherwise.
```
JeremyTheocharis, line 447: nit: "Stopped = healthy" is counterintuitive without context.

```suggestion
	// Intentionally stopped DFCs count as healthy. Without this,
	// stopping one DFC would degrade the entire bridge.
	readIntentionallyStopped := p.specConfig.ReadDFCDesiredState == OperationalStateStopped
```

umh-core/pkg/communicator/actions/edit-protocolconverter.go
FalconTube: Done

umh-core/pkg/service/protocolconverter/protocolconverter.go
FalconTube: Left comment to make it more clear.
Although probably renaming the variable itself would also be an option

umh-core/pkg/service/protocolconverter/protocolconverter.go
FalconTube: Fixed

umh-core/pkg/service/protocolconverter/runtime_config/runtime_config.go
FalconTube: Fixed

umh-core/pkg/fsm/protocolconverter/actions.go
FalconTube: Done

umh-core/pkg/config/protocolconverterserviceconfig/config_customized.go
FalconTube: Will do this in a separate ticket, because the current "override" is required for testing.
@Linear create a ticket for this as a subticket for ENG-4097

umh-core/pkg/communicator/actions/get-protocolconverter.go
FalconTube: Will also put this to the separate subticket, because it is required for testing now.
@Linear add this to the new subticket of ENG-4097, which came from another comment

umh-core/pkg/communicator/actions/edit-protocolconverter_test.go
FalconTube: I want to tackle the tests, once this feature is more mature, because I think otherwise I will have to rewrite tests quite a lot.
@Linear create a subticket of ENG-4097 for testing all the new stuff from ENG-4097 itself

umh-core/pkg/communicator/actions/edit-protocolconverter.go
FalconTube: I think the last line of the comment says exactly that, or am I confusing something?

umh-core/pkg/communicator/actions/edit-protocolconverter.go
FalconTube: Will address that in a separate ticket, cause it needs some thinking on how to solve this in the frontend as well.
@Linear create a subticket for this issue below ENG-4097

FalconTube (commented):
Comment 5 -> Since this PR is not a properly finished state, I don't really want to add this to the changelog. Would make more sense to only add this, once it is clean imho.
Comment 10 and 11 -> Will solve it in this PR
Comment 12 -> Has to go in follow up ticket, cause it needs frontend decisions

About "tracing the action with an inline comment" -> While comments can be quite helpful, they also tend to lie. Ideally we should use the ginkgo tests to follow the behavior of the code, because tests don't lie. But I did not yet add the tests in this PR, as described in a comment above.

umh-core/pkg/communicator/actions/edit-protocolconverter.go
FalconTube: is ticket ENG-4686

umh-core/pkg/communicator/actions/edit-protocolconverter_test.go
FalconTube: Is ticket ENG-4685

umh-core/pkg/communicator/actions/get-protocolconverter.go
FalconTube: Is ticket ENG-4687

umh-core/pkg/config/protocolconverterserviceconfig/config_customized.go
FalconTube: Is ticket ENG-4687

umh-core/pkg/service/protocolconverter/runtime_config/runtime_config.go
```diff
@@ -194,6 +194,12 @@ func BuildRuntimeConfig(
 	}
 
 	vb.Internal["bridged_by"] = config.GenerateBridgedBy(config.ComponentTypeProtocolConverter, nodeName, pcName)
+	if locationPath == "" {
+		vb.Internal["umh_topic"] = "umh.v1.*"
+	} else {
+		vb.Internal["umh_topic"] = "umh.v1." + locationPath + ".*"
+
```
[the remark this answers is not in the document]
FalconTube: Won't do here, because will be removed in the near future anyways, when we generate the topics ourselfes, so there can not be a bad inject.

FalconTube:
Comments 10 and 11 are now adressed via the last two commits as well.

JeremyTheocharis:
PR LGTM. Went through all my comments against the current code, the 7 fixed items check out (including the last two commits for dual-flow health checks and bridge lifecycle decoupling).

Three items deferred to tickets:
- ENG-4685 - Tests for per-DFC state handling
- ENG-4686 - DFCTypeBoth log merging
- ENG-4687 - UNS enforcement + Get handler fix

One ask: please restructure ENG-4097 like ENG-4690 (rollout-style parent). The description is empty and the three deferred tickets have no blocking relationships, so they could get lost. Concretely:
1. Add a rollout checklist to ENG-4097 (what before FF, what's "done")
2. Mark ENG-4685/4686/4687 as blocking rollout

Doesn't block merge. Just want the remaining work tracked so nothing slips.

FalconTube:
I now updated the main rollout ticket ENG-4097.
I did **not** mark the other tickets as blocking for rollout, but rather I made them subtickets. So all tasks in there should be finished before rollout, which to me is the same (but easier to track) as using the standlone + blocker tickets.

JeremyTheocharis (approved):
Rollout ticket ENG-4097 restructured with all 3 deferred items (ENG-4685, ENG-4686, ENG-4687) as children — verified. Getting-to-production checklist and FF-enable criteria look good.

LGTM, ship it behind the flag.

PR #2497, opened 2026-03-25
feat(models): add FeatureUsage to status messages (ENG-4580)

JeremyTheocharis:
Adds a `FeatureUsage` struct to the status message so the Management Console can report which feature flags are active on each instance.

The struct carries one boolean per feature flag (config backup, FSMv2 transport/memory/protocol, resource limiting) plus a backup count that increments on each successful config backup write. Bool fields all end in `Enabled` so the frontend can auto-map them to PostHog properties without maintaining a manual list.

Built in `cmd/main.go` from env vars and YAML config, threaded through the status collector, included in every status push. Old instances that don't have this field just omit it (pointer + omitempty).

## How to verify

- `go test ./pkg/models/...` -- omitempty contract test
- `go test ./...` -- full suite
- `make betteralign-fix && git diff` -- no alignment changes

### DIFF
--- a/umh-core/cmd/main.go
+++ b/umh-core/cmd/main.go
@@ -139,18 +139,12 @@ func main() {
 
 	configData.Agent.UseFSMv2ProtocolConverter = v
 
-	now := time.Now()
-	var configBackupEnabledSince *time.Time
-	if configBackupEnabled {
-		configBackupEnabledSince = &now
-	}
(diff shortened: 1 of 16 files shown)


JeremyTheocharis:
@coderabbitai review

JeremyTheocharis:
@claude review

rpaland (approved):
LGTM

PR #2508, opened 2026-04-07
fix(container-monitor): guard throttle state update on cgroup reads

led0nk:
# Description

`wasThrottled` was reset to `false` on cgroup read failures, causing a "CPU throttling detected" warning when reading recovered while throttling was still active.

This PR guards the `wasThrottled` update behind `cgroupErr == nil`, so the throttle state is only updated when we have a valid reading.

Fixes ENG-4735


JeremyTheocharis (approved):
Guarding `wasThrottled` behind `cgroupErr == nil` fixes the spurious throttle warnings on transient cgroup read failures. Switch refactor is a nice cleanup too. LGTM.

PR #2509, opened 2026-04-07
Eng 4756 separate start and stop for read and write flows

FalconTube:
This PR does two things
- Only find the relevant DFC type (read or write) to be acted upon, when an `edit` action comes in. It compares the sent payload for both read and write to the `config.yaml` and checks, if an action is necessary. It then propagates this info via `DFCWrite`, `DFCRead`, `DFCBoth` or `DFCEmpty`
- Before editing/re-redploying a DFC (read or write), it first stops the the running flow. Afterwards it tries to start it as as usual, but it now correctly only reads inspects the logs from the the most recent deployment. This fixes the problem, where stale errors from a previous deployment would contaminate the new redeployment.

### DIFF
--- a/umh-core/pkg/communicator/actions/edit-protocolconverter.go
+++ b/umh-core/pkg/communicator/actions/edit-protocolconverter.go
@@ -155,41 +155,29 @@ func (a *EditProtocolConverterAction) Parse(payload interface{}) error {
 	a.protocolConverterUUID = *pcPayload.UUID
 	a.name = pcPayload.Name
 
-	// Determine which DFC(s) are being updated and convert to CDFCPayload.
-	switch {
-	case pcPayload.ReadDFC != nil && pcPayload.WriteDFC != nil:
-		a.dfcType = DFCTypeBoth
+	// Parse each DFC side independently.
+	if pcPayload.ReadDFC != nil {
 		readPayload := dfcToPayload(pcPayload.ReadDFC)
-		writePayload := dfcToPayload(pcPayload.WriteDFC)
 		a.readDFCPayload = &readPayload
-		a.writeDFCPayload = &writePayload
+		a.readDFCState = pcPayload.ReadDFC.State
 		if pcPayload.ReadDFC.IgnoreErrors != nil {
 			a.ignoreHealthCheck = *pcPayload.ReadDFC.IgnoreErrors
 		}
+	}
+	if pcPayload.WriteDFC != nil {
+		writePayload := dfcToPayload(pcPayload.WriteDFC)
+		a.writeDFCPayload = &writePayload
+		a.writeDFCState = pcPayload.WriteDFC.State
 		if pcPayload.WriteDFC.IgnoreErrors != nil {
 			a.ignoreHealthCheck = a.ignoreHealthCheck || *pcPayload.WriteDFC.IgnoreErrors
 		}
-		a.readDFCState = pcPayload.ReadDFC.State
-		a.writeDFCState = pcPayload.WriteDFC.State
-	case pcPayload.ReadDFC != nil:
-		a.dfcType = DFCTypeRead
-		readPayload := dfcToPayload(pcPayload.ReadDFC)
-		a.readDFCPayload = &readPayload
-		if pcPayload.ReadDFC.IgnoreErrors != nil {
-			a.ignoreHealthCheck = *pcPayload.ReadDFC.IgnoreErrors
-		}
-		a.readDFCState = pcPayload.ReadDFC.State
-	case pcPayload.WriteDFC != nil:
-		a.dfcType = DFCTypeWrite
-		writePayload := dfcToPayload(pcPayload.WriteDFC)
-		a.writeDFCPayload = &writePayload
-		if pcPayload.WriteDFC.IgnoreErrors != nil {
-			a.ignoreHealthCheck = *pcPayload.WriteDFC.IgnoreErrors
-		}
-		a.writeDFCState = pcPayload.WriteDFC.State
-	default:
-		a.dfcType = DFCTypeEmpty
 	}
+
+	// Determine dfcType by comparing incoming DFC configs against what is
+	// currently deployed. Only DFCs that actually differ need to be stopped
+	// and restarted.
+	a.dfcType = a.deriveDFCType()
+
 	if pcPayload.TemplateInfo != nil {
 		a.vb = pcPayload.TemplateInfo.Variables
 	} else {
@@ -1157,6 +1145,97 @@ func extractInjectFromRawYAML(rawYAML *models.CommonDataFlowComponentRawYamlConf
 	}
 
 	return rawYAML.Data
+}
+
+// deriveDFCType compares the incoming DFC payloads against the currently
+// deployed config and returns a DFCType that only includes the sides that
+// actually differ. If no config is deployed yet (first edit) or the config
+// cannot be read, it falls back to payload presence.
+func (a *EditProtocolConverterAction) deriveDFCType() DFCType {
+	hasRead := a.readDFCPayload != nil
+	hasWrite := a.writeDFCPayload != nil
+
+	if !hasRead && !hasWrite {
+		return DFCTypeEmpty
+	}
+
+	// Try to fetch the current config for diff comparison.
+	ctx, cancel := context.WithTimeout(context.Background(), constants.ActionTimeout)
+	defer cancel()
+
+	currentConfig, err := a.configManager.GetConfig(ctx, 0)
+	if err != nil {
+		a.actionLogger.Debugf("Cannot read current config for diff, falling back to payload presence: %v", err)
+		return dfcTypeFromPresence(hasRead, hasWrite)
+	}
+
+	// Find the protocol converter in the current config.
+	var currentPC *config.ProtocolConverterConfig
+	for i, pc := range currentConfig.ProtocolConverter {
+		if dataflowcomponentserviceconfig.GenerateUUIDFromName(pc.Name) == a.protocolConverterUUID {
+			currentPC = &currentConfig.ProtocolConverter[i]
+			break
+		}
+	}
+	if currentPC == nil {
+		a.actionLogger.Debugf("Protocol converter %s not found in current config, falling back to payload presence", a.protocolConverterUUID)
+		return dfcTypeFromPresence(hasRead, hasWrite)
+	}
+
+	readChanged := hasRead && a.dfcPayloadDiffers(*a.readDFCPayload, a.readDFCState,
+		currentPC.ProtocolConverterServiceConfig.Config.DataflowComponentReadServiceConfig,
+		currentPC.ProtocolConverterServiceConfig.ReadDFCDesiredState)
+	writeChanged := hasWrite && a.dfcPayloadDiffers(*a.writeDFCPayload, a.writeDFCState,
+		currentPC.ProtocolConverterServiceConfig.Config.DataflowComponentWriteServiceConfig,
+		currentPC.ProtocolConverterServiceConfig.WriteDFCDesiredState)
+
+	derived := dfcTypeFromPresence(readChanged, writeChanged)
+	a.actionLogger.Debugf("Derived dfcType=%s (readChanged=%v, writeChanged=%v)", derived, readChanged, writeChanged)
+
+	return derived
+}
+
+// dfcPayloadDiffers checks whether an incoming DFC payload differs from the
+// persisted config in config.yaml. This is intentionally simpler than
+// compareSingleDFCConfig, which compares rendered templates against runtime
+// observed state from the FSM snapshot. Here we only need to know whether the
+// user sent something new relative to what is already on disk.
+func (a *EditProtocolConverterAction) dfcPayloadDiffers(
+	payload models.CDFCPayload,
+	incomingState string,
+	deployedConfig dataflowcomponentserviceconfig.DataflowComponentServiceConfig,
+	deployedState string,
+) bool {
+	// State change counts as a diff.
+	if incomingState != "" && incomingState != deployedState {
+		return true
+	}
+
+	incomingBenthos, err := CreateBenthosConfigFromCDFCPayload(payload, a.name)
+	if err != nil {
+		a.actionLogger.Debugf("Cannot build benthos config for diff, treating as changed: %v", err)
+		return true
+	}
+
+	incomingDFCConfig := dataflowcomponentserviceconfig.DataflowComponentServiceConfig{
+		BenthosConfig: incomingBenthos,
+	}
+
+	return !deployedConfig.Equal(incomingDFCConfig)
+}
+
+// dfcTypeFromPresence returns a DFCType based on boolean flags for read/write.
+func dfcTypeFromPresence(read, write bool) DFCType {
+	switch {
+	case read && write:
+		return DFCTypeBoth
+	case read:
+		return DFCTypeRead
+	case write:
+		return DFCTypeWrite
+	default:
+		return DFCTypeEmpty
+	}
 }
 
 // getUserEmail implements the Action interface by returning the user email associated with this action.
--- a/umh-core/pkg/communicator/actions/edit-protocolconverter.go
+++ b/umh-core/pkg/communicator/actions/edit-protocolconverter.go
@@ -47,6 +47,7 @@ import (
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/config/protocolconverterserviceconfig"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/constants"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsm"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsm/dataflowcomponent"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsm/protocolconverter"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/logger"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/models"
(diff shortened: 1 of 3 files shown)


umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -747,8 +849,10 @@ func (a *EditProtocolConverterAction) awaitRollout(pcConfig config.ProtocolConve
 				// For "stopped" state: accept only "stopped"
 				hasReachedDesiredState := false
 
+				a.actionLogger.Errorf("desiredPCState empty: %s, currentState empty: %s", desiredPCState, instance.CurrentState)
```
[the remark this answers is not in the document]
FalconTube: Will fix

umh-core/pkg/fsm/benthos/actions.go
```diff
@@ -744,6 +744,16 @@ func (b *BenthosInstance) IsBenthosRunningForSomeTimeWithoutErrors(currentTime t
...
 func (b *BenthosInstance) IsBenthosLogsFine(currentTime time.Time, logWindow time.Duration) (bool, string) {
+	// Cap the log window to the current process uptime so that errors from a
+	// previous process instance (before a stop/start or config-change restart)
+	// are ignored. When uptime is 0 (process not yet running or just started),
+	// the window becomes 0 and no historical logs are considered.
+	currentUptime := b.ObservedState.ServiceInfo.S6ObservedState.ServiceInfo.Uptime
+	uptimeDuration := time.Duration(currentUptime) * time.Second
+	if uptimeDuration < logWindow {
+		logWindow = uptimeDuration
+	}
```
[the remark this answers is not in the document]
FalconTube: Will add changelog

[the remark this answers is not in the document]
FalconTube: We intentionally cap the logs to the current uptime here in order to not see logs from a previously running instance.
Check the comment above the code

```
	// Cap the log window to the current process uptime so that errors from a
	// previous process instance (before a stop/start or config-change restart)
	// are ignored. When uptime is 0 (process not yet running or just started),
	// the window becomes 0 and no historical logs are considered.
```

umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -1041,6 +1146,97 @@ func extractInjectFromRawYAML(rawYAML *models.CommonDataFlowComponentRawYamlConf
...
+	if err != nil {
+		a.actionLogger.Debugf("Cannot build benthos config for diff, treating as changed: %v", err)
+		return true
+	}
+
+	incomingDFCConfig := dataflowcomponentserviceconfig.DataflowComponentServiceConfig{
+		BenthosConfig: incomingBenthos,
+	}
+
+	return !deployedConfig.Equal(incomingDFCConfig)
```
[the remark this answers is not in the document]
FalconTube: This is a false claim, because the user variables are stored elsewhere and are thus not used here for the comparison

JeremyTheocharis (commented):
I went through the code myself first, then ran the Klaus multi-agent review (8 agents + devil's advocate verified against the PR branch). My own findings are marked separately. For the Klaus findings, I haven't traced through all of them myself, so treat those as "Klaus flagged this" and assess for yourself.

umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -521,6 +534,89 @@ func (a *EditProtocolConverterAction) persistConfig(atomicEditUUID uuid.UUID, ne
 	return oldConfig, nil
 }
 
+// stopAndAwaitDFCs stops the affected DFCs before applying a new configuration.
+// This ensures a clean redeployment: any previously degraded DFC is stopped first,
+// then started fresh with the new config, avoiding stale errors blocking progress.
+// Returns the original (pre-edit) config for rollback purposes.
+func (a *EditProtocolConverterAction) stopAndAwaitDFCs(atomicEditUUID uuid.UUID, newSpec config.ProtocolConverterConfig) (config.ProtocolConverterConfig, error) {
```
JeremyTheocharis: This feels like a workaround. The uptime cap on `IsBenthosLogsFine` (benthos/actions.go:751) already solves the stale error problem: when s6 restarts with new config, uptime resets and old logs fall outside the window. Is there a specific scenario where the process does NOT restart on config change and the uptime cap alone isn't enough? If not, let's discuss so we can find a proper fix rather than adding this complexity.

**Klaus flagged** — `stopAndAwaitDFCs` adds a temporary intermediate config write, a polling loop with an arbitrary half-timeout heuristic, and on timeout continues anyway (defeating the purpose of stopping first). 5 agents flagged the timeout-returns-nil pattern, DA confirmed the architectural concern.

_Jeremy's review + 5 agents · DA confirmed_

umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -299,14 +288,43 @@ func (a *EditProtocolConverterAction) Execute() (interface{}, map[string]interfa
...
+				errorMsg, a.outboundChannel, models.EditProtocolConverter)
+
+			return nil, nil, fmt.Errorf("%s", errorMsg)
+		}
+
+		// Persist the actual new config (DFCs will start fresh from stopped state)
+		persistCtx, persistCancel := context.WithTimeout(context.Background(), constants.ActionTimeout)
+		defer persistCancel()
+
+		_, err = a.configManager.AtomicEditProtocolConverter(persistCtx, atomicEditUUID, newSpec)
```
JeremyTheocharis: **Klaus flagged** — If `stopAndAwaitDFCs` succeeds (persisting the stop config) but `AtomicEditProtocolConverter` on line 310 fails, the system is left with DFCs configured as stopped and no new config applied. `oldConfig` is available from the stop phase but unused for rollback.

Low probability (requires a disk I/O failure between the two persists), but the consequence is DFCs stuck stopped with no auto-recovery. Requires manual intervention.

Suggested fix: Add a rollback attempt using `oldConfig` in the error branch (lines 311-316). Three lines of code.

_2 agents · DA confirmed as MEDIUM_

umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -521,6 +534,89 @@ func (a *EditProtocolConverterAction) persistConfig(atomicEditUUID uuid.UUID, ne
...
+
+	ticker := time.NewTicker(constants.ActionTickerTime)
+	defer ticker.Stop()
+
+	// Use half the normal timeout for the stop phase, leaving enough time for the start phase.
+	timeout := time.After(constants.DataflowComponentWaitForActiveTimeout / 2)
+
+	for {
+		select {
+		case <-timeout:
```
JeremyTheocharis: **Klaus flagged (MEDIUM)** — When the 15s stop timeout fires (line 571), the function logs a `Warnf` and returns nil error. The caller proceeds to deploy on top of potentially-still-running DFCs. The `Warnf` goes to Sentry but the user gets no feedback.

DA notes: the uptime cap in `IsBenthosLogsFine` means stale errors are still excluded even if stop is incomplete, so this is not a correctness bug. The missing piece is user visibility.

Suggested fix: Add a `SendActionReply` with `ActionExecuting` status in the timeout case so the user knows the stop was incomplete. Also add PC UUID and DFC type to the warn log for structured context.

_5 agents · DA demoted from CRITICAL to MEDIUM_

umh-core/pkg/fsm/benthos/actions.go
```diff
@@ -744,6 +744,16 @@ func (b *BenthosInstance) IsBenthosRunningForSomeTimeWithoutErrors(currentTime t
 //	ok     – true when logs look clean, false otherwise.
 //	reason – empty when ok is true; otherwise the first offending log line.
 func (b *BenthosInstance) IsBenthosLogsFine(currentTime time.Time, logWindow time.Duration) (bool, string) {
+	// Cap the log window to the current process uptime so that errors from a
+	// previous process instance (before a stop/start or config-change restart)
+	// are ignored. When uptime is 0 (process not yet running or just started),
+	// the window becomes 0 and no historical logs are considered.
+	currentUptime := b.ObservedState.ServiceInfo.S6ObservedState.ServiceInfo.Uptime
```
JeremyTheocharis, line 751: **Klaus flagged (MEDIUM)** — The uptime cap (lines 751-754) is injected into `IsBenthosLogsFine`, which is a shared function. When uptime=0, `logWindow` becomes 0 and no logs are checked.

This is safe in practice because: (1) `IsBenthosDegraded` calls `IsBenthosConfigLoaded` first, which requires uptime >= 5s, so uptime=0 never reaches `IsBenthosLogsFine` in the degradation path. (2) `IsBenthosRunningForSomeTimeWithoutErrors` has its own uptime >= 10s guard.

The concern is semantic coupling: the correctness of this function depends on call-order guarantees in different functions. If someone calls `IsBenthosLogsFine` directly without the guard, they silently get "logs fine" for a down process.

Suggested fix: Consider moving the uptime cap into `IsBenthosRunningForSomeTimeWithoutErrors` (the startup caller where this fix matters) rather than into the shared `IsBenthosLogsFine`. Or add a guard comment documenting the caller-contract.

_5 agents · DA confirmed as MEDIUM (demoted from HIGH)_

umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -1046,6 +1146,108 @@ func extractInjectFromRawYAML(rawYAML *models.CommonDataFlowComponentRawYamlConf
 	return rawYAML.Data
 }
 
+// deriveDFCType compares the incoming DFC payloads against the currently
+// deployed config and returns a DFCType that only includes the sides that
+// actually differ. If no config is deployed yet (first edit) or the config
+// cannot be read, it falls back to payload presence.
+func (a *EditProtocolConverterAction) deriveDFCType() DFCType {
```
JeremyTheocharis, line 1038: **Klaus flagged** — `deriveDFCType` (line 1153) has 7+ code paths (no payloads, config read failure, PC not found, connection change, read-only, write-only, both, neither). `dfcPayloadDiffers` (line 1213) has 3 branches. Neither has test coverage. All existing Execute tests pass `nil` for `systemSnapshotManager`, so the stop-then-start path is also untested.

A regression in `deriveDFCType` (e.g., `Equal` false positive) means config changes are silently treated as no-ops. The user clicks "save" and nothing happens.

Suggested fix: Add table-driven unit tests for `deriveDFCType` covering at minimum: config read failure fallback, connection-only change, read-only change, write-only change, no-change case. Can defer to a follow-up ticket if needed for the VE release, but should be tracked.

_2 agents · DA confirmed as HIGH_

umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -154,41 +155,29 @@ func (a *EditProtocolConverterAction) Parse(payload interface{}) error {
...
 		}
-		a.writeDFCState = pcPayload.WriteDFC.State
-	default:
-		a.dfcType = DFCTypeEmpty
 	}
+
+	// Determine dfcType by comparing incoming DFC configs against what is
+	// currently deployed. Only DFCs that actually differ need to be stopped
+	// and restarted.
+	a.dfcType = a.deriveDFCType()
```
JeremyTheocharis, line 177: **Klaus flagged** — `deriveDFCType()` (called from Parse at line 179) reads `config.yaml` via `configManager.GetConfig`. Then `applyMutation` (called from Execute) reads it again independently. Two disk reads per edit action, plus a TOCTOU window where the config could change between Parse and Execute.

The fallback on config-read failure is safe (degrades to payload presence), so not a correctness bug. Parse was previously pure payload extraction.

Suggested fix: Move `deriveDFCType()` to the start of `Execute()`, or pass the config read from `applyMutation` into the derivation. Lower priority, can address in follow-up.

_3 agents · DA confirmed as MEDIUM_

umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -1046,6 +1146,108 @@ func extractInjectFromRawYAML(rawYAML *models.CommonDataFlowComponentRawYamlConf
...
+	}
+	if currentPC == nil {
+		a.actionLogger.Debugf("Protocol converter %s not found in current config, falling back to payload presence", a.protocolConverterUUID)
+		return dfcTypeFromPresence(hasRead, hasWrite)
+	}
+
+	// Connection IP/PORT changes affect all DFCs since they use {{ .IP }}/{{ .PORT }} templates.
+	// So if a change is detected, all present DFCs need redeploy.
+	deployedVars := currentPC.ProtocolConverterServiceConfig.Variables.User
+	connectionChanged := fmt.Sprint(deployedVars["IP"]) != a.connectionIP ||
```
JeremyTheocharis: **Klaus flagged** — The connection-change detection at line 1187 uses `fmt.Sprint(deployedVars["IP"])`. If the "IP" or "PORT" key is absent from the deployed config's `Variables.User` map (e.g., PCs created before these variables were standard), the map lookup returns nil. `fmt.Sprint(nil)` returns the string `"<nil>"`, not `""`. Verified: `"<nil>"` never equals any real IP or port, so `connectionChanged` is always true for such PCs.

This causes all present DFCs to be treated as "connection changed" and triggers stop-before-redeploy even when the connection didn't change. Safe direction (unnecessary restart rather than missed restart), but causes avoidable downtime.

Suggested fix: Use type assertion instead of `fmt.Sprint`:
```go
deployedIP, _ := deployedVars["IP"].(string)
deployedPort, _ := deployedVars["PORT"].(string)
```
A missing key asserts to `""` which correctly means "no previous value."

_1 agent · DA initially dismissed (verified as real bug via `go run`)_

umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -1046,6 +1146,108 @@ func extractInjectFromRawYAML(rawYAML *models.CommonDataFlowComponentRawYamlConf
...
+// observed state from the FSM snapshot. Here we only need to know whether the
+// user sent something new relative to what is already on disk.
+func (a *EditProtocolConverterAction) dfcPayloadDiffers(
+	payload models.CDFCPayload,
+	incomingState string,
+	deployedConfig dataflowcomponentserviceconfig.DataflowComponentServiceConfig,
+	deployedState string,
+) bool {
+	// State change counts as a diff.
+	if incomingState != "" && incomingState != deployedState {
```
JeremyTheocharis: **Klaus flagged** — In `dfcPayloadDiffers` (line 1220): `incomingState != "" && incomingState != deployedState`. `ReadDFCDesiredState`/`WriteDFCDesiredState` are new fields added by this PR. Existing on-disk configs don't have them, so they unmarshal to `""`. The incoming state defaults to `"active"`. Therefore `"active" != ""` is always true on the first edit after upgrading.

Every existing protocol converter edit on a freshly upgraded system triggers an unnecessary stop-before-redeploy cycle, causing a brief DFC interruption for all users on first use post-upgrade.

Suggested fix: Normalize empty deployed state to the default:
```go
if deployedState == "" {
    deployedState = dataflowcomponent.OperationalStateActive
}
```

_1 agent · DA did not review (late finding)_

umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -521,6 +534,89 @@ func (a *EditProtocolConverterAction) persistConfig(atomicEditUUID uuid.UUID, ne
...
+
+	SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionExecuting,
+		"Stopping affected data flow components for clean redeployment...",
+		a.outboundChannel, models.EditProtocolConverter)
+
+	ticker := time.NewTicker(constants.ActionTickerTime)
+	defer ticker.Stop()
+
+	// Use half the normal timeout for the stop phase, leaving enough time for the start phase.
+	timeout := time.After(constants.DataflowComponentWaitForActiveTimeout / 2)
```
JeremyTheocharis: **nit:** — The comment on the stop timeout says "leaving enough time for the start phase," implying a shared 30s budget split 50/50. In reality, `awaitRollout` creates its own independent 30s timeout (line 646). Worst-case total wall time is 45s, not 30s.

Suggested fix: Update the comment to reflect that the stop timeout is independent, not a budget split.

_3 agents · DA confirmed_

umh-core/pkg/communicator/actions/edit-protocolconverter.go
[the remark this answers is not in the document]
FalconTube: Fixed — the debug `Errorf` has been removed.

umh-core/pkg/communicator/actions/edit-protocolconverter.go
FalconTube: Resolved — `stopAndAwaitDFCs` has been removed entirely. The root cause (premature recovery from degraded state) is now fixed in `IsBenthosDegraded` by replacing the weak `IsBenthosConfigLoaded` (5s) + `IsBenthosLogsFine` checks with `IsBenthosRunningForSomeTimeWithoutErrors` (10s uptime + clean logs + clean metrics), matching the starting path's validation gate.

umh-core/pkg/communicator/actions/edit-protocolconverter.go
FalconTube: Resolved — `stopAndAwaitDFCs` has been removed. The stop-before-redeploy approach was replaced by fixing `IsBenthosDegraded` in the benthos FSM to use the same 10s validation gate as the starting path.

umh-core/pkg/fsm/benthos/actions.go
FalconTube: The semantic coupling concern is now resolved. `IsBenthosDegraded` was changed to call `IsBenthosRunningForSomeTimeWithoutErrors` (which has its own uptime >= 10s guard) instead of calling `IsBenthosConfigLoaded` + `IsBenthosLogsFine` separately. Both callers of `IsBenthosLogsFine` (`IsBenthosRunningForSomeTimeWithoutErrors` in the starting path, and now also in the degraded recovery path) have an uptime guard that runs before `IsBenthosLogsFine` is reached, so the zero-window edge case cannot occur in practice.

The uptime cap itself belongs in `IsBenthosLogsFine` — it's a property of the function ("only check logs from the current process instance"), not of its callers.

umh-core/pkg/communicator/actions/edit-protocolconverter.go
FalconTube: Agreed — `deriveDFCType` and `dfcPayloadDiffers` need test coverage. Will add tests in a follow-up.

umh-core/pkg/communicator/actions/edit-protocolconverter.go
FalconTube: Acknowledged. The TOCTOU window between Parse and Execute is minimal (only validation runs in between) and the fallback on config-read failure is safe. Moving `deriveDFCType` to Execute would break the existing test contract (`GetDFCType()` is asserted after Parse). Keeping as-is for now — the risk is low and fixing it properly requires restructuring how config is passed between Parse and Execute.

umh-core/pkg/communicator/actions/edit-protocolconverter.go
FalconTube: Fixed — now uses comma-ok map lookups. If IP/PORT keys are absent, the connection-change check is skipped entirely (rather than treating `"<nil>" != realValue` as changed).

umh-core/pkg/communicator/actions/edit-protocolconverter.go
FalconTube: Fixed — the state comparison now requires both sides to be non-empty: `incomingState != "" && deployedState != "" && incomingState != deployedState`. On freshly upgraded systems where the deployed state is `""`, the check is skipped and falls through to the config comparison.

umh-core/pkg/communicator/actions/edit-protocolconverter.go
FalconTube: Resolved — `stopAndAwaitDFCs` and its timeout have been removed entirely.

umh-core/pkg/communicator/actions/edit-protocolconverter.go
FalconTube: Removed the  `stopAndAwaitDFCs` function again and now check the health/status in `IsBenthosDegraded`

JeremyTheocharis (approved):
LGTM — the `IsBenthosDegraded` rewrite aligning degraded recovery with the starting path's 10s stability gate is the right fix. Clean removal of `stopAndAwaitDFCs`. The uptime cap in `IsBenthosLogsFine` is well-placed.

Follow-up ticket for clean log abstraction (`GetLogsSinceProcessStart`) coming separately.

JeremyTheocharis:
created https://linear.app/united-manufacturing-hub/issue/ENG-4781/clean-up-benthos-log-abstraction-structure-logs-by-process-lifetime so we know the pain point when we move it to fsmv2

PR #2523, opened 2026-04-17
feat(fsmv2): enable communicator transport and memory cleanup by default (ENG-4231)

JeremyTheocharis:
Closes Stage 3 of [ENG-4231](https://linear.app/united-manufacturing-hub/issue/ENG-4231/fsmv2-communicator-rollout). Stage 4 (4-week stability window) begins post-merge.

## What

Flip two env-var defaults in `cmd/main.go`:

- `USE_FSMV2_TRANSPORT`: `false` → `true`
- `USE_FSMV2_MEMORY_CLEANUP`: `false` → `true`

Plus the corresponding `pkg/fsmv2/README.md` table and one CHANGELOG entry.

## Why now

All Stage 2 gates were met:

- Sentry feature tags in place; no new blocking issues for 4+ weeks with the flag on
- v-prefix fix (#2511) holding; Sentry clean on v0.44.12+
- VE team ran FSMv2 at customer sites in the preview window; confirmed ready
- Rollout approved at product meeting (15.04)

## Why both flags flip together

The memory cleanup routine is paired with the transport. Flipping `USE_FSMV2_TRANSPORT` alone would re-expose the unbounded state growth that `USE_FSMV2_MEMORY_CLEANUP` fixes (13 GB RAM incident under ENG-4292) for anyone who hadn't already opted into both. v0.44.8's CHANGELOG told users to set them as a pair, so this PR keeps that pairing.

`USE_FSMV2_PROTOCOL_CONVERTER` stays default-false — that migration is a separate rollout outside ENG-4231's scope.

## Rollback

```bash
docker run -e USE_FSMV2_TRANSPORT=false -e USE_FSMV2_MEMORY_CLEANUP=false ...
```

No config.yaml fallback — the `useFSMv2Transport:` yaml path was deprecated in v0.44.8.

## Reviewers

- Janik (code)
- Yannic (additional; led the FSMv2 chaos testing)

### DIFF
--- a/umh-core/pkg/fsmv2/README.md
+++ b/umh-core/pkg/fsmv2/README.md
@@ -310,7 +310,6 @@ docker run -d \
   -e API_URL=https://management.umh.app \
   -e USE_FSMV2_TRANSPORT=true \
   -e USE_FSMV2_MEMORY_CLEANUP=true \
-  -e USE_FSMV2_PROTOCOL_CONVERTER=true \
   umh-core:latest
 ```
 
@@ -320,7 +319,6 @@ docker run -d \
 | `API_URL` | Yes | Backend relay server URL (e.g., `https://management.umh.app`) |
 | `USE_FSMV2_TRANSPORT` | No | Enables the FSMv2 communicator. Defaults to `true`. Set to `false` to fall back to the legacy communicator |
 | `USE_FSMV2_MEMORY_CLEANUP` | No | Enables the FSMv2 memory cleanup (persistence worker). Defaults to `true`. Set to `false` to disable |
-| `USE_FSMV2_PROTOCOL_CONVERTER` | No | Enables the FSMv2 protocol converter. Defaults to `false`. Set to `true` to enable |
 
 ### Disabling FSMv2 Features
 
@@ -329,8 +327,6 @@ docker run -d \
 ```bash
 -e USE_FSMV2_TRANSPORT=false
 ```
-
-`USE_FSMV2_PROTOCOL_CONVERTER` defaults to `false`; omit it or set it to `false` to keep the legacy behavior.
 
 ### Verifying FSMv2 Communicator
 
--- a/umh-core/pkg/fsmv2/README.md
+++ b/umh-core/pkg/fsmv2/README.md
@@ -318,17 +318,19 @@ docker run -d \
 |----------|----------|-------------|
 | `AUTH_TOKEN` | Yes | Authentication token from Management Console |
 | `API_URL` | Yes | Backend relay server URL (e.g., `https://management.umh.app`) |
-| `USE_FSMV2_TRANSPORT` | No | Set to `true` to enable FSMv2 communicator |
-| `USE_FSMV2_MEMORY_CLEANUP` | No | Set to `true` to enable FSMv2 memory cleanup (persistence worker) |
-| `USE_FSMV2_PROTOCOL_CONVERTER` | No | Set to `true` to enable FSMv2 protocol converter |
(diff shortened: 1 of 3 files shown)


JeremyTheocharis:
Self-review complete (13 agents + 3-group DA triage). Second commit [`87de6b9`](https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/2523/commits/87de6b99bb084adc556ee5204c96b9ca74d41dee) addresses the three Important findings:

**CHANGELOG** — folded the stale `### Preview: FSMv2 Communicator` bullet into `### Improvements` (after the flip, the preview heading contradicts the graduation story). Added one sentence warning stale-yaml opt-outs that `useFSMv2Transport: false` in `config.yaml` has had no effect since v0.44.8. Changed "also on by default" → "also enabled by default" for internal consistency.

**main.go** — dropped the dead-code `if err != nil` branches at all four `GetAsBool` call sites. `GetAsBool` with `required=false` never returns an error on parse failure, so the warn branches were unreachable. Post-flip this was a real concrete risk: `USE_FSMV2_TRANSPORT=disabled` (or any non-canonical string) would silently opt the operator **INTO** the new transport despite clear intent to roll back. Discarded with `_` and documented the contract. Full signature fix tracked in [ENG-4809](https://linear.app/united-manufacturing-hub/issue/ENG-4809).

**README** — harmonized the `USE_FSMV2_PROTOCOL_CONVERTER` row to match the shape of the TRANSPORT/MEMORY_CLEANUP rows.

Full triage: 0 Critical, 3 Important (all addressed), 6 Suggestions (most demoted, one acted on), 12 false positives filtered. Detail in `.self-review/20260417-171441/da-triage.md` (not pushed).

umh-core/pkg/fsmv2/README.md
```diff
@@ -318,18 +318,20 @@ docker run -d \
 |----------|----------|-------------|
 | `AUTH_TOKEN` | Yes | Authentication token from Management Console |
 | `API_URL` | Yes | Backend relay server URL (e.g., `https://management.umh.app`) |
-| `USE_FSMV2_TRANSPORT` | No | Set to `true` to enable FSMv2 communicator |
-| `USE_FSMV2_MEMORY_CLEANUP` | No | Set to `true` to enable FSMv2 memory cleanup (persistence worker) |
-| `USE_FSMV2_PROTOCOL_CONVERTER` | No | Set to `true` to enable FSMv2 protocol converter |
+| `USE_FSMV2_TRANSPORT` | No | Enables the FSMv2 communicator. Defaults to `true`. Set to `false` to fall back to the legacy communicator |
+| `USE_FSMV2_MEMORY_CLEANUP` | No | Enables the FSMv2 memory cleanup (persistence worker). Defaults to `true`. Set to `false` to disable |
+| `USE_FSMV2_PROTOCOL_CONVERTER` | No | Enables the FSMv2 protocol converter. Defaults to `false`. Set to `true` to enable |
```
led0nk: we don't have that right? should we then leave this out from documenation purposes?

led0nk (approved):

umh-core/pkg/fsmv2/README.md
JeremyTheocharis: yeah, it popped up a couple of times already. i need to check what the original purpose here was and why it landed up (if it was just an LLM hallucination or a legit placeholder)

JeremyTheocharis: Removed from the README. Left the env read + config field in place for yaml backwards-compat (existing configs with `useFSMv2ProtocolConverter: true` persisted won't fail to parse). The flag has no runtime effect — we'll re-introduce the docs when the protocol converter worker actually lands.

FalconTube (approved):
Good to go, but add handling for mutex lock in transport worker in follow up PR, as discussed

PR #2528, opened 2026-04-21
Eng 4687 enforce uns on write dfc input and fix get handler for write

FalconTube:
Check frontend PR https://github.com/united-manufacturing-hub/ManagementConsole/pull/3097 for all info


JeremyTheocharis (changes requested):
As discussed in the meeting:

1. toStringSlice
2. CHANGELOG user facing
3. hard-error → modify+degrade
4. UMH_TOPICS variable placement

umh-core/pkg/service/protocolconverter/runtime_config/runtime_config.go
```diff
@@ -314,6 +315,13 @@ func renderConfig(
 		return protocolconverterserviceconfig.ProtocolConverterServiceConfigRuntime{}, err
 	}
 
+	// Inject umh_topics into write DFC input after template rendering.
+	// []string values can't be rendered via text/template, so we set them directly.
+	// UMH_TOPICS presence is guaranteed by BuildRuntimeConfig above.
+	if unsInput, ok := write.BenthosConfig.Input["uns"].(map[string]any); ok {
+		unsInput["umh_topics"] = scope["UMH_TOPICS"]
+	}
```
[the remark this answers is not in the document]
FalconTube: Not required, see comment: 	// UMH_TOPICS presence is guaranteed by BuildRuntimeConfig above.

FalconTube:
> As discussed in the meeting:
> 
>     1. toStringSlice
> 
>     2. CHANGELOG user facing
> 
>     3. hard-error → modify+degrade
> 
>     4. UMH_TOPICS variable placement

1. Adjusted the godoc to make it easier to understand, why we need the function
2. Made the changelog more user facing
3. Instead of returning error, now set the `UMH_TOPICS` variable to `umh.v1.topic_not_set_by_user` if undefined. That way it is consistent with other `runtime_config.go` behavior
4. I talked to @rpaland and he actually prefers it the way that I implemented it here. He said, that his usage with the `AddressMappings` was more work to get implemented in the frontend especially. Therefore I will leave my implementation as is and Roger might adjust it in the future, if needed.

JeremyTheocharis:
`UMH_TOPICS` and `AddressMappings` are typed config fields, but we store them in `Variables.User` instead of in typed config blocks. That's why this PR needs the `toStringSlice` helper, the default `umh.v1.topic_not_set_by_user` value, and the manual `umh_topics` injection after rendering. All of those are working around the same mismatch.

No need to fix it all here, but we should make the intent clear for a future refactor. Can you file a Linear ticket in the milestone for moving the bridge to FSMv2? Direction: variables stay templating-only (copy a template, swap values, redeploy). Typed fields like `UMH_TOPICS` belong in their own typed config. If you want to template a typed field, put `{{ .UMH_TOPICS }}` in it. The field's home is the typed block, not the variables map.

Then add TODO comments linking the ticket at the workaround sites, so the next person sees the connection. Each TODO briefly says why the code exists and that it goes away when UMH_TOPICS moves to typed config. Sites:
- `edit-protocolconverter.go` — the persist block that writes `UMH_TOPICS` into `Variables.User` (the cause)
- `runtime_config.go` — the default value injection in `BuildRuntimeConfig`
- `runtime_config.go` — the manual `umh_topics` injection in `renderConfig`
- `action_helper.go` — the `toStringSlice` helper
- `get-protocolconverter.go` — the `if key == "UMH_TOPICS"` filter on variable iteration

JeremyTheocharis:
The type assertion at `runtime_config.go:319-324` looks like it can silently fail, but it can't in practice: `GetDFCWriteServiceConfig` always sets `Input["uns"]` when `Output > 0`, which is the same condition that gates the default `umh.v1.topic_not_set_by_user` block above. Future readers shouldn't have to trace it.

umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -1180,9 +1204,17 @@ func (a *EditProtocolConverterAction) GetDFCType() string {
...
-		if err := ValidateCustomDataFlowComponentPayload(*payload, false); err != nil {
+		// For write DFCs, skip input validation since input is auto-generated
+		validateInput := label != "write"
+		if err := ValidateCustomDataFlowComponentPayload(*payload, validateInput, false); err != nil {
 			return fmt.Errorf("invalid %s DFC configuration: %w", label, err)
 		}
+		// UMH_TOPICS is only required when the payload carries actual output config.
+		// A state-only change (no Outputs.Data) skips this check; the FSM's
+		// BuildRuntimeConfig enforces the requirement when it renders the write DFC.
+		if label == "write" && len(payload.Outputs.Data) > 0 && len(payload.UMHTopics) == 0 {
```
JeremyTheocharis, line 1218: Can you add the intent here? I'm guessing `Outputs.Data > 0` means full deploy/edit, and `Outputs.Data == 0` means just enable/disable?

umh-core/pkg/service/protocolconverter/runtime_config/runtime_config.go
```diff
@@ -194,11 +194,12 @@ func BuildRuntimeConfig(
 	}
 
 	vb.Internal["bridged_by"] = config.GenerateBridgedBy(config.ComponentTypeProtocolConverter, nodeName, pcName)
-	if locationPath == "" {
-		vb.Internal["umh_topic"] = "umh.v1.*"
-	} else {
-		vb.Internal["umh_topic"] = "umh.v1." + locationPath + ".*"
 
+	// UMH_TOPICS is required in user variables when write DFC is configured
```
JeremyTheocharis, line 201: Can you add the intent here? Same condition as the validation at `edit-protocolconverter.go:1207`, but in `BuildRuntimeConfig`. When does this branch actually fire? My guess: when we load a config from disk that's missing UMH_TOPICS (older config, or YAML parsing dropped it). Is that right?

umh-core/pkg/service/protocolconverter/runtime_config/runtime_config.go
```diff
@@ -194,11 +194,12 @@ func BuildRuntimeConfig(
...
+	if len(spec.Config.DataflowComponentWriteServiceConfig.BenthosConfig.Output) > 0 {
+		if _, ok := vb.User["UMH_TOPICS"]; !ok {
+			vb.User["UMH_TOPICS"] = []string{"umh.v1.topic_not_set_by_user"}
```
JeremyTheocharis: Can you document that this will result in an error message, and what it would look like? Just to ensure this isn't silent failure but the actual deployment fails. If it would be silent, maybe set this to `topic_not_set_by_user` without the `umh.v1.` prefix so the deployment fails loudly.

umh-core/pkg/communicator/actions/edit-protocolconverter.go
FalconTube: Correct, will add this to the comment

umh-core/pkg/service/protocolconverter/runtime_config/runtime_config.go
FalconTube: This can basically only happen, when user does not set it correctly through the `config.yaml`.
We protect him here.
Added comment

umh-core/pkg/service/protocolconverter/runtime_config/runtime_config.go
FalconTube: I did now set it to `TOPIC_NOT_SET_BY_USER` without the `umh.v1` prefix and added a comment as well

FalconTube:
> `UMH_TOPICS` and `AddressMappings` are typed config fields, but we store them in `Variables.User` instead of in typed config blocks. That's why this PR needs the `toStringSlice` helper, the default `umh.v1.topic_not_set_by_user` value, and the manual `umh_topics` injection after rendering. All of those are working around the same mismatch.
> 
> No need to fix it all here, but we should make the intent clear for a future refactor. Can you file a Linear ticket in the milestone for moving the bridge to FSMv2? Direction: variables stay templating-only (copy a template, swap values, redeploy). Typed fields like `UMH_TOPICS` belong in their own typed config. If you want to template a typed field, put `{{ .UMH_TOPICS }}` in it. The field's home is the typed block, not the variables map.
> 
> Then add TODO comments linking the ticket at the workaround sites, so the next person sees the connection. Each TODO briefly says why the code exists and that it goes away when UMH_TOPICS moves to typed config. Sites:
> 
>     * `edit-protocolconverter.go` — the persist block that writes `UMH_TOPICS` into `Variables.User` (the cause)
> 
>     * `runtime_config.go` — the default value injection in `BuildRuntimeConfig`
> 
>     * `runtime_config.go` — the manual `umh_topics` injection in `renderConfig`
> 
>     * `action_helper.go` — the `toStringSlice` helper
> 
>     * `get-protocolconverter.go` — the `if key == "UMH_TOPICS"` filter on variable iteration

Good point, I added the references now in this commit: https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/2528/changes/6528c72306a0dbd143476024a919ef41402196c9

JeremyTheocharis (approved):
Approving

One I think slipped through: comment 2 (top-level), about the type assertion at `runtime_config.go:330` — `if unsInput, ok := write.BenthosConfig.Input["uns"].(map[string]any); ok`. The existing comment above it covers UMH_TOPICS presence; what's missing is why the `Input["uns"]` map assertion can't fail in practice — `GetDFCWriteServiceConfig` always sets it when `Output > 0`, the same gate as the default block above. Pick it up in any follow-up commit if you agree, not blocking.

PR #2534, opened 2026-04-24
refactor(fsmv2): typed deps and worker standardization (2/3)

JeremyTheocharis:
## Summary

The feel test: apply the cleaned-up worker API across every existing worker and see how writing one feels now.

Part 2 of 3. Stacked on PR1 via GitButler; PR3 on top. Part of the FSMv2 worker-API cleanup under ENG-4796, triggered by findings from Janik's cert-fetcher port on top of the draft API. PR1's description has the full story.

Seven production workers migrate onto `register.Worker[TConfig, TStatus, TDeps]` plus typed deps: helloworld, communicator, transport, push, pull, persistence, application. About 50 state files across the migrated workers also get swept from `fsmv2.Result[T1, T2]{...}` to `fsmv2.Transition(...)` via PR1's architecture test.

Bundled on purpose -- the point is whether the API feels right after seven workers in a row, not any single migration in isolation.

### One exception

`register.Worker` widens from two generics to three (`TConfig, TStatus, TDeps`) to carry the typed deps struct. It could have gone in PR1, but extending the signature breaks every existing caller, so the re-fit has to ship in the same commit.

Persistence and application took the most work. Their observed-state types were custom, so they had to adopt `WorkerBase[TConfig, TStatus]` first, then pick up typed deps.

### Before and after

Before, a parent worker published dependencies into a string-keyed map and each child unpacked them by key. In practice only three keys were live across all production workers (`store` for persistence, `transport_deps` for transport's push/pull children); two more in `cmd/main.go` were unused.

```go
factory.RegisterWorkerAndSupervisorFactoryByType("transport",
    func(spec config.UserSpec, extraDeps map[string]any) (fsmv2.Worker, error) {
        handler, ok := extraDeps["connectionHandler"].(*transport.ConnectionHandler)
        if !ok { return nil, errors.New("missing connectionHandler") }
        // ~30 more lines of the same.
    })
```

After, one call with a typed dep struct:

```go
func init() {
    register.Worker[TransportConfig, TransportStatus, *TransportDependencies](
        "transport", NewTransportWorker)
}
```

### After this merges

Janik rebases cert fetcher on top. His current port `init()` sits at ~34 lines of the old `RegisterWorkerAndSupervisorFactoryByType` dance with panics on every `extraDeps` unpack and a separate CSE-registry call. Post-rebase it becomes one `register.Worker[CertFetcherConfig, CertFetcherStatus, CertFetcherDeps]` call plus constructor validation. Around 5 lines. No string keys anywhere.

## Test plan
- [ ] `go build ./...`
- [ ] `go vet ./...`
- [ ] `go test ./umh-core/pkg/fsmv2/... -race`
- [ ] Existing integration tests green (transport, push, pull, persistence, application).


JeremyTheocharis:
Closing: will be re-created via but pr new after GitButler metadata fix.

PR #2513, opened 2026-04-09
refactor(fsmv2): migrate 5 workers to NewObservation + WorkerBase

JeremyTheocharis:
## Summary

Migrates all 5 existing workers (HelloWorld, Communicator, Transport, Push, Pull) to the NewObservation + WorkerBase patterns from PR1. Net -968 lines.

Each worker now defines a `config.go` with just its Config and Status types — the framework handles everything else (timestamps, metrics, action history, children counts). Writing a new worker no longer requires understanding snapshot boilerplate.

Covers leaf workers (HelloWorld), parent workers (Transport, Communicator), and child workers (Push, Pull) to prove the pattern works across all worker types.

## Test plan

- [ ] `go test ./pkg/fsmv2/... -race`
- [ ] Architecture tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2512, opened 2026-04-09
refactor(fsmv2): Worker API improvements — NewObservation, control-loop docs

JeremyTheocharis:
## Summary

Framework-only changes to reduce boilerplate when writing FSMv2 workers. No worker behavior changes.

Each FSMv2 worker is a control loop with three phases: **sense** (`CollectObservedState`), **decide** (`State.Next`), **act** (`Action.Execute`). This PR simplifies the sense and decide phases:

**Sense — `CollectObservedState` simplified**

Previously, developers had to manually copy framework fields (timestamps, metrics, action history, children counts) into their observed state struct. Now they just return their business data:

```go
// Before: 15 lines of manual plumbing
// After:
return fsmv2.NewObservation(MyStatus{...})
```

The framework fills in everything else automatically after `CollectObservedState` returns.

**Decide — `State.Next` simplified**

State files previously used nested access (`snap.Observed.X`, `snap.Desired.Y`) with type-unsafe conversions. Now they use flat, typed snapshots:

```go
snap := fsmv2.ConvertWorkerSnapshot[MyConfig, MyStatus](snapAny)
snap.IsShutdownRequested  // direct field
snap.Status.MyField       // your data
snap.Config.MyOption      // user configuration
```

**Docs rewritten with control-loop vocabulary** (sensor/controller/actuator) based on workshop feedback. Drops "v2" framing from README.

### What this does NOT change

- No worker behavior changes
- Existing workers continue to work unchanged (backward compatible)

## Test plan

- [ ] `go test ./pkg/fsmv2/... -race`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2536, opened 2026-04-24
refactor(fsmv2): improve API foundation (1/3)

JeremyTheocharis:
## Summary

FSMv2 worker-API cleanup, plus two additions that came out of Janik's cert fetcher port.

Part 1 of 3. Stacked on staging via GitButler; PR2 and PR3 on top.

### Background

FSMv2 was already rolling out (ENG-4231 Communicator) but writing a new worker was painful in two different ways.

**People didn't know how to think about it.** "Does this logic belong in `CollectObservedState` or in an Action? What's the difference between observed and desired state? Where does config validation go?" The existing reference workers (helloworld, communicator) were full-featured, so there was no obvious starting point. Three pieces here address the teaching side:

- Control-loop framing (sensor → controller → actuator) so you have a thinking frame for what a worker IS.
- A mini-worker onboarding template: a zero-feature worker you start from and fill in, not helloworld which already carries config, status, parent/child, metrics, post-parse hooks.
- Features that opt in one at a time: custom deps, parent/child orchestration, custom metrics, custom actions, config validation, and deferred dep resolution. You only carry what you need.

**The mechanics were noisy in ways any Go developer would flag.** Some of what the team workshop on 2026-03-24 pulled apart:

- Every transition carried its type parameters: `fsmv2.Result[MyObservedState, MyDesiredState]{...}`. Inference should handle that.
- Typed actions had to be wrapped by hand: `fsmv2.WrapAction[MyDeps](&StartAction{})`. Writing `&StartAction{}` should just work.
- Parent-to-child deps went through `extraDeps map[string]any`, children unpacking by string key and runtime type-asserting. In a strongly-typed language.
- Worker registration required a separate `storage.GlobalRegistry().RegisterWorkerType()` call with `reflect.TypeOf(&MyStatus{})`. Reflection in user code is always a smell.
- Each worker shipped its own architecture test to catch misuse. Those tests belong in the framework.

People got confused. Draft PRs #2512 (framework additions: `NewObservation`, `WorkerBase`, `ConvertWorkerSnapshot`) and #2513 (migrate the 5 existing workers onto those additions) were already open from a week earlier and sat in draft.

On 2026-04-14 I pulled everything under ENG-4796 "FSMv2 is intuitive for developers" as the umbrella, with ENG-4799 for the API-debt pieces. That same day Janik ported cert fetcher onto `feat/worker-api-v2-prove-pattern` to feel-test the draft API. It still didn't. Two more things came up at the sit-down the day after:

- State transitions carried generic type parameters on every line (`fsmv2.Result[any, any]{...}`). Janik: "in dem Moment, wo ich da diese scheiss Generics lese, ist für mich immer so: Moment mal, was passiert hier?" (roughly: "the moment I read those shitty generics I just think: wait, what's happening here?")
- Cert fetcher needs a `certHandler` runtime dep. The 2-generic `register.Worker` had no slot for it, so he fell back to `factory.RegisterWorkerAndSupervisorFactoryByType` with a ~34-line `init()` containing two `panic(...)` calls unpacking `extraDeps map[string]any`, plus more panics around constructor failure and a separate CSE-registry call with manual `reflect.TypeOf`.

Those got folded in.

### The split

Three PRs, each with a different verification mode:

- **PR1 (this one)** -- does the API *look* right? Framework pieces only. Unit-testable on its own. Carries the existing cleanup (`NewObservation`, `WorkerBase`, `ConvertWorkerSnapshot`, control-loop README) plus the additions from Janik's port: a non-generic `fsmv2.Transition(state, signal, action, reason)`, reflection-based auto-wrap for typed actions (benchmarked at +4ns per call vs the legacy path, zero allocation delta), and an architecture test that fails the build if the old forms come back.
- **PR2** -- does developing with it *feel* right? Applies the API across every existing worker. Only verifiable by reading all the migrated workers side by side.
- **PR3** -- did we actually migrate everything? Deletes the legacy surfaces. If a worker still calls one of them the build fails; that's the check.

Normal flow across the stack: add the new API in PR1 without breaking existing callers, move every worker onto it in PR2, delete the old one in PR3. The one exception is `register.Worker` itself, which widens from two generics to three -- that extension breaks every caller, so the API change has to ship with its first users in PR2, not alone in PR1.

### What it looks like

State files go from

```go
return fsmv2.Result[MyObserved, MyDesired]{
    Next:   StateReady,
    Signal: SignalReady,
    Action: fsmv2.WrapAction[MyDeps](&StartAction{}),
    Reason: "observed healthy",
}
```

to

```go
return fsmv2.Transition(StateReady, SignalReady, &StartAction{}, "observed healthy")
```

No workers change here; that's PR2.

## Test plan
- [ ] `go build ./...`
- [ ] `go vet ./...`
- [ ] `go test ./umh-core/pkg/fsmv2/...`

---
This is **part 1 of 3 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #2538 
- <kbd>&nbsp;2&nbsp;</kbd> #2537 
- <kbd>&nbsp;1&nbsp;</kbd> #2536 👈


JeremyTheocharis:
Superseded by FSMv2 re-stack series.

This PR was part of the original 5-PR cascade (PR1..PR5: #2536, #2537, #2538, #2561, #2562). The cascade is being decomposed into ~11 concern-oriented lifecycle PRs for reviewability — each PR tells one feature story (introduce + migrate-all + delete-replaced) instead of bundling multiple migrations.

First lifecycle PR (L1: Observation[T] framework) lands as a fresh draft shortly. The remaining lifecycles (L2a..L9) follow sequentially.

Branch 'worker-api-v2-*' on the remote is preserved as a recovery snapshot of the cascade-tip state — not deleted. Closing the PR is a UI signal that review attention should move to the new lifecycle PRs.

End-state of the re-stack series matches end-state of this cascade (verified via cumulative diff-equality gate).

PR #2537, opened 2026-04-24
refactor(fsmv2): typed deps and worker standardization (2/3)

JeremyTheocharis:
## Summary

The feel test: apply the cleaned-up worker API across every existing worker and see how writing one feels now.

Part 2 of 3. Stacked on PR1 via GitButler; PR3 on top. Part of the FSMv2 worker-API cleanup under ENG-4796, triggered by findings from Janik's cert-fetcher port on top of the draft API. PR1's description has the full story.

Seven production workers migrate onto `register.Worker[TConfig, TStatus, TDeps]` plus typed deps: helloworld, communicator, transport, push, pull, persistence, application. About 50 state files across the migrated workers also get swept from `fsmv2.Result[T1, T2]{...}` to `fsmv2.Transition(...)` via PR1's architecture test.

Bundled on purpose -- the point is whether the API feels right after seven workers in a row, not any single migration in isolation.

### One exception

`register.Worker` widens from two generics to three (`TConfig, TStatus, TDeps`) to carry the typed deps struct. It could have gone in PR1, but extending the signature breaks every existing caller, so the re-fit has to ship in the same commit.

Persistence and application took the most work. Their observed-state types were custom, so they had to adopt `WorkerBase[TConfig, TStatus]` first, then pick up typed deps.

### Before and after

Before, a parent worker published dependencies into a string-keyed map and each child unpacked them by key. In practice only three keys were live across all production workers (`store` for persistence, `transport_deps` for transport's push/pull children); two more in `cmd/main.go` were unused.

```go
factory.RegisterWorkerAndSupervisorFactoryByType("transport",
    func(spec config.UserSpec, extraDeps map[string]any) (fsmv2.Worker, error) {
        handler, ok := extraDeps["connectionHandler"].(*transport.ConnectionHandler)
        if !ok { return nil, errors.New("missing connectionHandler") }
        // ~30 more lines of the same.
    })
```

After, one call with a typed dep struct:

```go
func init() {
    register.Worker[TransportConfig, TransportStatus, *TransportDependencies](
        "transport", NewTransportWorker)
}
```

### After this merges

Janik rebases cert fetcher on top. His current port `init()` sits at ~34 lines of the old `RegisterWorkerAndSupervisorFactoryByType` dance with panics on every `extraDeps` unpack and a separate CSE-registry call. Post-rebase it becomes one `register.Worker[CertFetcherConfig, CertFetcherStatus, CertFetcherDeps]` call plus constructor validation. Around 5 lines. No string keys anywhere.

## Test plan
- [ ] `go build ./...`
- [ ] `go vet ./...`
- [ ] `go test ./umh-core/pkg/fsmv2/... -race`
- [ ] Existing integration tests green (transport, push, pull, persistence, application).

---
This is **part 2 of 3 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #2538 
- <kbd>&nbsp;2&nbsp;</kbd> #2537 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #2536


JeremyTheocharis:
Superseded by FSMv2 re-stack series.

This PR was part of the original 5-PR cascade (PR1..PR5: #2536, #2537, #2538, #2561, #2562). The cascade is being decomposed into ~11 concern-oriented lifecycle PRs for reviewability — each PR tells one feature story (introduce + migrate-all + delete-replaced) instead of bundling multiple migrations.

First lifecycle PR (L1: Observation[T] framework) lands as a fresh draft shortly. The remaining lifecycles (L2a..L9) follow sequentially.

Branch 'worker-api-v2-*' on the remote is preserved as a recovery snapshot of the cascade-tip state — not deleted. Closing the PR is a UI signal that review attention should move to the new lifecycle PRs.

End-state of the re-stack series matches end-state of this cascade (verified via cumulative diff-equality gate).

PR #2543, opened 2026-04-27
fix: redpanda integration test

led0nk:
## Summary
- fix flaky `redpanda-only` integration test that fails because it checks reconcile histogram values after container start, before the control loop completes its first cycle
- The `Eventually` block now waits for `umh_core_reconcile_duration_milliseconds` to appear in metrics (not just HTTP 200), matching the pattern used by all other integration tests

## Root Cause
The test was the only one that didn't wait for the histogram metric to exist before asserting on its values. The container's metrics endpoint starts instantly (memory, errors, starved counters are available), but the control loop histogram isn't emitted until the first reconciliation completes. Every other integration test already includes this check.


FalconTube (approved):
Looks reasonable, let's see how it holds up in the CI :)

PR #2544, opened 2026-04-27
Eng 4685 add tests for per dfc state handling in write flows

FalconTube:
- Add tests for write flows and disable-read-flow


JeremyTheocharis (approved):
LGTM. Tests drive the realistic Parse → Validate → Execute path with JSON-shaped payloads — strong user-behavior fit. A few names promise more than their assertions pin (state preservation, awaitRollout convergence) but nothing that should hold up the merge.

PR #2545, opened 2026-04-27
[ENG-4823] Populate Bridge Data Contracts

rpaland:


rpaland:
Will be implemented at a later point, in a different approach.

PR #2546, opened 2026-04-27
Eng 4754 sentry integration for write flows

FalconTube:
- Add sentry tests for `disable-read-flow` feature and `write flows` feature
- Both behind same feature flag in umh-core and in sentry

### DIFF
--- a/umh-core/pkg/communicator/actions/edit-protocolconverter.go
+++ b/umh-core/pkg/communicator/actions/edit-protocolconverter.go
@@ -48,6 +48,7 @@ import (
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/constants"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsm"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsm/protocolconverter"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsmv2/deps"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/logger"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/models"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/service/protocolconverter/runtime_config"
@@ -95,8 +96,9 @@ type EditProtocolConverterAction struct {
 	// Runtime observation for health checks
 	systemSnapshotManager *fsm.SnapshotManager
 
-	actionLogger   *zap.SugaredLogger
-	userEmail      string
+	actionLogger *zap.SugaredLogger
+	fsmLogger    deps.FSMLogger
+	userEmail    string
 	name           string // protocol converter name (optional for updates)
 	dfcType        DFCType
 	connectionPort string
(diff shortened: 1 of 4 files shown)


umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -832,10 +843,13 @@ func (a *EditProtocolConverterAction) awaitRollout(pcConfig config.ProtocolConve
 					_, err := a.configManager.AtomicEditProtocolConverter(ctx, a.atomicEditUUID, pcConfig)
 					if err != nil {
 						a.actionLogger.Errorf("failed to roll back protocol converter %s: %v", a.name, err)
+						a.fsmLogger.SentryError(deps.FeatureDisableReadFlows, "", err, "edit_protocol_converter_config_error_rollback_failed")
```
JeremyTheocharis: do you have here everything or wouldn't it maybe help to provide the full actionLogger history on this action as artifact or however it is called in sentry? otherwise you have this error, but probably wouldnt know what to do with it. 

same for all the other additions of sentry warn

umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -569,10 +578,12 @@ func (a *EditProtocolConverterAction) awaitRollout(pcConfig config.ProtocolConve
 			if err != nil {
 				a.actionLogger.Errorf("Failed to rollback to previous configuration: %v", err)
 				stateMessage := fmt.Sprintf("Protocol converter '%s' edit timeout reached. It did not become %s in time. Rolling back to previous configuration failed: %v", a.name, desiredPCState, err)
+				a.fsmLogger.SentryError(deps.FeatureDisableReadFlows, "", err, "edit_protocol_converter_rollback_failed")
 
 				return models.ErrRetryRollbackTimeout, fmt.Errorf("%s", stateMessage)
 			} else {
 				stateMessage := fmt.Sprintf("Protocol converter '%s' edit timeout reached. It did not become %s in time. Rolled back to previous configuration", a.name, desiredPCState)
+				a.fsmLogger.SentryWarn(deps.FeatureDisableReadFlows, "", "edit_protocol_converter_rollback_on_timeout")
```
JeremyTheocharis: maybe you can include a snapshot or the previous outputted logs to identify what made it timeout. if this error would appear, i would probably need to check the fsm state and all state transitions. maybe you can add it here, and probably in other places as well, as an "artifact"

umh-core/pkg/fsmv2/deps/feature.go
```diff
@@ -71,7 +71,10 @@ const (
 	// writing, backup, and validation.
 	FeatureFSMv1ConfigManager Feature = "fsmv1_config_manager"
 
-	// FeatureDisableReadFlows controls the feature about activating and deactivating read flows.
+	// FeatureDisableReadFlows covers errors from the read-flow and write-flow
```
JeremyTheocharis, line 74: have you configured so far the sentry rule that will make this auto-create tickets for you?

JeremyTheocharis (changes requested):

umh-core/pkg/fsmv2/deps/feature.go
FalconTube: Thanks for heads up, added this now and verified it works

umh-core/pkg/communicator/actions/edit-protocolconverter.go
FalconTube: Added more info to all Sentry issues in last two commits

umh-core/pkg/communicator/actions/edit-protocolconverter.go
FalconTube: Added more info to all Sentry issues in last two commits

JeremyTheocharis (approved):

PR #2552, opened 2026-04-30
fix(s6): bridges self-recover from corrupted service directory (ENG-4862)

JeremyTheocharis:
## Problem

When a bridge's internal service directory gets corrupted or deleted at runtime, umh-core detects the problem but cannot recover. The bridge gets stuck in a starting state. The UI shows it spinning, manual delete fails, and the only fix is to restart umh-core or recreate the bridge under a different name. This pattern has appeared at multiple customer sites since 2024 (ENG-3877, ENG-4760 still open in Ideas).

Why it gets stuck: umh-core already had a force-remove fallback for corruption that happens while a bridge is running or stopped (PR #2259, ENG-3468). But the lifecycle states (`Creating`, `to_be_created`, `Removing`) had no equivalent. When corruption hit during startup, the bridge entered `Creating` and stayed there. This PR closes that gap.

## What we're shipping

Introduced a simplified API: `Service.Health(ctx, fs)`. Every S6 service now checks in all available states whether health is OK, whether to wait (`ActionWait`), or whether to recreate (`ActionRecreate`).

| Directive | Meaning |
|-----------|---------|
| `ActionOK` | Proceed normally |
| `ActionWait` | Transient condition (I/O error or removal in progress); skip this tick |
| `ActionRecreate` | Corruption detected; trigger force-remove and rebuild |

## What we're NOT shipping

- **FSMv2**: no s6 worker exists in FSMv2 yet. The service layer migrates unchanged when it does.
- **`Teardown(mode)` consolidation**: collapsing `Remove` + `ForceRemove` into one method; larger blast radius, deferred.
- **20-tick fallback no-op**: `(*BaseFSMInstance).Remove` returns nil for `Creating` state; the fallback fires but does nothing. The `Health()` path makes this irrelevant for the corruption case, but the underlying no-op is a separate fix.
- **13 sibling FSMs**: all have the same `IsLifecycleState` dispatch pattern. They don't have the same `artifacts` tracking as s6, so the bug class doesn't apply now. Deferred cleanup once this bakes.
- **Root cause**: what corrupts the service directory in the first place (network blip, OOM, disk hiccup). A filesystem snapshot at the next occurrence is needed. This PR handles the recovery; diagnosis is separate.

### DIFF
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -9,7 +9,7 @@
 
 ### Fixes
 
-- A bridge that got stuck in a starting state due to file corruption now recovers automatically. Previously, recovery required restarting umh-core or recreating the bridge under a different name.
+- A data flow (bridge, standalone, or stream processor) could get stuck in a starting state when its on-disk service directory was left in an inconsistent state — for example after a container restart or OOM-kill interrupted setup. umh-core now detects the inconsistency and rebuilds the directory automatically. Previously, recovery required restarting umh-core or recreating the data flow under a different name.
 
 ## [0.44.19]
 
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -7,6 +7,10 @@
 - Previously, the component tasked with communicating with the UI could stop working if it rebuilt its HTTP connection (which happens after persistent network failures) while a message was being sent or received. Connection rebuilds and in-flight requests are now coordinated so they cannot interfere.
 - Config backup is now enabled by default. Previously, automatic config.yaml backups required setting `ENABLE_CONFIG_BACKUP=true`. After validating the feature in customer environments for over two months, every umh-core instance now writes timestamped copies of `config.yaml` to `/data/config-backups/` on every config write, retaining the 100 most recent versions (roughly 5–20 MB). To opt out, set `ENABLE_CONFIG_BACKUP=false`. To roll back a bad config, use the second-most-recent backup — the most recent one mirrors the bad write
 
+### Fixes
+
+- A bridge whose underlying service directory got corrupted at runtime previously stayed in a starting state until umh-core was restarted or replaced with a different bridge name. umh-core now detects the corruption and recovers the bridge automatically.
+
 ## [0.44.19]
 
 ### Improvements
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -9,7 +9,7 @@
 
 ### Fixes
 
-- A bridge whose underlying service directory got corrupted at runtime previously stayed in a starting state until umh-core was restarted or replaced with a different bridge name. umh-core now detects the corruption and recovers the bridge automatically.
+- A bridge that got stuck in a starting state due to file corruption now recovers automatically. Previously, recovery required restarting umh-core or recreating the bridge under a different name.
 
 ## [0.44.19]
 
--- a/umh-core/pkg/fsm/s6/healthbad_lifecycle_integration_test.go
+++ b/umh-core/pkg/fsm/s6/healthbad_lifecycle_integration_test.go
@@ -1,0 +1,114 @@
+// Copyright 2025 UMH Systems GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package s6_test
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/config"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsm"
+	s6fsm "github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsm/s6"
+	s6service "github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/service/s6"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/serviceregistry"
+)
+
+var _ = Describe("S6 recovery directive — FSM dispatch", func() {
+	var (
+		ctx      context.Context
+		cancel   context.CancelFunc
+		mockSvc  *s6service.MockService
+		registry serviceregistry.Provider
+		inst     *s6fsm.S6Instance
+	)
+
+	BeforeEach(func() {
+		ctx, cancel = context.WithCancel(context.Background())
+
+		mockSvc = s6service.NewMockService()
+		mockSvc.MockExists = true
+		mockSvc.MockHealthAction = s6service.ActionOK
+
+		registry = serviceregistry.NewMockRegistry()
+		cfg := config.S6FSMConfig{
+			FSMInstanceConfig: config.FSMInstanceConfig{
+				Name:            "test-svc",
+				DesiredFSMState: s6fsm.OperationalStateRunning,
+			},
+		}
+		var err error
+		inst, err = s6fsm.NewS6InstanceWithService("/dev/null", cfg, mockSvc)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	AfterEach(func() { cancel() })
+
+	// driveTicks runs Reconcile across `count` ticks starting at `startTick`,
+	// returning early if the FSM reaches Removed. Errors are intentionally
+	// swallowed so the loop can drive past tick boundaries.
+	driveTicks := func(startTick uint64, count int) {
+		for tick := startTick; tick < startTick+uint64(count); tick++ {
+			_, _ = inst.Reconcile(ctx, fsm.SystemSnapshot{Tick: tick}, registry)
+			if inst.IsRemoved() {
+				return
+			}
+		}
+	}
+
+	It("drives the FSM to Removed when Health returns ActionRecreate from any state", func() {
+		mockSvc.MockHealthAction = s6service.ActionRecreate
+
+		driveTicks(1, 30)
+
+		Expect(inst.IsRemoved()).To(BeTrue(),
+			"FSM must reach Removed via Health() recovery dispatch within 30 ticks")
+		Expect(mockSvc.ForceRemoveCalled).To(BeTrue(),
+			"ForceRemove must fire on the underlying service to clear corruption")
+	})
+
+	It("does not re-fire after recovery (ForceRemove called exactly once across 50 post-recovery ticks)", func() {
+		mockSvc.MockHealthAction = s6service.ActionRecreate
+		driveTicks(1, 30)
+		Expect(inst.IsRemoved()).To(BeTrue())
+		Expect(mockSvc.ForceRemoveCallCount).To(Equal(1),
+			"ForceRemove fires exactly once per corruption event")
+
+		mockSvc.MockHealthAction = s6service.ActionOK
+
+		for tick := uint64(31); tick < 81; tick++ {
+			_, _ = inst.Reconcile(ctx, fsm.SystemSnapshot{Tick: tick}, registry)
+		}
+
+		Expect(mockSvc.ForceRemoveCallCount).To(Equal(1),
+			"once Health flips back to ActionOK, no further ForceRemove calls may fire")
+	})
+
+	It("escapes from operational state when Health returns ActionRecreate mid-flight", func() {
+		driveTicks(1, 30)
+		Expect(inst.IsRemoved()).To(BeFalse(),
+			"baseline: FSM should not be Removed under healthy mock with desired=Running")
+		Expect(mockSvc.ForceRemoveCalled).To(BeFalse(),
+			"baseline: no recovery should fire under healthy mock")
+
+		mockSvc.MockHealthAction = s6service.ActionRecreate
+		driveTicks(31, 30)
+
+		Expect(inst.IsRemoved()).To(BeTrue(),
+			"FSM must escape from operational state via Health() recovery dispatch")
+		Expect(mockSvc.ForceRemoveCalled).To(BeTrue())
+	})
+})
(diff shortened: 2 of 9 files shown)


JeremyTheocharis:
## Docker experiment: reproduction + fix validation

Before calling this PR validated, I ran the same corruption test against both builds locally.

**Method** — continuous deletion loop (50ms interval) inside the container, targeting both the s6 scan symlink and the repository directory. Same config (single benthos `hello-world` generate service), same timing (68s corruption, 60s observation after stopping).

```
RED (no fix): ghcr.io/united-manufacturing-hub/umh-core:sha-8079ac3 (staging)
GREEN (fix):  local build from this branch
```

To reproduce yourself:
```bash
docker run -d --name umh-test -e LOGGING_LEVEL=DEBUG -e S6_PERSIST_DIRECTORY=true \
  -v <config.yaml>:/data/config.yaml <image>

# wait ~30s for benthos service to start, then:
docker exec -d umh-test sh -c '
  while [ ! -f /tmp/stop ]; do
    rm -rf /run/service/benthos-hello-world /data/services/benthos-hello-world 2>/dev/null
    sleep 0.05
  done'

# stop after 60s:
docker exec umh-test touch /tmp/stop
```

---

### RED: staging, no fix

| Signal | Count |
|--------|-------|
| `"directory integrity is bad"` | 1481 |
| `"Force removing S6 service"` | **1** |
| `"requires recreation"` (fix log, absent on staging) | 0 |
| `"Failed to ensure service supervision"` | **1476** |

The 1 ForceRemove fired 2s into corruption from the existing operational-state escape (PR #2259). The FSM then recreated into `creating` and stopped there:

```
07:26:22  FSM benthos-hello-world: 'to_be_created' → 'creating'
          [no further transitions — 7+ minutes of observation]
```

Corruption stopped. 60s passed. Still stuck. The 1476 supervision failures match the production log signature from the Sentry report: CheckForCreation failing every 100ms tick with no lifecycle escape.

---

### GREEN: fix branch

Same method, same timing.

| Signal | Count |
|--------|-------|
| `"directory integrity is bad"` | 75 |
| `"Force removing S6 service"` | **24** |
| `"requires recreation"` (fix log) | **69** |
| `"Failed to ensure service supervision"` | 0 |

The fix firing from `creating` state (last 3 entries before corruption stopped):
```
07:32:09.471  S6 service benthos-hello-world requires recreation (state=creating)
07:32:09.573  S6 service benthos-hello-world requires recreation (state=creating)
07:32:09.670  S6 service benthos-hello-world requires recreation (state=creating)
```

Corruption stopped at 07:32:10. Full recovery from the log:
```
07:32:10.684  FSM benthos-hello-world: 'to_be_created' → 'creating'
07:32:10.777  FSM benthos-hello-world: 'creating'     → 'stopped'
07:32:10.870  FSM benthos-hello-world: 'stopped'      → 'starting'
07:32:10.970  FSM benthos-hello-world: 'starting'     → 'running'
07:32:11.974  FSM hello-world:         'degraded'     → 'active'
```

1.3 seconds from corruption-stop to `active`.

---

ForceRemove goes from 1 to 24 because on staging the FSM wedges in `creating` after the first recovery and never escapes again. On this branch, every tick with HealthBad fires another cycle (Health() → ActionRecreate → permanent error → ForceRemove), so the count climbs throughout the corruption window. When the loop stops, the next Create succeeds, Health() returns ActionOK on the following tick, and the FSM converges.

Zero supervision failures on GREEN is the key signal. On staging, `CheckForCreation` calls `EnsureSupervision` on every tick with missing dirs — 1476 times. On this branch, `Health()` short-circuits at the top of `reconcileStateTransition` before `CheckForCreation` runs, so `EnsureSupervision` is never called under corruption.

Full raw experiment log: `003_klaus/artifacts/2026-04-29_eng4862_triage/reimpl/docker_experiment.md`

JeremyTheocharis:
@claude review

umh-core/pkg/service/s6/lifecycle_force.go
```diff
@@ -75,12 +75,26 @@ func (s *DefaultService) ForceCleanup(ctx context.Context, artifacts *ServiceArt
 		s.logger.Warnf("Failed to remove log directory: %v", err)
 	}
 
+	// Step 4: Remove the repository directory. The scan dir (ServiceDir) is a
+	// symlink; the actual service files live in RepositoryDir. Without this step
+	// the repository is orphaned on disk and only recovered when the next Create
+	// call detects the stale directory.
+	if artifacts.RepositoryDir != "" {
+		if err := s.removeDirectoryWithTimeout(ctx, artifacts.RepositoryDir, fsService); err != nil {
+			s.logger.Warnf("Failed to remove repository directory: %v", err)
```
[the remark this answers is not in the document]
JeremyTheocharis: Dismissing as intentional design. `removeDirectoryWithTimeout` explicitly overrides to `context.Background()` — the comment at line 113 documents this, and `ForceRemove` reinforces it by wrapping the caller context with `context.WithTimeout(context.Background(), constants.ForceRemovalTimeout)` before entering `ForceCleanup` (ENG-3420 comment in s6.go). Force cleanup is designed to run to completion regardless of reconcile context expiry. The `Create` recreate path also calls `ForceCleanup` directly with the reconcile ctx, and the same override applies — aggressive cleanup during a recreate should not be cancelled mid-way by a reconcile timeout.

umh-core/pkg/service/s6/lifecycle_force.go
```diff
@@ -75,12 +75,26 @@ func (s *DefaultService) ForceCleanup(ctx context.Context, artifacts *ServiceArt
...
 	logExists, _ := fsService.PathExists(ctx, artifacts.LogDir)
+	repoExists := false
+	if artifacts.RepositoryDir != "" {
+		repoExists, _ = fsService.PathExists(ctx, artifacts.RepositoryDir)
+	}
 
-	if serviceExists || logExists {
-		return fmt.Errorf("force cleanup incomplete: service=%v, log=%v", serviceExists, logExists)
+	if serviceExists || logExists || repoExists {
+		return fmt.Errorf("force cleanup incomplete: service=%v, log=%v, repo=%v", serviceExists, logExists, repoExists)
```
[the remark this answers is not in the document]
JeremyTheocharis: Fixed in commit 6d26fa51d. Propagated errors from all three PathExists checks (ServiceDir, LogDir, RepositoryDir) rather than just the new one — lines 89-90 had the same pre-existing discard pattern. A PathExists I/O error now causes ForceCleanup to return an error, which causes ForceRemove to skip Store(nil), so Health() keeps returning ActionRecreate on the next tick and the FSM retries rather than falsely reporting cleanup complete.

umh-core/pkg/service/s6/s6.go
```diff
@@ -394,15 +447,24 @@ func (s *DefaultService) Remove(ctx context.Context, servicePath string, fsServi
...
-		// No tracked files - service is in an inconsistent state
-		// Remove() requires tracked files to work properly
+		// No tracked files - service is in an inconsistent state. Returning the
+		// typed ErrServiceNotExist lets the FSM-layer caller use errors.Is to
+		// treat a no-tracking Remove as success (idempotent), rather than
+		// pattern-matching a string error.
 		s.logger.Debugf("No tracked files for service %s, cannot do tracked removal", servicePath)
 
-		return fmt.Errorf("service %s has no tracked files - use ForceRemove() instead", servicePath)
+		return ErrServiceNotExist
```
led0nk, line 500: doesn't this leak the dir then onto the disk after agent reastart when it was previously forceRemoved instead?

umh-core/pkg/fsm/s6/healthbad_lifecycle_integration_test.go
```diff
@@ -0,0 +1,114 @@
...
+		driveTicks(1, 30)
+
+		Expect(inst.IsRemoved()).To(BeTrue(),
+			"FSM must reach Removed via Health() recovery dispatch within 30 ticks")
+		Expect(mockSvc.ForceRemoveCalled).To(BeTrue(),
+			"ForceRemove must fire on the underlying service to clear corruption")
+	})
+
+	It("does not re-fire after recovery (ForceRemove called exactly once across 50 post-recovery ticks)", func() {
+		mockSvc.MockHealthAction = s6service.ActionRecreate
```
led0nk, line 85: i think this here only flips and doesnÄt then walk the fsm into "creating". Please double-check, if so the test doesnt really make sense

umh-core/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- A bridge that got stuck in a starting state due to file corruption now recovers automatically. Previously, recovery required restarting umh-core or recreating the bridge under a different name.
```
led0nk: might be then relevant for standalones as well right? in theory also for streamprocessors

led0nk (changes requested):
Also out of an additional self-review with that i got this here:

🟠 Recreate-loop on flaky filesystems (design — discussion thread)
  On a transient-but-persistent FS issue (NFS hiccup, EIO), the loop is:
  Health=ActionRecreate → ForceRemove → Create succeeds → ResetState
  (reconcile.go:160) wipes backoff → next tick FS still flaky → repeat. No
  per-instance rate limit on ForceRemove. Worth a backoff or recreate-counter
  before merge.

umh-core/CHANGELOG.md
JeremyTheocharis: Yes — verified the delegation chain: `streamprocessor → dataflowcomponent → benthos → s6`. All three flow types delegate process supervision to s6 sub-FSMs, so the fix applies uniformly. Updated the entry to:

> A data flow (bridge, standalone, or stream processor) could get stuck in a starting state when its on-disk service directory was left in an inconsistent state — for example after a container restart or OOM-kill interrupted setup. umh-core now detects the inconsistency and rebuilds the directory automatically. Previously, recovery required restarting umh-core or recreating the data flow under a different name.

umh-core/pkg/fsm/s6/healthbad_lifecycle_integration_test.go
JeremyTheocharis: You were right — the rename didn't close the gap. Added two more cases in `a7a6f5058`: `ActionRecreate` from `Creating` and from `Running`. Both reach `Removed` with exactly one `ForceRemove`.

umh-core/pkg/service/s6/s6.go
JeremyTheocharis: Confirmed. Traced the failure mode:

**The leak.** Service is created normally; on-disk dirs + in-memory `artifacts` agree. umh-core crashes (OOM, segfault, hard kill). On restart, `s.artifacts.Load()` is nil — fresh process has no record. User then removes the data flow from config. FSM enters `Removing` → calls `Remove()` → no tracked files → returns `ErrServiceNotExist`. The FSM-layer at `actions.go:114` treats that as idempotent-delete success → state=Removed → manager destroys the instance. ForceRemove never fires. On-disk dirs are orphaned for the lifetime of the agent.

This is a pre-existing gap (`artifacts` was already in-memory pre-PR), but this PR's `atomic.Pointer` migration touches exactly this code path — right time to close it.

**The fix** — commit `d0642422f` ("probe disk on no-tracking Remove to prevent orphans"), pushed to this PR. Inside the no-tracked-files branch of `Remove()`, probe the canonical disk paths (servicePath, repoPath via `constants.GetS6RepositoryBaseDir`, logPath via `constants.S6LogBaseDir`). If any exist, return a `backoff.PermanentFailureError`-prefixed error instead of `ErrServiceNotExist`. The existing machinery does the rest: next tick → `HandlePermanentError` → `forceRemove` → `ForceRemove` synthesizes artifacts from the canonical paths → `ForceCleanup` wipes disk → `Store(nil)` → state=Removed.

Genuinely-empty case (artifacts=nil + nothing on disk) still returns `ErrServiceNotExist`. PathExists I/O errors propagate up.

JeremyTheocharis:
You called the loop correctly. The mechanism turned out to be one layer deeper than the trace suggested, and that pushed the fix to a different place than a backoff/counter.

Root cause: `CheckArtifactsHealth` was examining s6-supervise's own `supervise/` and `log/supervise/` directories and flagging asymmetric mid-bringup state as service-author corruption. Per skarnet's docs (servicedir, s6-svscan), those directories are s6-supervise's private runtime state, brought up in two non-atomic steps. Asymmetry during bringup is the documented normal mid-state, not corruption. The Health() → ActionRecreate → ForceRemove dispatch you reviewed is correct; the integrity check it dispatches from was the wrong abstraction.

Two architectural corrections in the stacked PR:
- Drop the supervise-dir asymmetry check from `CheckArtifactsHealth`. Narrows HealthBad to legitimate UMH-authored corruption only.
- Complete the canonical skarnet manual-unlink recipe in `ForceCleanup` by adding `s6-svscanctl -an` after `RemoveAll`. We were doing the `rm` half and skipping the scanner-notification half; on `s6-svunlink` exit 111 (s6 temporary error) that left the scanner's internal supervisor table out of sync and producing a cumulative wedge.

One nit on your trace: the loop wasn't via `ResetState` wiping backoff — it formed at the manager level via instance recycling after each ForceRemove. Doesn't change the conclusion (loop was real); just where it lived.

Verified two ways. (1) Independent clean-room read of skarnet docs from scratch + audit of our code, by an agent with no prior context, converged on the same fix shape — verdict SHIP. (2) End-to-end auto-heal experiment: 30s of supervise-dir corruption produces 0 force-removes (vs 6 in prior runs) and 0 `directory integrity is bad` warnings (vs 3789). The FSM never even leaves `active`. Wedge prevention by construction, not recovery.

Stacked PR: https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/2569

led0nk (approved):
good to go now

umh-core/pkg/fsm/s6/healthbad_lifecycle_integration_test.go
```diff
@@ -0,0 +1,176 @@
...
+	// returning early if the FSM reaches Removed. Errors are intentionally
+	// swallowed so the loop can drive past tick boundaries.
+	driveTicks := func(startTick uint64, count int) {
+		for tick := startTick; tick < startTick+uint64(count); tick++ {
+			_, _ = inst.Reconcile(ctx, fsm.SystemSnapshot{Tick: tick}, registry)
+			if inst.IsRemoved() {
+				return
+			}
+		}
+	}
```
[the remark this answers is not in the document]
JeremyTheocharis: Resolved without code change. Rationale: this file uses deterministic tick-stepping (`driveTicks`/`driveUntilState`) because FSM reconciliation advances per-tick, not per-wall-clock. `Eventually()`'s `WithTimeout/WithPolling` semantics introduce a wall-clock race the FSM doesn't have. Repo-wide check: `Eventually()` appears in exactly one FSM test file (`base_manager_test.go`), all six `It(...)` blocks in this file use manual tick loops. Switching only the two new tests would create mixed style in the file; switching all six is a separate refactor PR with no test-behavior win for this deterministic case.

PR #2570, opened 2026-05-12
feat(fsmv2/L2a): WorkerBase[T] embed, ShouldStop rename, InitBase propagation

JeremyTheocharis:
# FSMv2: WorkerBase[T] generic embed + ShouldStop rename — 12 workers, typed config/deps moved into the framework

## Why

Every worker implemented the same four interface methods (`DeriveDesiredState`, `GetInitialState`,
`BindDeps`, `GetDependenciesAny`) by hand. The bodies were nearly identical: call the base
constructor, stash the typed config, stash the typed deps, return a zero-value desired state.
Easy to mis-wire the type assertion, easy to shadow the `ok` variable in the deps factory, and
reviewers don't catch it because the repetition looks intentional.

`IsStopRequired()` existed alongside the new `ShouldStop()` name added in the earlier rename
commit — two names for the same predicate made grep ambiguous and left the architecture
validator with a broken discriminant (it searched for `"ShouldStop()"` in snapshot.go, which
never existed there).

`WorkerBase[TConfig, TStatus, TDeps]` moves that wiring into the framework. Workers embed it
and get typed config, typed deps, and a default `DeriveDesiredState` for free. `ShouldStop()`
is now the only name.

## Scale

Twelve workers shed four boilerplate methods each; the framework absorbs the plumbing as
generic code with full test coverage. The old `helpers.BaseWorker` shim and the
`CHILD_MUST_USE_SHOULD_STOP` architecture validator (which enforced a now-vacuous invariant)
are deleted. Every file that carried `IsStopRequired` — 32 source files and 3 markdown docs —
is updated atomically in a single sweep.

## What's in this PR

**`WorkerBase[TConfig, TStatus, TDeps]` generic embed** (`worker_base.go`) — provides
`DeriveDesiredState` (calls `BindDeps` internally, returns the typed config), `GetInitialState`,
`BindDeps`, and `GetDependenciesAny`. Workers embed it and override only the methods where they
need custom behavior. Also adds a small public `BaseUserSpec` interface in `api.go` (~6 lines)
so `DeriveDesiredState` can duck-type `GetState()` off `TConfig` without a concrete type
assertion.

**`ShouldStop()` rename** — Atomic 35-file sweep replaces every `IsStopRequired` reference with
`ShouldStop`. `ShouldStop` becomes the sole entry point.

**`ErrNoDesiredState` propagation through `DesiredStateProvider`** (`collector.go`) — Transient
errors (store timeout, deserialization failure) that are not `ErrNoDesiredState` skip the
collection cycle and log at `Info` with `deps.Err(err)`. They do not call `SentryWarn`, which
would double-fire through api.go's closure and exhaust Sentry quota on sticky errors.
`ErrNoDesiredState` is the normal "nothing configured yet" path and logs at trace.

**12 workers migrate to `WorkerBase`** — communicator, persistence, application, transport
(parent + push + pull), and all 6 example workers (helloworld, examplechild, examplefailing,
examplepanic, exampleparent, exampleslow). Each worker's `NewXxxWorker` constructor now takes
`deps.Identity` and `deps.FSMLogger` instead of ad-hoc parameter shapes (implicit in the
`InitBase` requirement; no in-tree production callers outside `pkg/fsmv2/`).

**Dead storage removal** — `stateReader deps.StateReader` and `baseDeps *deps.BaseDependencies`
fields removed from `WorkerBase`. Both were written in `InitBase` but never read back; the
supervisor accesses deps through the typed field on each worker, not through these base fields.

**`example_panic` → `examplepanic` package rename** — Applied alongside the examplepanic
WorkerBase migration. Go convention: no underscores in package names. All in-tree callers
updated; no external consumers.

**`CHILD_MUST_USE_SHOULD_STOP` validator deleted** — Forward-deletion: the validator checked
that child workers call `ShouldStop()`, but its `isChildWorker()` discriminant searched for
`"ShouldStop()"` in snapshot/snapshot.go (where it never existed), so it always returned false
and never actually validated anything. The discriminant used `ParentMappedState` as its fix —
but `ParentMappedState` is itself scheduled for deletion in L5. Deleting the whole block now
avoids a validator that silently skips its target today and breaks L5 if left in place.

**InitBase propagation across 10 workers** — Each worker dep factory took `(logger, stateReader, identity, ...)` and built its own `*deps.BaseDependencies`. The supervisor's framework-metrics injection writes to the dep struct returned by `GetDependenciesAny()` — the factory's instance — so the framework-internal `*BaseDependencies` the worker still held was dead state. Factories now take `bd *deps.BaseDependencies`; the worker constructor captures `bd := w.InitBase(...)` and passes it in. Helloworld already used this shape (`workers/example/helloworld/dependencies.go:39`); the other ten now match. Workers touched: `communicator`, `persistence`, `transport`, `transport/push`, `transport/pull`, `examplechild`, `examplefailing`, `examplepanic`, `exampleparent`, `exampleslow`. `application` is unaffected (`TDeps = struct{}`, overrides `GetDependenciesAny()` to nil).

**`ParentUserSpec` fold into `ExampleparentConfig`** — exampleparent was the lone example worker still carrying a `XxxUserSpec` parse type after L2a's UserSpec→Config rename. Both types embedded `BaseUserSpec`; `ParentUserSpec` carried `ChildWorkerType`, `ChildConfig`, `ChildrenCount` — the parse target at `worker.go:97` read them, built an `ExampleparentConfig` in-memory, and discarded `ParentUserSpec`. The fold moves the three fields onto `ExampleparentConfig`, renames the in-memory `ChildCount` to `ChildrenCount` (user-facing YAML key is `children_count` per `worker_test.go`, `examples/simple.go`, integration tests), moves `GetChildWorkerType()` onto the config, and deletes `userspec.go`.

## Bonus fixes

- **Persistence factory shadowed-ok bug** — `BindDeps` outer `if raw, ok := params["dependencies"]; ok`
  was shadowing the inner `typed, ok := raw.(*PersistenceDependencies)` — the inner ok was never
  evaluated as false, swallowing wrong-type panics into a silent nil-deps path.
- **`ERROR_DOT_ERROR` sentry invariant** — `collector.go`'s transient-error log used
  `deps.String("error", err.Error())`; replaced with `deps.Err(err)` to preserve the error chain
  for Sentry grouping. Caught by the sentry architecture test.
- **Racy panic-test assertion** — `time.Sleep(200ms) + Expect(callCount > 1)` replaced with
  `Eventually(callCount.Load, 3s, 50ms).Should(BeNumerically(">", int32(1)))` so the test waits
  for the post-panic recovery tick instead of asserting before the goroutine scheduler fires.
- **Fence-symbol comment regression** — self-review subagent re-introduced `IsBeingRemoved` by
  name in a prose comment in `worker_snapshot_test.go`; removed.
- **Stale godoc + planning shorthand** — `transport/` and `supervisor/` comments cleaned of
  planning shorthand (`TODO(rename)`, mid-sentence dashes).
- **No-deps footgun documented; closes `TODO(PR2)`** — `WorkerBase[..., struct{}].GetDependenciesAny()`
  returns a non-nil `any` wrapping `struct{}{}`. Test added to `worker_base_test.go` documenting
  this behaviour; workers requiring true nil must override `GetDependenciesAny()`.
  ApplicationWorker is the first no-deps production worker in this PR; it overrides
  `GetDependenciesAny()` to return `nil`, satisfying the TODO's pre-condition ("Resolve before
  shipping the first NoDeps production worker") in `supervisor/api.go`.

## Forward-deletion signaling carried from L1

The following deprecated symbols survive this PR unchanged; callers must not be added:

| Symbol | Deprecated since | Target layer |
|--------|-----------------|--------------|
| `WorkerSnapshot.IsShutdownRequested` | L1 | renamed `IsBeingRemoved` in L9 |
| `WorkerSnapshot.ParentMappedState` | L1 | deleted in L5 |
| `MappedParentStateProvider`, `setMappedParentState()`, `applyStateMapping()` | L1 | deleted in L5 |

## Deferred

| Item | Target | Reason deferred |
|------|--------|-----------------|
| `IsBeingRemoved` rename (from `IsShutdownRequested`) | L9 | Rename is cosmetic until L9 reconciles full lifecycle vocabulary; churn now gains nothing. |
| `ParentMappedState` + mapping functions deletion | L5 | Requires `applyStateMapping()` removal in supervisor; designed as L5 scope. |
| `WrapStatus`, `SetChildSpecsFactory`, `register.Worker[` | L5+ | Not needed until children-v2 API ships. |

## Stack position

```
origin/staging
   ↓
feature/wapi-L0-strip-arch-validators  (PR #2565 — L0)
   ↓
feature/wapi-L1-clean  (PR #2566 — L1)
   ↓
feature/wapi-L1.5-shutdown  (PR #2571 — L1.5)
   ↓
feature/wapi-L2a  (THIS PR)
```


umh-core/pkg/fsmv2/supervisor/api.go
```diff
@@ -339,25 +339,27 @@ func (s *Supervisor[TObserved, TDesired]) AddWorker(identity deps.Identity, work
...
+				// startup noise (persistence.ErrNotFound is the store-level sentinel).
+				if errors.Is(err, persistence.ErrNotFound) {
+					return nil, fsmv2.ErrNoDesiredState
 				}
 
-				return nil
+				s.logger.SentryWarn(deps.FeatureFSMv2, identity.HierarchyPath, "desired_state_load_failed",
+					deps.Err(err))
+
+				return nil, err
```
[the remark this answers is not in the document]
JeremyTheocharis: Correction to the earlier message: the `SentryWarn`-only behavior at this site is intentional. `ErrNotFound` returns `ErrNoDesiredState` to skip the cycle silently; all other errors call `SentryWarn` once per occurrence. No follow-up ticket, no Loki-only `Warn` method planned.

@coderabbitai resolve

umh-core/pkg/fsmv2/worker_base_test.go
```diff
@@ -0,0 +1,467 @@
...
+				worker.GetInitialState()
+			}).To(PanicWith(ContainSubstring("no initial state registered for worker type")))
+		})
+
+		It("returns registered state for the worker type", func() {
+			const testWorkerType = "wb-test-registered-type"
+			mockState := &wbMockState{}
+			fsmv2.RegisterInitialState(testWorkerType, mockState)
+			defer fsmv2.ResetInitialStateRegistry()
+
```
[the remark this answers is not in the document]
JeremyTheocharis: Won't fix in this PR. `ResetInitialStateRegistry` is documented as "for testing only" and this is the only spec in the framework package's test binary touching the registry — no production worker `init()` runs here (workers live under `pkg/fsmv2/workers/...`). If a second spec ever needs shared registry state we'd move to Ordered/BeforeAll. Today it's safe.

@coderabbitai resolve

umh-core/pkg/fsmv2/workers/communicator/worker.go
```diff
@@ -85,11 +84,18 @@ func NewCommunicatorWorker(
...
-	}, nil
+	dependencies := NewCommunicatorDependencies(transportParam, bd)
+	w.BindDeps(dependencies)
+
+	return w, nil
+}
+
+// GetDependencies returns the typed CommunicatorDependencies.
+func (w *CommunicatorWorker) GetDependencies() *CommunicatorDependencies {
+	return w.GetDependenciesAny().(*CommunicatorDependencies)
```
led0nk: as flagged by CR this should be guarded

umh-core/pkg/fsmv2/workers/transport/worker.go
```diff
@@ -96,12 +95,19 @@ func NewTransportWorker(
...
 
-	return &TransportWorker{
-		BaseWorker: helpers.NewBaseWorker(dependencies),
-	}, nil
+	return w, nil
+}
+
+// GetDependencies returns the typed TransportDependencies.
+func (w *TransportWorker) GetDependencies() *TransportDependencies {
+	return w.GetDependenciesAny().(*TransportDependencies)
```
led0nk: should as well get a guard here

umh-core/pkg/fsmv2/worker_base.go
```diff
@@ -0,0 +1,207 @@
...
+	w.logger = logger
+	w.initialized = true
+	w.mu.Unlock()
+
+	return bd
+}
+
+// Config returns the TConfig cached from the last DeriveDesiredState call.
+// Returns zero-value TConfig before the first successful DeriveDesiredState.
+func (w *WorkerBase[TConfig, TStatus, TDeps]) Config() TConfig {
```
led0nk, line 78: Config() and ConfigReady() are dead today and ship as a latent footgun.

The base DeriveDesiredState writes w.config at lines 103 and 147, but every real worker (communicator, transport, persistence, application) overrides DeriveDesiredState and re-parses from scratch without calling w.WorkerBase.DeriveDesiredState(spec) first. So w.config is never written for them — Config() returns the zero TConfig forever, ConfigReady() returns false forever.

Workers read config via fsmv2.ExtractConfig[T](desired) from the snapshot instead. Nothing reads Config() today, so the bug is latent. But the next contributor who sees these accessors on WorkerBase and uses one will get silent zero-value config with no panic.

led0nk (changes requested):
the worker_base came from self-review and somehow makes sense to me - dependant on what comes from next PR, this is something i cannot foresee

umh-core/pkg/fsmv2/workers/communicator/worker.go
JeremyTheocharis: Fixed in 0bd9e7f07. Comma-ok + nil-guard pattern matching pull/push/persistence.

umh-core/pkg/fsmv2/workers/transport/worker.go
JeremyTheocharis: Fixed in 0bd9e7f07. Same comma-ok + nil-guard pattern. Also swept the 6 example workers (helloworld, examplechild, examplepanic, examplefailing, exampleparent, exampleslow) for consistency since they had the same unguarded shape.

umh-core/pkg/fsmv2/worker_base.go
JeremyTheocharis: You're right on every point. I subagent-investigated and confirmed:

- Zero production callers of Config() or ConfigReady() anywhere in pkg/.
- All 11 real workers override DeriveDesiredState without calling the base, so w.config is never populated.
- Workers actually read config via fsmv2.ExtractConfig[T](desired) from the snapshot.

Our original L1..L9 re-stack plan already scheduled the deletion for L3's design pass.

To keep L2a's scope tight on the WorkerBase + ShouldStop theme, I went with the in-between option in 0bd9e7f07:

- Added `Deprecated:` godoc to Config() and ConfigReady() pointing at ExtractConfig[T] as the runtime pattern.
- Updated the DeriveDesiredState godoc to drop the misleading "call w.Config() after" guidance.
- Struck Config()/ConfigReady() from the CLAUDE.md "What WorkerBase provides" affordances list with an L3-deletion footnote.

So the footgun is signposted now, full deletion lands in L3. Happy to pull the deletion forward into L2a if you'd rather have it gone — it's ~50 LOC (two methods, one cache field, eight test sites, plus restructuring one InitBase behavioral test that incidentally asserts on ConfigReady).

umh-core/pkg/fsmv2/CLAUDE.md
```diff
@@ -140,7 +186,7 @@ Key types used when building workers:
 - **`WorkerSnapshot[TConfig, TStatus]`**: typed snapshot for state `Next()` methods
 - **`ConvertWorkerSnapshot[TConfig, TStatus]`**: entry-point type assertion in states
 - **`ExtractConfig[TConfig](desired)`**: typed config access in `CollectObservedState`
-**Architecture validators** require `ConvertWorkerSnapshot` as the single entry-point in `Next()` methods; `snap.IsStopRequired()` (WorkerSnapshot) is the canonical shutdown check.
+**Architecture validators** require `ConvertWorkerSnapshot` as the single entry-point in `Next()` methods; `snap.ShouldStop()` (WorkerSnapshot) is the canonical shutdown check.
```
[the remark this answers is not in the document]
JeremyTheocharis: Not making this change — the premise is wrong. `IsShutdownRequested` and `ShouldStop()` are two distinct constructs that coexist on purpose:

- `IsShutdownRequested` is a bool field on `WorkerSnapshot` (see `pkg/fsmv2/api.go` near line 324) carrying the desired-state flag.
- `ShouldStop()` is a method that returns `s.IsShutdownRequested || s.ParentMappedState == "stopped"`.

The architecture validator at `pkg/fsmv2/internal/validator/state.go` explicitly accepts both as valid first conditionals. Real state files use both: e.g. `state_running.go` files typically call `snap.ShouldStop()`, while `state_stopped.go` files read `snap.IsShutdownRequested` directly for re-stay logic.

Substituting one for the other in the invariants table would create a fresh contradiction with `doc.go` and the ~10 state files that read `snap.IsShutdownRequested` directly. The current table entry is correct as written.

led0nk (approved):
looks good now

PR #2571, opened 2026-05-12
FSMv2: graceful shutdown drains workers through the live tick loop

JeremyTheocharis:
# FSMv2: graceful shutdown drains workers through the live tick loop

## Why

FSMv2 shutdown is a sequence: the parent supervisor asks each worker to stop, then ticks each through its shutdown states until the worker reports it can be removed.

That sequence has a bug. `Shutdown()` stops the parent's tick loop. The workers can't advance through their shutdown states without it, so the supervisor waits for them to finish -- but they can't, and `Shutdown()` hangs until `gracefulShutdownTimeout` fires.

The bug isn't visible today. Each `Shutdown()` hangs for ~30 s, but CI's 30-way parallelism spreads those hangs across workers, so no single worker exceeds the suite deadline. Production has no signal handler in `main.go`, so the container is killed before `Shutdown()` runs at all. Sequential `go test` accumulates every hang in one process and blows past the deadline immediately.

This PR:

1. **Fixes the bug**: the tick loop keeps running until every worker has finished its shutdown sequence; only then does the supervisor stop ticking and clean up. Ongoing `CollectObservedState` and action functions don't receive a ctx cancel from the shutdown -- worker code stays simple, but shutdown waits for them to finish. Accepted for now; every production action is bounded.
2. **Adds the signal handler in `cmd/main.go`**: `signal.NotifyContext(..., SIGINT, SIGTERM)` so `docker stop` / kubelet termination reaches `Shutdown()`. `main()` waits for the supervisor to shut down before exit.


led0nk (approved):
Looks good from my perspective, some non-blocking foundings from the review skill:


Concerns (non-blocking)

  1. s.ctx read at lifecycle.go:284 without ctxMu — supervisor.go:171-174 documents ctxMu protects ctx, but requestShutdown(s.ctx, ...) reads unlocked. Benign (s.ctx never overwritten after Start) but inconsistent with stated invariant.
  2. Drain loop has no ctx.Done() escape (lifecycle.go:296-320) — second SIGTERM during drain has no fast-exit; process waits full gracefulShutdownTimeout. Intentional per design (Phase 4 cancels ctx, not Phase 3), but worth a comment.
  3. Cascade race window — parent tick() at reconciliation.go:903 sets childrenSpecs=nil and calls reconcileChildren(nil) after ShutdownRequested==true. Cascade in Shutdown() already invoked child.Shutdown() synchronously. Child's Shutdown() is documented idempotent (lifecycle.go:189-215), so no panic — verified safe.

JeremyTheocharis:
## Shutdown behaviour experiment on `sha-e0b4b55`

A two-hypothesis acceptance experiment for the operator-visible behaviours this PR claims.

### Hypotheses
- **H1:** Stopping the application (e.g. `docker stop`) causes a clean shutdown — every supervisor in the hierarchy drains, then the process exits with code 0.
- **H2:** A force stop (a second SIGTERM while shutdown is already in progress) makes every supervisor abort its drain immediately, instead of waiting out each level's 5 s graceful timeout.

### Method
Container with `USE_FSMV2_TRANSPORT=true`, unreachable API_URL (TEST-NET-1), `LOGGING_LEVEL=DEBUG`. Hierarchy spawned: `application → communicator → transport → {push, pull}` plus sibling `persistence`. Two trials:

- **Trial A (H1):** one `docker kill -s SIGTERM` → measure time to exit and check the FSMv2 log lines under `/data/logs/umh-core/current`.
- **Trial B (H2):** two SIGTERMs ~395 ms apart, delivered directly to the umh-core process inside the container (`docker exec ... kill -SIGTERM $(pgrep -f /usr/local/bin/umh-core)`). Required because s6-overlay (PID 1) drops the second signal — see "Side note" below.

Log lines we look for in `/data/logs/umh-core/current`:
- `supervisor_shutting_down` — INFO line each supervisor emits at the start of its `Shutdown()`. One per supervisor.
- `graceful_shutdown_force_exit` — WARN line a supervisor emits when its drain loop observes the closed `forceExit` channel and breaks out early.

### Evidence

**Trial A.** Drain-to-exit: 33.8 s. Exit code: 0. `supervisor_shutting_down`: 6 (one per active supervisor — nothing skipped). `graceful_shutdown_force_exit`: 0. Cascade follows the documented 5 s-per-level pattern.

**Trial B.** All six supervisors emit `graceful_shutdown_force_exit` within 817 µs of each other:
```
12:49:13.754687  graceful_shutdown_force_exit (push)
12:49:13.754872  graceful_shutdown_force_exit (pull)
12:49:13.755186  graceful_shutdown_force_exit (transport)
12:49:13.755195  graceful_shutdown_force_exit (communicator)
12:49:13.755497  graceful_shutdown_force_exit (persistence)
12:49:13.755504  graceful_shutdown_force_exit (application)
```
Each event reports `remaining_worker_count=1` — i.e. drain was actively interrupted, not coincidentally finished.

### Conclusion
- **H1 supported.**
- **H2 supported, with one operator-facing caveat.** Force-stop works correctly when the second signal reaches the umh-core process. Under s6-overlay (current container PID 1), `docker kill -s SIGTERM` twice in a row results in only one signal reaching umh-core; the second is dropped by s6. Pre-L1.5 had no second-signal path at all, so this isn't a regression — but the PR text should call it out, and we may want a follow-up that wires an s6 `finish` hook (or equivalent) so repeat signals propagate through.

### Side note
Phase 2C (`parent.mu` released around `child.Shutdown()`) is *not* covered by this experiment. The refactor is correct by code inspection of `reconciliation.go`, but no runtime scenario here exposes the lock-release benefit. Suggest framing it as "architecturally verified" rather than "empirically validated" in the PR description.

PR #2581, opened 2026-05-19
[ENG-4970] DFC cannot be reconfigured after failed initial deployment

rpaland:
## Problem
When a bridge was created but deployment failed, it was partially saved (connection only). With the write flows enabled, there is no option in the read flow anymore to select a protocol, leaving the bridge in a stuck state.
We could not just show empty input, processing and output fields, since relative to the protocol (opcua, s7comm, modbus) the ui looks different.

## Changes
The protocol and processing mode and now stored on the first deployment step, so users are able to re-prefill the UI afterwards.


rpaland:
We will implement that the whole config is always saved, even when deployment fails and therefor closing this.

PR #2582, opened 2026-05-20
Eng 4876 processing for write flows only nodered js

FalconTube:
> [!NOTE]
> Use frontend branch https://github.com/united-manufacturing-hub/ManagementConsole/pull/3180 for testing this PR. Won't work with current staging/main due to breaking changes.

## Product wise

Rework write flow yaml in this format. Naming and smaller adjustments will be done in follow up PR.

```
dataflowcomponent_write:
    input_topics: |-
        umh.v1.UMHDEVTEST._historian.my_data_generate0
        umh.v1.UMHDEVTEST._historian.my_data_generate1
        umh.v1.UMHDEVTEST._historian.my_data_generate2
    processing_nodered_js: |-
        msg.payload.foo = "bar";
        return msg;
    output:
        http_client:
            url: http://{{ .IP }}:{{ .PORT}}/get
                  verb: GET
```

This not makes it that
1. topics are not a user variable any more, but directly wired into write flow
2. processing is fixed to `nodered_js` (will have a type in follow up PR)
3. All of the parameters are now strings and support templating. Sent in requests like this and then parsed + rendered into types in umh-core

## Code wise
  Write flows stored input topics in Variables.User["UMH_TOPICS"] (untyped map[string]any), requiring three-form type
  conversion to survive YAML round-trips. This PR moves to a typed DataflowComponentWriteConfigInput struct with an
  input_topics string field.

  Changes:
  - Wire format is now input_topics (string) + write_output (map), replacing the old benthos.input/output nesting.
  - WriteDFCRequest and WriteDFCResponse collapsed into a single WriteDFC type — they were structurally identical.
  - ProtocolConverterRequest removed; ProtocolConverter now used for both deploy/edit actions and GET responses.
  - toStringSlice(), equalTopicSets(), and all UMH_TOPICS handling deleted.
  - Write DFC no longer borrows the read DFC's benthos config shape.
  - Validation split into validateReadProtocolConverterDFC / validateWriteDFCConfig. Template vars in input_topics (e.g., {{
   .location_path }}) are allowed at validation time; rendering happens in BuildRuntimeConfig.
  - Example config: dataflowcomponent_write.write_output instead of dataflowcomponent_write.benthos.output.
  - CHANGELOG updated to reference input_topics instead of UMH_TOPICS.

### DIFF
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -4,7 +4,9 @@
 
 ### Preview: Write Flows
 
-- The write flow wire format now uses a typed `input_topics` field (string or Go template) and a `write_output` block, replacing the old `benthos.input`/`benthos.output` nesting. The `UMH_TOPICS` user variable is no longer used.
+- The write flow wire format now uses a typed `input_topics` field (string or Go template) and an `output` block, replacing the old `benthos.input`/`benthos.output` nesting. The `UMH_TOPICS` user variable is no longer used.
+
+  **Migration (internal staging users with write flows enabled):** Existing write flows that used the old `benthos.{input,output}` shape are not automatically converted. Re-create any write flow via the UI using the new format. There is no data loss — only the write-flow configuration is affected; read flows and connection settings are preserved.
 
 ## [0.44.21]
 
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -4,9 +4,7 @@
 
 ### Preview: Write Flows
 
-- The write flow wire format now uses a typed `input_topics` field (string or Go template) and an `output` block, replacing the old `benthos.input`/`benthos.output` nesting. The `UMH_TOPICS` user variable is no longer used.
-
-  **Migration (internal staging users with write flows enabled):** Existing write flows that used the old `benthos.{input,output}` shape are not automatically converted. Re-create any write flow via the UI using the new format. There is no data loss — only the write-flow configuration is affected; read flows and connection settings are preserved.
+- Write flows (preview) now take their input topics directly on the flow, with Go template support for values like `{{ .location_path }}`. Write flows configured on this preview before this release are not carried over and need to be re-created from the UI; read flows and connection settings are unaffected
 
 ## [0.44.21]
 
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Preview: Write Flows
+
+- The write flow wire format now uses a typed `input_topics` field (string or Go template) and a `write_output` block, replacing the old `benthos.input`/`benthos.output` nesting. The `UMH_TOPICS` user variable is no longer used.
+
 ## [0.44.21]
 
 ### Fixes
@@ -41,7 +45,7 @@ Enables FSMv2 communicator by default.
 
 ### Preview: Write Flows
 
-- Previously, write flows used raw `input` for benthos. We now define a typed `input_topics` field and automatically generate the proper `input` in umh-core, which provides an improved user experience and less surface for errors.
+- Previously, write flows used raw `input` for benthos. We now define a user variable `UMH_TOPICS` and automatically generate the proper `input` in umh-core, which provides an improved user experience and less surface for errors.
 
 ## [0.44.17]
  
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -41,7 +41,7 @@ Enables FSMv2 communicator by default.
 
 ### Preview: Write Flows
 
-- Previously, write flows used raw `input` for benthos. We now define a user variable `UMH_TOPICS` and automatically generate the proper `input` in umh-core, which provides an improved user experience and less surface for errors.
+- Previously, write flows used raw `input` for benthos. We now define a typed `input_topics` field and automatically generate the proper `input` in umh-core, which provides an improved user experience and less surface for errors.
 
 ## [0.44.17]
  
--- a/umh-core/pkg/communicator/actions/edit-protocolconverter.go
+++ b/umh-core/pkg/communicator/actions/edit-protocolconverter.go
@@ -149,6 +149,8 @@ func NewEditProtocolConverterAction(userEmail string, actionUUID uuid.UUID, inst
 // Parse implements the Action interface by extracting the protocol converter UUID and
 // dataflow component configuration from the payload.
 func (a *EditProtocolConverterAction) Parse(payload interface{}) error {
+	a.ignoreHealthCheck = false
+
 	// Parse the payload directly as a complete ProtocolConverter object
 	pcPayload, err := ParseActionPayload[models.ProtocolConverter](payload)
 	if err != nil {
@@ -187,6 +189,8 @@ func (a *EditProtocolConverterAction) Parse(payload interface{}) error {
 		input := pcPayload.WriteDFCPayload.DataflowComponentWriteConfigInput
 		a.writeDFCInput = &input
 		a.writeDFCState = pcPayload.WriteDFCPayload.State
+		// OR-merge: if either DFC requests skipping errors, skip the health-check wait
+		// for the entire PC. See deploy-protocolconverter.go for the rationale.
 		if pcPayload.WriteDFCPayload.IgnoreErrors != nil {
 			a.ignoreHealthCheck = a.ignoreHealthCheck || *pcPayload.WriteDFCPayload.IgnoreErrors
 		}
@@ -222,8 +226,9 @@ func (a *EditProtocolConverterAction) Validate() error {
 		return err
 	}
 
-	// Validate read DFC state
-	if a.readDFCSvcCfg != nil && a.readDFCState != "" {
+	// Validate read DFC state — validate independently of whether config was provided,
+	// so state-only edits (readDFCSvcCfg == nil) are also checked.
+	if a.readDFCState != "" {
 		if err := ValidateDataFlowComponentState(a.readDFCState); err != nil {
 			return fmt.Errorf("invalid read DFC state: %w", err)
 		}
@@ -389,8 +394,10 @@ func (a *EditProtocolConverterAction) applyMutation() (config.ProtocolConverterC
 		instanceToModify.ProtocolConverterServiceConfig.Config.DataflowComponentReadServiceConfig = *a.readDFCSvcCfg
 	}
 
-	// Apply write DFC config if provided — store the original input (template string preserved).
-	if a.writeDFCInput != nil && a.writeDFCInput.HasOutput() {
+	// Apply write DFC config when the Output field was explicitly included in the payload
+	// (non-nil, even if empty). A nil Output means a state-only change; the existing
+	// config is preserved. An explicitly empty Output ({}) clears the write DFC output.
+	if a.writeDFCInput != nil && a.writeDFCInput.Output != nil {
 		instanceToModify.ProtocolConverterServiceConfig.Config.DataflowComponentWriteServiceConfig = *a.writeDFCInput
 	}
 
@@ -935,7 +942,7 @@ func (a *EditProtocolConverterAction) renderDesiredDFCConfig(pcSnapshot *protoco
 		modifiedSpec,
 		agentLocation,
 		nil,             // TODO: add global vars
-		"unimplemented", // Must stay "unimplemented": FSM uses the same sentinel, so bridged_by values match. If FSM changes its node-name source, update here in lockstep.
+		runtime_config.BridgedByPlaceholder,
 		pcName,
 	)
 	if err != nil {
@@ -992,21 +999,34 @@ func (a *EditProtocolConverterAction) deriveDFCType() DFCType {
 		return dfcTypeFromPresence(hasRead, hasWrite)
 	}
 
-	// Connection IP/PORT changes affect all DFCs since they use {{ .IP }}/{{ .PORT }} templates.
-	// So if a change is detected, all present DFCs need redeploy.
-	// Use comma-ok map lookups: if IP/PORT keys are absent (e.g. PCs created
-	// before these variables were standard), skip the connection-change check
-	// rather than treating it as changed. Without this, fmt.Sprint(nil) returns
-	// "<nil>" which never equals any real value, forcing a spurious redeploy.
+	// Any variable change (IP, PORT, or custom vars like baudRate) forces redeploy of all
+	// present DFCs because template rendering uses all variables. Keys absent in the deployed
+	// config are skipped — treating nil as changed would cause spurious redeploys on PCs
+	// created before a variable was standard (fmt.Sprint(nil) → "<nil>" ≠ any real value).
 	deployedVars := currentPC.ProtocolConverterServiceConfig.Variables.User
-	deployedIP, ipOk := deployedVars["IP"]
-	deployedPort, portOk := deployedVars["PORT"]
 
-	connectionChanged := ipOk && portOk &&
-		(fmt.Sprint(deployedIP) != a.connectionIP || fmt.Sprint(deployedPort) != a.connectionPort)
+	incomingVars := make(map[string]any, len(a.templateVars)+2)
+	for _, v := range a.templateVars {
+		incomingVars[v.Label] = v.Value
+	}
+	if a.connectionIP != "" {
+		incomingVars["IP"] = a.connectionIP
+	}
+	if a.connectionPort != "" && a.connectionPort != "0" {
+		incomingVars["PORT"] = a.connectionPort
+	}
+
+	connectionChanged := false
+	for k, incomingVal := range incomingVars {
+		if deployedVal, ok := deployedVars[k]; ok {
+			if fmt.Sprint(deployedVal) != fmt.Sprint(incomingVal) {
+				a.actionLogger.Debugf("Variable %q changed (%v → %v), all present DFCs need redeploy", k, deployedVal, incomingVal)
+				connectionChanged = true
+				break
+			}
+		}
+	}
 	if connectionChanged {
-		a.actionLogger.Debugf("Connection changed (IP or PORT), all present DFCs need redeploy")
-
 		return dfcTypeFromPresence(hasRead, hasWrite)
 	}
 
@@ -1056,7 +1076,28 @@ func (a *EditProtocolConverterAction) writeDFCConfigDiffers(
 	if incomingState != "" && deployedState != "" && incomingState != deployedState {
 		return true
 	}
-	return !reflect.DeepEqual(incoming, deployedConfig)
+	// Normalize nil maps to empty maps before comparison: YAML round-trip converts
+	// nil maps (Output/Buffer) to empty maps, and reflect.DeepEqual treats them as
+	// different, causing spurious drift detection on every reconcile.
+	return !reflect.DeepEqual(normalizeWriteConfigInput(incoming), normalizeWriteConfigInput(deployedConfig))
+}
+
+// normalizeWriteConfigInput replaces nil map fields with empty maps so that
+// reflect.DeepEqual comparisons are not tripped up by YAML round-trip nil→{} coercions.
+func normalizeWriteConfigInput(c dataflowcomponentserviceconfig.DataflowComponentWriteConfigInput) dataflowcomponentserviceconfig.DataflowComponentWriteConfigInput {
+	if c.Output == nil {
+		c.Output = map[string]any{}
+	}
+	if c.Buffer == nil {
+		c.Buffer = map[string]any{}
+	}
+	// Mirror the default applied by ToDataflowComponentServiceConfig so that a config
+	// stored without an explicit JS snippet compares equal to one that was just parsed
+	// from a payload carrying the default "return msg;".
+	if c.ProcessingNoderedJS == "" {
+		c.ProcessingNoderedJS = "return msg;"
+	}
+	return c
 }
 
 // getUserEmail implements the Action interface by returning the user email associated with this action.
--- a/umh-core/pkg/communicator/actions/edit-protocolconverter.go
+++ b/umh-core/pkg/communicator/actions/edit-protocolconverter.go
@@ -34,6 +34,7 @@ import (
 	"errors"
 	"fmt"
 	"maps"
+	"reflect"
 	"slices"
 	"strconv"
 	"time"
@@ -41,8 +42,6 @@ import (
 	"github.com/google/uuid"
 	"github.com/tiendc/go-deepcopy"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/config"
-	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/config/benthosserviceconfig"
-	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/config/connectionserviceconfig"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/config/dataflowcomponentserviceconfig"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/config/protocolconverterserviceconfig"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/constants"
@@ -106,14 +105,14 @@ type EditProtocolConverterAction struct {
 	readDFCState   string // desired state for the read DFC ("active" or "stopped"; empty = default active)
 	writeDFCState  string // desired state for the write DFC ("active" or "stopped"; empty = default active)
 
-	readDFCPayload  *models.CDFCPayload
-	writeDFCPayload *models.CDFCPayload
+	templateVars []models.ProtocolConverterVariable
 
-	vb []models.ProtocolConverterVariable
-
-	// Desired DFC configs for comparison during health checks
-	desiredReadDFCConfig  dataflowcomponentserviceconfig.DataflowComponentServiceConfig
-	desiredWriteDFCConfig dataflowcomponentserviceconfig.DataflowComponentServiceConfig
+	// readDFCSvcCfg is the built+validated read DFC service config, set in Parse.
+	// nil means no read DFC was provided in the request.
+	readDFCSvcCfg *dataflowcomponentserviceconfig.DataflowComponentServiceConfig
+	// writeDFCInput is the raw template-string form of the write config, persisted to the spec.
+	// nil means no write DFC was provided.
+	writeDFCInput *dataflowcomponentserviceconfig.DataflowComponentWriteConfigInput
 
 	actionUUID   uuid.UUID
 	instanceUUID uuid.UUID
@@ -125,6 +124,10 @@ type EditProtocolConverterAction struct {
 	atomicEditUUID uuid.UUID
 
 	ignoreHealthCheck bool
+	// lastRenderErr holds the most recent renderDesiredDFCConfig error so the
+	// awaitRollout timeout message can surface the real cause instead of just
+	// "did not become active in time".
+	lastRenderErr error
 }
 
 // NewEditProtocolConverterAction returns an un-parsed action instance.
@@ -147,7 +150,7 @@ func NewEditProtocolConverterAction(userEmail string, actionUUID uuid.UUID, inst
 // dataflow component configuration from the payload.
 func (a *EditProtocolConverterAction) Parse(payload interface{}) error {
 	// Parse the payload directly as a complete ProtocolConverter object
-	pcPayload, err := ParseActionPayload[models.ProtocolConverter](payload)
+	pcPayload, err := ParseActionPayload[models.ProtocolConverterRequest](payload)
 	if err != nil {
 		return fmt.Errorf("failed to parse protocol converter payload: %w", err)
 	}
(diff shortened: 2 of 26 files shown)


umh-core/pkg/config/protocolconverterserviceconfig/generator.go
```diff
@@ -153,3 +159,19 @@ func normalizeConfig(raw map[string]any) map[string]any {
...
+		zap.S().Errorf("writeConfigToMap: failed to marshal write DFC config, write side will be empty in YAML output: %v", err)
+		return map[string]any{}
+	}
+	var m map[string]any
+	if err := yaml.Unmarshal(data, &m); err != nil {
+		zap.S().Errorf("writeConfigToMap: failed to unmarshal write DFC config, write side will be empty in YAML output: %v", err)
+		return map[string]any{}
+	}
+	return m
+}
```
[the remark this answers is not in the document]
FalconTube: resolved

umh-core/pkg/config/protocolconverterserviceconfig/generator.go
```diff
@@ -46,6 +47,7 @@ func (g *Generator) RenderConfig(cfg ProtocolConverterServiceConfigSpec) (string
 	}
 
 	yamlStr := string(yamlBytes)
+	zap.S().Errorf("rendered ProtocolConverter config: %s", yamlStr)
```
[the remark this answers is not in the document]
FalconTube: resolved

umh-core/pkg/config/protocolconverterserviceconfig/comparator.go
```diff
@@ -58,7 +58,7 @@ func (c *Comparator) ConfigsEqual(desired, observed ProtocolConverterServiceConf
 	return connectionEqual &&
 		locationEqual &&
 		comparatorDFC.ConfigsEqual(dfcReadD, dfcReadO) &&
-		comparatorDFC.ConfigsEqual(dfcWriteD, dfcWriteO) &&
+		reflect.DeepEqual(dfcWriteD, dfcWriteO) &&
 		comparatorVariable.ConfigsEqual(desired.Variables, observed.Variables) &&
```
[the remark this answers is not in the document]
FalconTube: non-issue and will touch this in foolow up PR again

umh-core/pkg/communicator/actions/get-protocolconverter.go
```diff
@@ -175,6 +178,67 @@ func determineProtocol(readDFC *models.ProtocolConverterDFC) string {
...
+			}
+		}
+	}
+
+	templateInfo = &models.ProtocolConverterTemplateInfo{
+		IsTemplated: isTemplated,
+		Variables:   variables,
+		RootUUID:    uuid.Nil,
+	}
+	return
```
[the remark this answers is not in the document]
FalconTube: existing issue, will not fix here

umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -433,52 +382,20 @@ func (a *EditProtocolConverterAction) applyMutation(readBenthosConfig, writeBent
...
-	if a.writeDFCPayload != nil {
-		if len(a.writeDFCPayload.UMHTopics) > 0 {
-			instanceToModify.ProtocolConverterServiceConfig.Variables.User["UMH_TOPICS"] = a.writeDFCPayload.UMHTopics
-		} else {
-			delete(instanceToModify.ProtocolConverterServiceConfig.Variables.User, "UMH_TOPICS")
-		}
+	// Apply write DFC config if provided — store the original input (template string preserved).
+	if a.writeDFCInput != nil {
+		instanceToModify.ProtocolConverterServiceConfig.Config.DataflowComponentWriteServiceConfig = *a.writeDFCInput
 	}
```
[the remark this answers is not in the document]
FalconTube: resolved

JeremyTheocharis (commented):
I went through the PR using Klaus and the self-review skill, plus an experiment to check whether it works end-to-end. During self-review and the experiment a couple of points came up; listed inline below.

CodeRabbit's open thread was pre-audited so items you already addressed are not repeated.

The happy path works. Exp1: umh-core accepted the new `writeDFC` wire shape end-to-end (action handler logged "write DFC configuration verified"). Exp2: the template literal `{{ .location_path }}` is preserved through the FE → MC API → BE → re-display round-trip without being rendered; the topic editor auto-switches to Code mode on template syntax.

4 Critical, 11 Important, 6 nits below. Three findings (one Important + two HIGH) share one root cause: no boundary does strict decode (config.yaml load, action payload, GET response). A single follow-up adding strict decode + redaction at those boundaries closes all three.

Two CodeRabbit findings deferred to a "follow-up PR" (`comparator.go:61` reflect.DeepEqual on map[string]any; `get-protocolconverter.go:230-240` synthetic-var leak): could you link Linear tickets before merge? Klaus reproduced both as concrete failure modes; without a ticket they regress.

Two extra findings from the experiment (bottom of the inline list): duplicate `edit-protocol-converter` push on save, and an orphaned `writeFlowEnabled` $state in MC#3180.

umh-core/pkg/config/dataflowcomponentserviceconfig/write_config.go
```diff
@@ -0,0 +1,216 @@
...
+func (c DataflowComponentWriteConfigInput) ToDataflowComponentServiceConfig(bridgedBy string) DataflowComponentServiceConfig {
+	return c.ToWriteConfig().ToDataflowComponentServiceConfig(bridgedBy)
+}
+
+// Render executes InputTopics as a Go template with scope, splits the result
+// into individual topic strings, and returns a fully typed DataflowComponentWriteConfig.
+func (c DataflowComponentWriteConfigInput) Render(scope map[string]any) (DataflowComponentWriteConfig, error) {
+	rendered := c.InputTopics
+	if strings.Contains(c.InputTopics, "{{") {
+		tpl, err := template.New("input_topics").Option("missingkey=zero").Parse(c.InputTopics)
```
JeremyTheocharis: Klaus flagged (CRITICAL): `Render()` uses `missingkey=zero`; production rendering at `templating.go:228` uses `missingkey=error`. A typo in a Connection variable (`{{ .location_pathh }}`) renders to `<no value>` here; `splitTopics` preserves the malformed string; the bridge then subscribes to a broken UNS path. Nothing logs the divergence.

Secondary: `Render()` has zero non-vendor callers (grep). If you keep it, match the policy in `templating.go`. Otherwise delete it so a future caller cannot pick up the lax variant by accident.

umh-core/pkg/communicator/actions/get-protocolconverter.go
```diff
@@ -175,6 +178,67 @@ func determineProtocol(readDFC *models.ProtocolConverterDFC) string {
...
+			ip = valueStr
+		case "Port", "port", "PORT":
+			if n, err := strconv.ParseUint(valueStr, 10, 32); err == nil {
+				port = uint32(n)
+			} else {
+				logger.Warnw("Failed to parse port from variable", "port", valueStr, "error", err)
+			}
+		}
+	}
+	isTemplated = len(variables) > 0
```
JeremyTheocharis: Klaus flagged (CRITICAL): `isTemplated = len(variables) > 0` is now always true. `deploy-protocolconverter.go:266-267` injects `IP` and `PORT` into `Variables.User` for every PC unconditionally, so the flag cannot distinguish a true template instance from a stand-alone bridge with auto-injected connection vars.

This concretises CodeRabbit's deferred "synthetic-var leak" finding. Could you link a Linear ticket before merge? If the fix is allowlist-based (filter system-injected keys when computing `isTemplated`), it is small enough to land in this PR.

umh-core/pkg/config/dataflowcomponentserviceconfig/write_config.go
```diff
@@ -0,0 +1,216 @@
...
+			return DataflowComponentWriteConfig{}, fmt.Errorf("failed to render input_topics template: %w", err)
+		}
+		rendered = buf.String()
+	}
+	return DataflowComponentWriteConfig{
+		InputTopics:         splitTopics(rendered),
+		ProcessingNoderedJS: c.ProcessingNoderedJS,
+		Output:              c.Output,
+		Buffer:              c.Buffer,
+	}, nil
```
JeremyTheocharis: Klaus flagged (MEDIUM): Dead instance method `Render` (zero non-vendor call sites) exists alongside the live `config.RenderTemplate`. As noted on the Critical, it also uses a different `missingkey` policy. Delete the dead method or unify.

umh-core/pkg/communicator/actions/get-protocolconverter.go
```diff
@@ -244,123 +308,39 @@ func (a *GetProtocolConverterAction) Execute() (interface{}, map[string]interfac
...
 					readDFC, err = buildProtocolConverterDFCFromConfig(readDFCConfig, a)
 					if err != nil {
-						a.actionLogger.Warnf("Failed to build read DFC: %v", err)
-						SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionExecuting,
-							fmt.Sprintf("Warning: Failed to build read DFC for protocol converter '%s': %v", instance.ID, err),
-							a.outboundChannel, models.GetProtocolConverter)
+						errMsg := fmt.Sprintf("Failed to build read DFC for protocol converter '%s': %v", instance.ID, err)
+						SendActionReply(a.instanceUUID, a.userEmail, a.actionUUID, models.ActionFinishedWithFailure,
+							errMsg, a.outboundChannel, models.GetProtocolConverter)
+						return nil, nil, fmt.Errorf("%s", errMsg)
```
JeremyTheocharis, line 348: Klaus flagged (MEDIUM): Read DFC build failure now returns `ActionFinishedWithFailure` and aborts the whole GET. Customer loses visibility of write DFC, template info, connection info whenever the read DFC is malformed, even though the write side might still be deployable. Consider returning a partial payload with the read-DFC error attached, not a hard abort.

umh-core/pkg/communicator/actions/protocolconverter_helpers.go
```diff
@@ -0,0 +1,168 @@
...
+	if cfg != nil && cfg.HasOutput() && strings.TrimSpace(cfg.InputTopics) == "" {
+		return errors.New("write DFC requires at least one input topic (input_topics)")
+	}
+	if state != "" {
+		if err := ValidateDataFlowComponentState(state); err != nil {
+			return fmt.Errorf("invalid write DFC state: %w", err)
+		}
+	}
+	return nil
+}
```
JeremyTheocharis, line 55: Klaus flagged (MEDIUM): Two things on the comparator/validation surface that you flagged for "follow-up":

1. `validateWriteDFCConfig` only checks `strings.TrimSpace(InputTopics) != ""`. A malformed Go template (e.g. `umh.v1.{{ .IP`, unclosed action) passes validation, gets persisted, then fails at `BuildRuntimeConfig` render time; `lastRenderErr` is logged only on rollback. A `template.New().Parse(InputTopics)` call at validation time catches this at the store boundary instead of the deploy boundary.

2. `comparator.go:61` `reflect.DeepEqual` on `DataflowComponentWriteConfigInput`: nil-map vs empty-map divergence after YAML round-trip (`Output: nil` vs `Output: map[string]any{}`) will cause spurious "drift detected" → constant FSM reconcile churn on any write PC that round-trips through config save.

Both of these need Linear tickets if deferred (#1 is small enough to land here).

umh-core/pkg/config/dataflowcomponentserviceconfig/write_config.go
```diff
@@ -0,0 +1,216 @@
...
+	var input map[string]any
+	if len(c.Output) > 0 {
+		umhTopics := c.InputTopics
+		if len(umhTopics) == 0 {
+			// Sentinel: action validation normally rejects an empty InputTopics when
+			// Output is present, but set-config-file bypasses actions and can produce
+			// this state. The sentinel makes the misconfiguration visible in Benthos
+			// logs rather than silently subscribing to nothing.
+			umhTopics = []string{"TOPIC_NOT_SET_BY_USER"}
+		}
```
JeremyTheocharis, line 142: Klaus flagged (MEDIUM): `splitTopics(",,,,")` returns `[]string{}` after the validator passes (validator only rejects whitespace-only input). `ToDataflowComponentServiceConfig` then injects the `TOPIC_NOT_SET_BY_USER` sentinel with no `SentryWarn`. Per the umh-core sentry-routing convention, this should be a SentryWarn at sentinel-injection time so we see it in production.

umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -1028,7 +934,7 @@ func (a *EditProtocolConverterAction) renderDesiredDFCConfig(pcSnapshot *protoco
 		modifiedSpec,
 		agentLocation,
 		nil,             // TODO: add global vars
-		"unimplemented", // TODO: add node name
+		"unimplemented", // Must stay "unimplemented": FSM uses the same sentinel, so bridged_by values match. If FSM changes its node-name source, update here in lockstep.
```
JeremyTheocharis: Klaus flagged (MEDIUM): The `"unimplemented"` literal is duplicated across two packages (here + `umh-core/pkg/fsm/protocolconverter/actions.go:334`) and held together by a 30-word inline comment that says "update in lockstep". Per umh-core CLAUDE.md (no inline comments), this should be one exported const in the FSM package, referenced from both call sites.

Empirical: the literal is rendered into the Kafka `consumer_group` name as `protocol-converter_unimplemented_<bridge>`. Customers see it in Redpanda consumer-group listings and any tooling that introspects consumer groups.

umh-core/pkg/service/protocolconverter/runtime_config/runtime_config.go
```diff
@@ -315,26 +303,26 @@ func renderConfig(
...
-	if unsInput, ok := write.BenthosConfig.Input["uns"].(map[string]any); ok {
-		unsInput["umh_topics"] = scope["UMH_TOPICS"]
+	// Extract the resolved bridged_by value from the scope to wire up the UNS consumer group.
+	// If scope["internal"]["bridged_by"] is absent (e.g. in structural comparisons that never
+	// hit the Benthos wire), bridgedBy stays "" and ToDataflowComponentServiceConfig will
+	// emit consumer_group: "" — acceptable for diff logic, but invalid for a live bridge.
+	// The FSM always injects this key before deploying, so a blank value at runtime is a bug.
+	var bridgedBy string
+	if internal, ok := scope["internal"].(map[string]any); ok {
+		bridgedBy, _ = internal["bridged_by"].(string)
```
JeremyTheocharis, line 324: Klaus flagged (MEDIUM): The comment claims "blank `bridgedBy` at runtime is a bug" but the code accepts the empty value and proceeds. `BuildRuntimeConfig` always sets it before `renderConfig` today, so this is not a current regression. But it is a footgun for the next maintainer: if the injection is forgotten on a new path, every bridge joins the same empty consumer group, causing cross-bridge message theft. Either assert non-empty when `Output != nil`, or pass `bridgedBy` as an explicit parameter so the type system enforces it.

umh-core/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Preview: Write Flows
+
+- The write flow wire format now uses a typed `input_topics` field (string or Go template) and a `write_output` block, replacing the old `benthos.input`/`benthos.output` nesting. The `UMH_TOPICS` user variable is no longer used.
+
 ## [0.44.21]
 
 ### Fixes
```
JeremyTheocharis, line 11: Klaus flagged (MEDIUM): Two CHANGELOG gaps that compound:

1. Entry says `write_output` but the actual YAML tag at `write_config.go:57` is `output`. Customers copy-pasting the example get a broken config with no error.

2. No migration note on the breaking wire-format change. Legacy `dataflowcomponent_write.benthos.{input,output}` configs drop the write flow on in-place upgrade: the new `DataflowComponentWriteConfigInput` only carries `input_topics/processing_nodered_js/output/buffer` YAML tags; strict parse rejects with "field benthos not found", but the `allowUnknownFields=true` fallback in `WriteYAMLConfigFromString` (`manager.go:1105-1113`) discards the error and returns a zero-value struct. On the next save the write DFC is gone with no Sentry event.

Per our prior discussion, breaking the write-flow wire format is fine because write flows are FF-gated (`disableReadFlow` + `disable-read-flow` SupportedFeature) and have no production users yet. But internal staging users with the FF on need to know what happens on upgrade. Call it out in the CHANGELOG with the migration steps (rebuild from `umh.uns` input + `Variables.User.UMH_TOPICS`, or re-create the write flow from scratch via the new UI).

umh-core/pkg/communicator/actions/get-protocolconverter.go
JeremyTheocharis: Klaus flagged (MEDIUM): Legacy `UMH_TOPICS` user-variable on upgraded configs is still shown as user-editable here. Edits to it are no-ops post-PR (no field reads it). Either allowlist what shows up in connection info or do a one-shot migration delete on first PC read.

umh-core/pkg/config/dataflowcomponentserviceconfig/write_config.go
```diff
@@ -0,0 +1,216 @@
...
+
+	return DataflowComponentServiceConfig{
+		BenthosConfig: BenthosConfig{
+			Pipeline: pipeline,
+			Output:   c.Output,
+			Buffer:   c.Buffer,
+		},
+	}
+}
+
```
JeremyTheocharis, line 177: Klaus flagged (MEDIUM): No `write_config_test.go` exists. Zero unit tests for `splitTopics` / `ToWriteConfig` / `ToDataflowComponentServiceConfig` / `HasOutput` / `Render`. Wire-format contract for a breaking change has no unit-level pin. One ~80-line table-driven file covering: comma-vs-newline splitting, comma-inside-template-rendered-topic (sneaky failure mode), CRLF, sentinel branch, dead-method drift, legacy-shape rejection.

umh-core/pkg/models/action_models.go
```diff
@@ -813,15 +812,30 @@ type ProtocolConverterTemplateInfo struct {
 	IsTemplated bool                        `json:"isTemplated" mapstructure:"isTemplated" yaml:"isTemplated"`
 }
 
+// WriteDFCPayload is the wire format for write DFC configuration, used for both inbound
+// (deploy/edit actions) and outbound (GET responses) payloads i.e. what is transmitted to/from the client.
+// The embedded struct uses mapstructure:",squash" so mapstructure flattens its fields
+// into this struct during action payload decoding. Go's encoding/json promotes embedded
+// struct fields automatically without a tag, so no json:",squash" is needed or exists.
+type WriteDFCPayload struct {
+	dataflowcomponentserviceconfig.DataflowComponentWriteConfigInput `mapstructure:",squash"`
```
JeremyTheocharis, line 819: Klaus flagged (HIGH): No `mapstructure.ErrorUnused` on write-DFC payload decoding (this struct + its decoder path). If MC#3180 ships after this PR in production, the old frontend's `writeDFC.benthos.{input,output}` payload decodes to a struct with all fields zero. Action returns success; the next edit writes a zero-value `DataflowComponentWriteConfigInput`. Strict decode (or sniff the legacy shape and reject) turns this into a clean error.

umh-core/pkg/communicator/actions/get-protocolconverter.go
```diff
@@ -175,6 +178,67 @@ func determineProtocol(readDFC *models.ProtocolConverterDFC) string {
...
+// writeDFCFromSpec builds a WriteDFC from the raw spec config (as stored
+// in config.yaml). Returns nil when no output is configured (write DFC not deployed).
+func writeDFCFromSpec(specWrite dataflowcomponentserviceconfig.DataflowComponentWriteConfigInput, state string) *models.WriteDFCPayload {
+	if !specWrite.HasOutput() {
+		return nil
+	}
+	return &models.WriteDFCPayload{
+		DataflowComponentWriteConfigInput: specWrite,
+		State:                             state,
+	}
```
JeremyTheocharis, line 262: Klaus flagged (HIGH): `Output map[string]any` in `WriteDFCPayload` (declared at `write_config.go:57`, embedded at `action_models.go:821`) can hold MQTT/HTTP credentials (`password`, `tls.client_certs[].key`). The GET handler returns the whole map via `writeDFCFromSpec` without redaction. If the `connection-manager` outbound boundary does not already filter sensitive keys, every GET ships credentials to the frontend.

Could you confirm where redaction happens? If nowhere, add an `outputForResponse()` filter at the GET boundary.

JeremyTheocharis:
## Klaus — findings that couldn't anchor inline

Three Criticals (outside the diff range), three cross-repo / unfiled findings, and the nit pile.

### Critical — outside-diff range

**`umh-core/pkg/communicator/actions/deploy-protocolconverter.go:215-258`**

Klaus flagged (CRITICAL): Same finding CodeRabbit raised outside the diff range. `waitForComponentToAppear` is correctly gated by `!a.ignoreHealthCheck`, but the subsequent `SendActionReply("reached expected state")` is unconditional. With `ignoreHealthCheck=true` the user sees a success message even though no convergence check happened.

Either gate the SendActionReply on the same flag, or change the message to "deployed (health-check skipped)".

**`umh-core/pkg/communicator/actions/edit-protocolconverter.go:989-1018` (`deriveDFCType`)**

Klaus flagged (CRITICAL): Same finding CodeRabbit raised outside the diff range. `deriveDFCType` only diffs `IP` and `PORT`. Any other user-defined template variable change (`{{ .baudRate }}`, `{{ .deviceId }}`, custom-template vars) leaves `dfcType=""` and the DFC rollout verification step is skipped. So a write flow edited via template-var change deploys without the verification step.

### Important — line not in diff

**`umh-core/pkg/models/action_models.go:41` (`CDFCPayload.UMHTopics`)**

Klaus flagged (MEDIUM): `CDFCPayload.UMHTopics` field is declared with `json:"umh_topics,omitempty"` but no code reads it. A stale frontend sending `umh_topics` is accepted by decode and then dropped with no log. Either reject at decode time (with strict decode, see the version-skew note on `action_models.go:821`) or remove the field entirely.

### Cross-repo (MC#3180)

**Duplicate `edit-protocol-converter` push on save (confirmed bug)**

Klaus flagged (MEDIUM): Every save of a write flow generates two pushes from MC#3180 to umh-core. Confirmed not different action types: same `actionType=edit-protocol-converter`, same `actionUUID` (`64e68f86-f027-4860-94c8-7cc82ca68445` in the captured pair), byte-identical content. umh-core dedupes by actionUUID so it is not data-corrupting, but the FE save handler is double-firing. Worth a one-line fix in the MC#3180 save call.

**Orphaned feature gate**

Klaus flagged (LOW): `frontend/src/lib/stores/bridge/bridgeContext.svelte.ts:37` declares `writeFlowEnabled = $state(false)` and exports it at line 49, but nothing in the diff ever sets it to true. Either dead code or half-finished feature gating. If gating was the intent, the Write Flow tab is unconditionally exposed today (gated only by the existing `disableReadFlow` FF + `disable-read-flow` SupportedFeature, not by this field).

### Other

**agent.location vs bridge.location merge precedence**

Klaus flagged (LOW-MEDIUM): During the experiment I set bridge location to `{0: site1, 1: line2}`. `agent.location.0` in the host config is `UMH-Systems-GmbH---Dev-Team`. The rendered `{{ .location_path }}` came out as `UMH-Systems-GmbH---Dev-Team.line2`. agent.0 overrode bridge.0; bridge filled in indices the agent didn't define. Per CLAUDE.md "Agent location + bridge location = `{{ .location_path }}`", users likely expect concatenation, not index-wise override.

Could be intentional (multi-tenant: agent enforces enterprise level, bridge can only refine downward). If so, document the precedence rule in CLAUDE.md and in the Bridge location-editor tooltip. If unintentional, fix the merge.

### Nits

- `umh-core/pkg/config/dataflowcomponentserviceconfig/write_config.go:164`: promote `"TOPIC_NOT_SET_BY_USER"` to `const PlaceholderUMHTopicUnset`.
- `umh-core/pkg/models/action_models.go:820-826`: `WriteDFCPayload` naming reads `pc.WriteDFCPayload.DataflowComponentWriteConfigInput`. Rename to `WriteDFC` or split inbound/outbound after MC#3180 stabilizes. Document that `IgnoreErrors` is request-only.
- `umh-core/pkg/communicator/actions/edit-protocolconverter.go:1043, 1049, 1103`: "exposed primarily for testing" docstrings, move to `export_test.go`.
- `umh-core/pkg/config/dataflowcomponentserviceconfig/write_config.go`: "normalise" (British) → "normalize" (US house style).
- PR body + new comments: em-dash sweep; drop "first-class" buzzword.
- `umh-core/pkg/config/dataflowcomponentserviceconfig/write_config.go:107`: defense-in-depth: restrict template scope to `.user.*` for `input_topics` rendering (currently has access to `.internal.bridged_by`).

umh-core/pkg/config/dataflowcomponentserviceconfig/write_config.go
FalconTube: Fixed — `Render()` was dead code (zero non-vendor callers) and has been deleted entirely.

umh-core/pkg/communicator/actions/get-protocolconverter.go
FalconTube: Fixed. `systemInjectedVarKeys` (IP, PORT, location, location_path) are now excluded before computing `isTemplated`. A separate `hasConnectionVars` bool gates the Nmap fallback, so the two concerns no longer share a single flag.

umh-core/pkg/config/dataflowcomponentserviceconfig/write_config.go
FalconTube: Fixed — deleted.

umh-core/pkg/communicator/actions/get-protocolconverter.go
FalconTube: Not changed in this PR. Partial-payload returns on a GET add non-trivial complexity; will revisit when there is a concrete customer report. For now, a malformed read DFC is a configuration error and a hard failure is acceptable.

umh-core/pkg/communicator/actions/protocolconverter_helpers.go
FalconTube: Both items addressed:
1. Template syntax validation removed — write flows are a breaking-change area with no production users, so there is nothing to validate against a legacy shape.
2. nil-map vs empty-map divergence fixed via `normalizeWriteConfigInput`, which normalises `Output`, `Buffer`, and `ProcessingNoderedJS` to their defaults before `reflect.DeepEqual`.

umh-core/pkg/config/dataflowcomponentserviceconfig/write_config.go
FalconTube: Not addressed in this PR. The sentinel already makes the misconfiguration visible in Benthos logs (`TOPIC_NOT_SET_BY_USER` will appear in the consumer subscription). Sentry warn can be added in a follow-up once we decide the right severity level.

umh-core/pkg/communicator/actions/edit-protocolconverter.go
FalconTube: Fixed. `BridgedByPlaceholder` const extracted to the `runtime_config` package and imported from both call sites — no more magic string or lockstep comment.

umh-core/pkg/service/protocolconverter/runtime_config/runtime_config.go
FalconTube: Fixed. `BuildRuntimeConfig` now returns an error when `bridgedBy == ""` and write output is configured, so a future caller that forgets to inject the value fails loudly at the deploy boundary instead of silently joining an empty consumer group.

umh-core/CHANGELOG.md
FalconTube: Both gaps fixed: `write_output` → `output` in the wire-format description, and a migration note added for internal staging users.

umh-core/pkg/communicator/actions/get-protocolconverter.go
FalconTube: Fixed. `UMH_TOPICS` removed from `systemInjectedVarKeys` — it was the legacy write-DFC topic field and is now a no-op, so it should not be shown as user-editable or counted toward `isTemplated`.

umh-core/pkg/config/dataflowcomponentserviceconfig/write_config.go
FalconTube: Fixed. `write_config_test.go` added with table-driven coverage for `splitTopics` (newline, comma, mixed, whitespace, empty, single), `HasOutput` (nil/empty/present), `ToDataflowComponentServiceConfig` (sentinel branch, consumer-group wiring, no-output gate, default nodered_js code), and a round-trip fidelity test.

umh-core/pkg/models/action_models.go
FalconTube: Not addressed — intentionally. Write flows are a breaking-change area; there are no production users and we are not maintaining backward compatibility with the old `benthos.{input,output}` shape. Old configs must be recreated via the UI.

umh-core/pkg/communicator/actions/get-protocolconverter.go
FalconTube: Deferred. Write flows are still a FF-gated preview with no production users. Credential redaction is a valid concern but adds complexity; we will add it when the feature graduates from preview.

umh-core/CHANGELOG.md
JeremyTheocharis: I've applied the new changelog skill and I quite like it. What are your thoughts?

```
Write flows (preview) now take their input topics directly on the flow, with Go template support for values like `{{ .location_path }}`. Write flows configured on this preview before this release are not carried over and need to be re-created from the UI; read flows and connection settings are unaffected
```

i think the rest of it e.g., UMH_TOPICS is not that relevant. If one has to re-create it anyway, then i doesn't matter how it was called before (except when we want to provide a migration guide.

FalconTube: Agreed that this is better, I updated the changelog now

JeremyTheocharis (approved):
LGTM. The wire-format rework is clean, the substantive findings from the earlier review round were addressed, and the changelog reads well. Ship it.

PR #2584, opened 2026-05-22
feat(fsmv2/L4): register.Worker replaces factory.RegisterWorkerAndSupervisorFactoryByType across 12 workers

JeremyTheocharis:
## Why

Workers currently register through `factory.RegisterWorkerAndSupervisorFactoryByType` — each `init()` is a ~20-line block with a worker-factory closure, a supervisor-factory closure, error handling, and an untyped `extraDeps map[string]any` for parent→child dep sharing.

L4 replaces both seams with `register.Worker[TConfig, TStatus, TDeps]` (typed, one-line) and `register.SetDeps[T]` / `GetDeps[T]` (typed deps registry). Each worker's `init()` shrinks to a single line; the untyped extraDeps map disappears.

## Scale

12 worker.go files migrate. 4 new files in `pkg/fsmv2/register/`. Legacy `factory.RegisterWorkerAndSupervisorFactoryByType` survives as internal plumbing inside `register.Worker` (no longer called from worker code). Net ~150 LOC negative.

## What's in this PR

- **Workers register with one `register.Worker[TConfig, TStatus, TDeps]` line** instead of a ~20-line factory closure.
- **Parent→child dep sharing is now typed** via `register.SetDeps[T]` / `GetDeps[T]`. Replaces the untyped `extraDeps map[string]any`.
- **All 12 workers migrated**: helloworld, communicator, persistence, application, transport (+ push + pull), and the 5 example workers.
- **Worker code no longer references `factory.RegisterWorkerAndSupervisorFactoryByType` or the untyped `extraDeps map[string]any`** — both survive as internal plumbing inside `register.Worker`.
- **CSE TypeRegistry wiring** consolidates into the `register` package — registration entry-point is one location.

## Forward-deletion signaling carried from L1/L2a

| Symbol | Deprecated since | Target layer |
|--------|-----------------|--------------|
| `WorkerSnapshot.IsShutdownRequested` | L1 | renamed `IsBeingRemoved` in L9 |
| `WorkerSnapshot.ParentMappedState` | L1 | deleted in L5 |
| `MappedParentStateProvider`, `setMappedParentState()`, `applyStateMapping()` | L1 | deleted in L5 |

## Deferred

| Item | Target | Reason deferred |
|------|--------|-----------------|
| Supervisor reads `NextResult.Children` (with legacy `WrappedDesiredState.ChildrenSpecs` fallback) | L5 | Reconciliation rewrite — coherent with the four parent migrations |
| `exampleparent`, `application`, `transport`, `communicator`: move `RenderChildren` from `DeriveDesiredState` into `State.Next()` | L5 | Per-parent surgery; bundles with supervisor wiring |
| `ChildSpec.Enabled` gating + `Supervisor.Reducer` (CHANGE-19) | L5 | Same reconciliation pass |
| `ParentMappedState` + `ChildStartStates` removal | L5 | Replaced by Enabled gating |
| `IsShutdownRequested` → `IsBeingRemoved` rename | L9 | Naming cleanup |
| `WrapStatus` removal from worker surface | L7 (W-3) | Bundles with `MetricsRecorder` promotion + capability interface work |
| Drop dead `extraDeps map[string]any` param from `factory.RegisterWorkerAndSupervisorFactoryByType` signature | post-restack cleanup | Param is dead from worker POV; survives in cascade-pr5 too. Signature trim is cosmetic, not load-bearing. Defer until cascade is fully landed. |

## Stack position

\`\`\`
origin/staging
   ↓
feature/wapi-L0-strip-arch-validators  (PR #2565 — L0, MERGED)
   ↓
feature/wapi-L1-clean  (PR #2566 — L1, MERGED)
   ↓
feature/wapi-L1.5-shutdown  (PR #2571 — L1.5, MERGED)
   ↓
feature/wapi-L2a  (PR #2570 — L2a)
   ↓
feature/wapi-L3  (PR #2573 — L3)
   ↓
feature/wapi-L4  (THIS PR)
\`\`\`


umh-core/pkg/fsmv2/register/deps_registry.go
```diff
@@ -0,0 +1,90 @@
...
+
+// SetDeps publishes typed deps for workerType. Parent wiring calls this
+// before the register.Worker factory closure runs for that type. Thread-safe.
+// Overwrites any prior value for the same key.
+func SetDeps[TDeps any](workerType string, deps TDeps) {
+	depsRegistryMu.Lock()
+	defer depsRegistryMu.Unlock()
+
+	depsRegistry[workerType] = deps
+}
```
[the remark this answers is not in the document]
JeremyTheocharis: Klaus flagged a real collision class — confirmed against substrate.

Root cause is API conflation: `register.SetDeps[T]` carries two semantically different storage operations under one name. The persistence callsite (`cmd/main.go:614`) is a process-startup-injected singleton and is correct. The transport callsite (`workers/transport/worker.go:236`) publishes the parent's full deps instance keyed by workerType — correct for singleton-transport today, last-writer-wins for multi-instance parents (Bridge, DFC, Stream Processor, S6 services).

Not amending L4. Lands in a dedicated typed-deps restructure layer under a static-injection model where the supervisor extracts parent deps directly and passes them through the factory:

- Parent → child wiring drops the parent→child registry path entirely. The supervisor already owns the parent worker (in `s.workers`); it extracts the parent's deps via a new `SharedDeps()` accessor at child-factory-invocation time and passes them through `factory.NewWorkerByType` to the child's typed SetDepsBuilder closure.
- Process-startup singletons (persistence) keep a workerType-keyed registry entry under the renamed `SetSharedDeps[T](workerType, deps)`. The registry shrinks to this one use.
- Children read a parent-shared-deps interface (canonical: `transport.TransportSharedDeps`), not the parent's full concrete deps. Per-worker fields (parent's own bd, sentry binding) aren't on the interface.

Per-worker deps (MetricsRecorder, bd, logger binding) never cross the registry boundary under any path; constructed fresh inside factory closures via `deps.NewBaseDependencies`.

Layer ships before any multi-instance parent migration (ENG-4941 S6 rewrite is the trigger).

umh-core/pkg/fsmv2/workers/transport/pull/worker.go
```diff
@@ -147,31 +146,40 @@ func (w *PullWorker) GetInitialState() fsmv2.State[any, any] {
...
-		func(cfg interface{}) interface{} {
-			return supervisor.NewSupervisor[fsmv2.Observation[snapshot.PullStatus], *fsmv2.WrappedDesiredState[snapshot.PullDesiredState]](
-				cfg.(supervisor.Config))
-		},
-	); err != nil {
-		panic(fmt.Sprintf("failed to register pull worker: %v", err))
-	}
+			return d
+		})
 }
```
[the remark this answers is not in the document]
JeremyTheocharis: Klaus flagged: the factory closure calls `w.InitBase(id, logger, sr)` and discards the returned `*BaseDependencies`, while the SetDepsBuilder closure constructs its own via `deps.NewBaseDependencies(logger, sr, id)`. Two bd allocations per worker init.

The "metrics written to the wrong instance" half doesn't hold at L4 tip. `WorkerBase` stores `identity`, `logger`, `initialized` — not the bd from `InitBase`'s return (`worker_base.go:56-70`). The discarded bd is GC'd; the bd inside `PullDependencies` is the one the supervisor reads via `GetDependenciesAny`.

The double allocation is still a smell. The canonical pattern documented in `pkg/fsmv2/CLAUDE.md` is the source. The `SetDepsBuilder` signature `func(id, logger, sr) T` invites the duplicate `NewBaseDependencies` call by shape.

Not amending L4. Absorbed into the typed-deps restructure layer (sketch linked from Finding 1's reply). That layer changes SetDepsBuilder's signature to `func(bd *deps.BaseDependencies, parent TParent) T` — single bd allocation by construction, parent comes from supervisor injection, no registry lookup inside the closure.

umh-core/pkg/fsmv2/workers/transport/push/worker.go
```diff
@@ -147,31 +146,40 @@ func (w *PushWorker) GetInitialState() fsmv2.State[any, any] {
...
-		func(cfg interface{}) interface{} {
-			return supervisor.NewSupervisor[fsmv2.Observation[snapshot.PushStatus], *fsmv2.WrappedDesiredState[snapshot.PushDesiredState]](
-				cfg.(supervisor.Config))
-		},
-	); err != nil {
-		panic(fmt.Sprintf("failed to register push worker: %v", err))
-	}
+			return d
+		})
 }
```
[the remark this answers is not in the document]
JeremyTheocharis: Klaus flagged: identical shape to the pull-worker finding — factory closure discards InitBase's bd, builder closure allocates a second one (`worker.go:148-185`, mirrors pull).

Same analysis: no bug today (`WorkerBase` doesn't retain InitBase's return); same smell (the SetDepsBuilder signature drives the duplicate allocation).

Not amending L4. Same fix and same home as the pull-worker thread — the typed-deps restructure layer; push and pull collapse together via the new `SetDepsBuilder[T, TParent]` signature.

umh-core/pkg/fsmv2/workers/transport/push/worker.go
```diff
@@ -147,31 +146,40 @@ func (w *PushWorker) GetInitialState() fsmv2.State[any, any] {
...
 
 func init() {
-	if err := factory.RegisterWorkerAndSupervisorFactoryByType(
-		"push",
-		func(id deps.Identity, logger deps.FSMLogger, stateReader deps.StateReader, extraDeps map[string]any) fsmv2.Worker {
-			parentDepsRaw, ok := extraDeps["transport_deps"]
+	register.Worker[snapshot.PushDesiredState, snapshot.PushStatus, *PushDependencies]("push",
+		func(id deps.Identity, logger deps.FSMLogger, sr deps.StateReader) (fsmv2.Worker, error) {
```
led0nk, line 148: `NewPushWorker` rejects a nil logger with a typed error, but the closure here skips that check. Means a nil logger now panics deep inside `deps.NewBaseDependencies` instead of bubbling up cleanly. Can we restore the guard, or just call `NewPushWorker` from the closure and forward the err?

umh-core/pkg/fsmv2/workers/transport/pull/worker.go
```diff
@@ -147,31 +146,40 @@ func (w *PullWorker) GetInitialState() fsmv2.State[any, any] {
...
 
 func init() {
-	if err := factory.RegisterWorkerAndSupervisorFactoryByType(
-		"pull",
-		func(id deps.Identity, logger deps.FSMLogger, stateReader deps.StateReader, extraDeps map[string]any) fsmv2.Worker {
-			parentDepsRaw, ok := extraDeps["transport_deps"]
+	register.Worker[snapshot.PullDesiredState, snapshot.PullStatus, *PullDependencies]("pull",
+		func(id deps.Identity, logger deps.FSMLogger, sr deps.StateReader) (fsmv2.Worker, error) {
```
led0nk, line 148: same as for push

umh-core/pkg/fsmv2/workers/transport/push/worker.go
```diff
@@ -147,31 +146,40 @@ func (w *PushWorker) GetInitialState() fsmv2.State[any, any] {
...
+			pdeps, ok := rawDeps.(*PushDependencies)
+			if !ok || pdeps == nil {
+				return nil, fmt.Errorf("push deps builder returned %T; want *PushDependencies (parent transport deps may not be published)", rawDeps)
 			}
-
-			worker, err := NewPushWorker(id, logger, stateReader, parentDeps)
+			if id.WorkerType == "" {
+				id.WorkerType = "push"
+			}
+			w := &PushWorker{}
```
led0nk: The closure rebuilds the worker by hand (`&PushWorker{}` + `InitBase` + `BindDeps`) and never goes through `NewPushWorker` anymore. So we've got two construction paths, and the public one isn't reachable from the factory — they'll drift. Either drop `NewPushWorker` if it's only used by tests now, or have the closure call it through a small adapter so there's one source of truth.

umh-core/pkg/fsmv2/workers/transport/pull/worker.go
```diff
@@ -147,31 +146,40 @@ func (w *PullWorker) GetInitialState() fsmv2.State[any, any] {
...
+			pdeps, ok := rawDeps.(*PullDependencies)
+			if !ok || pdeps == nil {
+				return nil, fmt.Errorf("pull deps builder returned %T; want *PullDependencies (parent transport deps may not be published)", rawDeps)
 			}
-
-			worker, err := NewPullWorker(id, logger, stateReader, parentDeps)
+			if id.WorkerType == "" {
+				id.WorkerType = "pull"
+			}
+			w := &PullWorker{}
```
led0nk: same as for push

umh-core/pkg/fsmv2/workers/transport/push/worker.go
JeremyTheocharis: Fixed in 16433a39d. The closure now ends `return NewPushWorker(id, logger, sr, pdeps)` instead of hand-rolling `&PushWorker{}` + InitBase + BindDeps, so production and tests share one constructor. NewPushWorker takes the already-built `*PushDependencies`; the SetDepsBuilder closure still builds them and keeps its SentryError logging. Matches cascade-pr5.

umh-core/pkg/fsmv2/workers/transport/push/worker.go
JeremyTheocharis: Same change (16433a39d): routing through NewPushWorker means its `logger == nil` guard (plus a new `dependencies == nil` guard) gates the single construction path. Caveat: on the production path SetDepsBuilder still calls `NewBaseDependencies(logger, …)` before the constructor runs (same as cascade-pr5), so a genuinely nil logger still panics there. The guard is the constructor's contract for direct/test callers, not a new production short-circuit.

umh-core/pkg/fsmv2/workers/transport/pull/worker.go
JeremyTheocharis: Same as push — fixed in 16433a39d (closure calls NewPullWorker).

umh-core/pkg/fsmv2/workers/transport/pull/worker.go
JeremyTheocharis: Same as push — fixed in 16433a39d.

JeremyTheocharis:
Re Janik's review: I audited all 12 workers' `register.Worker` construction against cascade-pr5. Communicator had the same hand-rolled construction (the only other one). All three now go through their `New*Worker` constructor in 16433a39d, and communicator's constructor gained the `logger == nil` guard the other workers already had.

led0nk (approved):

PR #2587, opened 2026-05-27
fix(metrics): correct output metrics for switch / broker / fallback outputs (ENG-5006)

JeremyTheocharis:
## Problem

Sebastian built an injection-molding monitoring data flow with a `switch` output that routes events to three different SQL writers (job-start, job-end, logbook). The data lands in Postgres correctly: he can query the rows. The Management Console shows the data flow's output metrics as zero. He asked the team whether his pipeline is broken or whether the UI is lying to him.

The UI is lying. Any user with a `switch`, `broker`, or `fallback` output sees the same thing: throughput in the database, zero in the metrics panel. That panel is how operators check whether a pipeline is healthy. When it disagrees with the database, the next message in Slack is "is the bridge broken?" and someone burns half a day proving the answer is no.

## What We're Shipping

The metrics panel now shows the correct total for outputs that write to more than one destination. Each input and output is also tracked per-destination in the metric stream, so the total reflects every writer instead of whichever one the parser happened to keep last.

## Scope

- umh-core's metrics parser only. The benthos wire format was already correct.
- NOT a Management Console UI change. The per-destination data is now in the stream, but how (or whether) the UI breaks it out per-destination is a separate ticket.
- NOT a change to input throughput or processor metrics (those were never affected).

## What changed

- `Metrics.Inputs` and `Metrics.Outputs` are now `map[string]InputInstance` / `map[string]OutputInstance` keyed by benthos `path`, replacing the flat `Input` / `Output` fields.
- The parser populates the maps. Throughput, `IsMetricsErrorFree`, and the FSM actions read the new `*Total()` helpers. The communicator emits one `DfcMetric` per path.
- Shrinking `Metrics` triggered `betteralign` reorders in 6 unrelated files.

## Where to look

It is a big diff, but most of it is mechanical. The actual change is in a few files:

**The fix:**
- `pkg/service/benthos_monitor/benthos_monitor.go` is the heart of it: the per-path `Inputs`/`Outputs` model, the 8 `*Total()` helpers, and the parser rewrite.

**Consumers moved onto the new model (real, but small):**
- `pkg/communicator/actions/providers/metrics_provider.go` and `.../get-dataflowcomponent-metrics.go` now emit one metric set per destination instead of a hardcoded single path.
- `pkg/service/benthos/benthos.go`, `pkg/service/benthos_monitor/metrics_state.go`, and the two `pkg/fsm/*/actions.go` are one-line swaps from a flat field to a `*Total()` helper.

**New tests (read to confirm behavior):**
- `switch_output_repro_test.go` (the reproduction), `get-dataflowcomponent-metrics_test.go`, `metrics_state_test.go`.

**Mechanical, safe to skim:**
- Roughly ten `*_test.go` and `mock.go` files: synthetic `Metrics{}` literals rewritten from the old flat shape to the new map shape. No logic.
- `protocolconverter.go`, `streamprocessor.go`, `topicbrowser.go`, `dataflowcomponent.go`, and two `observed_state_snapshot.go`: `betteralign` field reordering only, from the struct shrinking when the flat fields were dropped. No behavior change.


JeremyTheocharis:
@coderabbitai review

JeremyTheocharis:
@coderabbitai review

FalconTube (approved):
Looks good to me.
I just pushed one big boilerplate simplification to this branch.
Take a look at it.
If you find it ok, then can be merged from my side

PR #2588, opened 2026-05-28
feat(fsmv2/L5a): parents declare children inline in State.Next(); 3 of 4 parents migrated

JeremyTheocharis:
## Why

Before L5a, a parent worker declared its children in `DeriveDesiredState()` and chose which parent states they ran in via `ChildStartStates`. The state machine that controlled everything else didn't control children. Pausing a child without losing its retry counters, buffers, and health state wasn't possible — the only way to "stop" a child was to remove it from `ChildStartStates`, which despawned it.

L3 introduced `Transition(state, signal, action, reason, children)` but every callsite passed `nil` for the children argument. L5a is the cutover that actually uses it.

## What's in this PR

- **Parents declare children inline in `State.Next()`.** `transport`, `communicator`, and `exampleparent` now compute their child set in a package-level `RenderChildren(cfg, enabled bool)` function and return it as the fifth `Transition()` argument. The supervisor reads `NextResult.Children`; the legacy `WrappedDesiredState.ChildrenSpecs` route stays alive only for the un-migrated `application` worker.
- **Pause without despawn via `Enabled=false`.** Setting `Enabled` to false on a `ChildSpec` keeps the child resident in its Stopped state. The child's dependency state (retry trackers, buffers, health) is preserved across the pause. Replaces the old "remove from `ChildStartStates`" pattern, which despawned the child and dropped its state.
- **3 of 4 parents migrated.** transport, communicator, exampleparent. The 4th parent is `application`, which accepts arbitrary user-declared children as opaque YAML and needs structural typing before it can use `NewChildSpec[T]`. Deferred to L5c.
- **One typed constructor for all child specs.** `NewChildSpec[T]` replaces the per-parent `makePushChildSpec` / `makePullChildSpec` / `makeTransportChildSpec` factories.
- **Discoverability.** A new worker author opening `pkg/fsmv2` shouldn't have to read `CLAUDE.md` to figure out how children work. Until L5a they did: `pkg/fsmv2/README.md § Child worker management` showed a code example referencing a `Children: []ChildSpec` field that no current worker uses (copy-pasting it didn't compile), `pkg/fsmv2/doc.go § Parent-child workers` described the pre-`Enabled` `ChildStartStates` gating model, and `examplechild/state/state_stopped.go` had the three-way Stopped branch in code with no godoc explaining it. Rewrote both prose sections with examples that compile, and added a godoc on `examplechild/state/state_stopped.go` (shutdown → remove; disabled → stay; otherwise → resume).
- **CI gate.** `MISSING_CHILDSPEC_VALIDATION` rejects migrated parent worker packages that lack a `RenderChildren` function.

## Deferred

| Item | Target | Note |
|------|--------|------|
| `application` worker migrates onto `RenderChildren` + `NewChildSpec[T]` | L5c | Needs structural typing first; affects `cmd/main.go`, config-parse, many scenarios. |
| `WorkerSnapshot.ParentMappedState` field removal + child-side `ShouldStop()` simplification | L5b | `application` still reads `ParentMappedState` via the legacy `computeMappedState`/`applyStateMapping` path until L5c. If L5c slips, L5b ships only the migrated workers' simplification and defers the field removal. |
| `WrappedDesiredState.ChildrenSpecs` route removal | L5c | `application` is its last reader. |
| `ChildSpec.ChildStartStates` field removal | L5c | `application` is its last reader. |
| `ChildSpec.UserSpec.Config` raw-YAML field removal | L5c | Atomic with `application`'s typed migration. |
| `computeMappedState` / `applyStateMapping` removal | L5c (`application`-side) + L5b (supervisor wiring) | Legacy gating machinery. |
| `WorkerSnapshot.IsShutdownRequested` renamed to `IsBeingRemoved` (~79 files) | L9 | Naming cleanup. |
| Stopped-state three-way branch for non-migrated workers (helloworld, examplefailing, examplepanic, exampleslow, persistence) | L9 | Same branch shape extended to the rest. |
| Reason-string observability sweep (deferred CodeRabbit findings from PR #2573) | L5b+0 (ENG-4799 sub-issue) | Post-cascade canonical shape. |
| README `snapshot/` template + DEPENDENCIES.md API conflict + folder/type naming rule | L5c | Doc staleness folded into L5c's cleanup. |

## Stack position

```
origin/staging
   ↓
… L0 / L1 / L1.5 / L2a / L3 (all MERGED)
   ↓
feature/wapi-L4   (#2584 — awaiting re-approval after update-branch)
   ↓
feature/wapi-L5a  (THIS PR — draft)
```


umh-core/pkg/fsmv2/config/childspec.go
```diff
@@ -218,7 +231,18 @@ type ChildSpec struct {
...
+	// empty. Remove together with computeMappedState when the application
+	// worker moves to RenderChildren.
+	ChildStartStates []string `json:"childStartStates,omitempty" yaml:"childStartStates,omitempty"` // Parent FSM states where child should run (empty = always run)
+
+	// Enabled is the parent's per-tick request for this child to run. Zero
+	// value (false) means "stopped but resident" — children stay in Stopped
+	// without being despawned. Setting Enabled=false then Enabled=true resumes
+	// the child on the next tick. Children read snap.IsDisabled, not Enabled
+	// directly; the disable-mapping pass translates Enabled to IsDisabled.
+	Enabled bool `json:"enabled" yaml:"enabled"`
```
led0nk, line 250: Enabled zero-value (false) means disabled, so any child constructed without explicitly setting it is held in Stopped every ~100ms tick by applyDisableMapping — no validation error, DEBUG-only log. The workaround in application/worker.go:138 (forcing Enabled=true with a comment that disable-mapping would otherwise kill all children) shows the trap already bites. No production parent hits it today, but the next parent built on the legacy GetChildrenSpecs path will. Consider inverting to `Disabled bool` (zero = enabled), or normalizing Enabled=true in one place on the legacy path.

umh-core/pkg/fsmv2/supervisor/lifecycle.go
```diff
@@ -682,6 +681,77 @@ func (s *Supervisor[TObserved, TDesired]) handleWorkerRestart(ctx context.Contex
 	return nil
 }
 
+// setDisabled mirrors requestShutdown but uses the Disableable interface.
+// The disable-mapping pass is the exclusive caller; no other subsystem may call this.
+func (s *Supervisor[TObserved, TDesired]) setDisabled(ctx context.Context, workerID string, disabled bool) error {
```
led0nk, line 686: setDisabled does LoadDesiredTyped + Marshal + Unmarshal on every call with no compare-to-current, and applyDisableMapping calls it per child every reconcile tick while s.mu is held (reconciliation.go:1443). That serializes 2N+ store round-trips inside the parent write critical section, blocking AddWorker/RemoveWorker — the opposite of children_view.go:23, which deliberately drops s.mu before per-child access. Make disable diff-driven: short-circuit when the loaded Disabled bit already equals target, snapshot intents under the lock, persist outside it. Preserve same-tick visibility of the bit to the child snapshot — the transport enable-one-tick-early trick in state_stopped.go:58 depends on it.

umh-core/pkg/fsmv2/supervisor/reconciliation.go
```diff
@@ -1577,6 +1618,13 @@ func (s *Supervisor[TObserved, TDesired]) reconcileChildren(specs []config.Child
 		}
 	}
 
+	// applyDisableMapping propagates ChildSpec.Enabled into each child's IsDisabled bit.
+	// Runs inside reconcileChildren, after the per-spec exists-check and before despawn.
+	// Nil specs are a no-op: leaf workers legitimately return nil from GetChildrenSpecs.
+	if specs != nil {
+		s.applyDisableMapping(context.Background(), specs)
```
led0nk, line 1625: The Disabled bit applyDisableMapping sets here is overwritten later in the same tick: the derive-save preserve block (~L828) re-applies only ShutdownRequested, then SaveDesired writes Disabled=false. A child resuming purely on IsDisabled sees it flipped back to enabled. Only examplechild exercises this today, and the existing test passes incidentally via the one-tick observed-state lag (api.go:546). Preserve Disabled alongside ShutdownRequested, and add a test that ticks the child through a full parent tick asserting the stored bit stays true and the child stays Stopped.

umh-core/pkg/fsmv2/workers/transport/state/children_helpers.go
```diff
@@ -0,0 +1,42 @@
...
+
+package state
+
+import (
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsmv2/config"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsmv2/workers/transport/snapshot"
+)
+
+// yaml.Marshal of the static BaseUserSpec cannot fail in practice; on the
+// hypothetical error we still return nil so the supervisor falls back to
```
led0nk: This comment is wrong for migrated workers. Returning nil does not fall back to a safe set: nil rendered children make tick() call GetChildrenSpecs(), which returns nil here (ChildrenSpecs unset), so reconcileChildren(nil) marks every resident child pendingRemoval — exactly the despawn the comment says it prevents, with no operator signal. Likelihood is low (the only error source is static-spec marshal, which fails deterministically in tests). Fix the comment and log before returning nil, or return the last-known child set.

umh-core/pkg/fsmv2/workers/transport/state/state_starting.go
```diff
@@ -64,7 +64,7 @@ func (s *StartingState) Next(snapAny any) fsmv2.NextResult[any, any] {
 			if time.Since(snap.Status.LastAuthAttemptAt) < delay {
 				return fsmv2.Transition(s, fsmv2.SignalNone, nil,
 					fmt.Sprintf("auth backoff: %d errors (%s), delay %s",
-						snap.Status.ConsecutiveErrors, snap.Status.LastErrorType, delay.Round(time.Second)), nil)
+						snap.Status.ConsecutiveErrors, snap.Status.LastErrorType, delay.Round(time.Second)), childrenAlive(snap.Config))
```
led0nk, line 67: Untested lifecycle widening. Pre-cutover push/pull carried ChildStartStates=[Running,Degraded] (deleted in 6245fd28d); now empty maps to Running unconditionally, so push/pull reach Running during parent auth and during a permanent auth failure. Action stays gated on HasTransport && HasValidToken so they sit idle, and this looks intentional (relates to ENG-4672). The gap is it is untested and the intent is undocumented — the comment at L81 still says Children start via ChildStartStates. Add a parity test across Starting/AuthFailed/Stopping and document the intent.

umh-core/pkg/fsmv2/config/childspec.go
JeremyTheocharis: Disabled-by-default is deliberate, locked during the cascade design (2026-04-28): the zero value was accepted because ChildSpec is rebuilt every tick from explicit parent construction — `NewChildSpec` takes `enabled` as a required parameter, and every in-tree `RenderChildren` sets the field explicitly. The YAML path in `application/worker.go` escaped that premise, hence the normalization loop you found; the migration API (ENG-5088) centralizes exactly that normalization for every config-born child.

On inverting to `Disabled`: forgetting the field would then fail toward running, which is the worse direction. I also blind-probed the API shape (3 agents writing the child declaration cold, no docs): two guessed a field literally named `Enabled`, none reached for `Disabled` — but two of the three expected bare declarations to run, so the silent default was a real doc gap. abac437dc makes it loud at the three places a worker author reads: the `Enabled` godoc, the fsmv2 README, and CLAUDE.md.

umh-core/pkg/fsmv2/supervisor/lifecycle.go
JeremyTheocharis: Right on all points: no compare-to-current, and the round-trips run inside `s.mu`. ENG-5095 touches the same write path (design B replaces the whole-doc `SaveDesired` with field-scoped `SaveDesiredFields`) but adds neither the compare-to-current short-circuit nor any lock-scope relief, and it is parked behind the cascade — so the contention fix needs its own ticket. The migration API (ENG-5088) will also increase churn through this path via runtime child upserts. Filed ENG-5134 as a new ticket under the fsmv2 tech-debt parent ENG-4342, cross-linked to ENG-5095 so the two don't collide on the same write primitive: make disable diff-driven (short-circuit when the loaded bit already equals the target, snapshot intents under the lock, persist outside it), with the `state_stopped.go:58` one-tick-early visibility called out as the rabbit hole. Not taking it in this PR: contention, not correctness.

umh-core/pkg/fsmv2/supervisor/reconciliation.go
JeremyTheocharis: Correct on this commit — and already fixed directly above in the stack: #2589's "fix(fsmv2/supervisor): preserve Disabled across DeriveDesiredState re-derivation" (3837a5534, authored 2026-05-29 and in #2589 since it opened on 2026-05-31; the SHA will change when the stack rebases onto this tip) extends the preserve block to carry `Disabled` alongside `ShutdownRequested`. It lives there because L5b's deletion of `applyStateMapping` is what turns the timing-accidental clobber deterministic, so the fix travels with the change that exposes it; patching it here would collide with that hunk. The full-parent-tick test you ask for is a real gap (the existing `disabled_child_test.go` passed incidentally via the one-tick lag): filed ENG-5135.

umh-core/pkg/fsmv2/workers/transport/state/children_helpers.go
JeremyTheocharis: Pushed 4bcd65e03: the comment now states the real path (nil rendered children → nil `GetChildrenSpecs()` for migrated workers → `reconcileChildren(nil)` despawn) and both helpers log at error level before returning nil. 8c4c37b9d extends the same guard+log to exampleparent's copy of the helper (its render builds specs literally and cannot fail today, so the comment there marks the guard as future-proofing). Scope note: the direct `RenderChildren` call sites (transport's Stopped/Stopping states, the inline copies in communicator and exampleparent's StoppedState) still return nil silently — ENG-5115's keep-last-rendered removes that failure mode outright instead of logging it at every site. The semantic fix is tracked in ENG-5115 (filed the same day during the adoption-wave triage); it is blocked by the migration API, its unblock point, since deleting the nil fallback needs every worker rendering children from `Next()`. For transport, today's only render-error source is the static BaseUserSpec marshal you noted.

umh-core/pkg/fsmv2/workers/transport/state/state_starting.go
JeremyTheocharis: Intended — the widening in this PR is deliberate: children stay enabled (idle-but-resident) through the whole auth trajectory including AuthFailed. Actions gate separately on `HasTransport && HasValidToken`, and holding children resident avoids despawn/spawn churn through auth windows. Pushed 8c629fafe: `children_gating_test.go` asserts the emitted child set for every parent-state branch (Stopped/Starting/Running/Degraded/AuthFailed/Stopping, 22 specs). One scope note: push/pull have no Starting/AuthFailed states of their own, so the parity test lives at the transport parent level. The stale `ChildStartStates` comments are already rewritten in #2589, and 31b1b9138 fixes the one survivor: the AuthFailedState godoc still claimed children remain stopped until RunningHealthy.

umh-core/pkg/fsmv2/config/childspec.go
JeremyTheocharis: so in short: per design, but made it more clear in commit abac437dc

umh-core/pkg/fsmv2/workers/transport/state/state_starting.go
JeremyTheocharis: double checking this

JeremyTheocharis: i would leave it like this

JeremyTheocharis: reasoning: we could have let the push/pull be removed during authentication, but that cause unnecessary cycles when something during auth hiccups, so here we just let them remain running (and they wait untikl the y have a good token anyway)

led0nk (approved):
lgtm, approved needs formatting but free to go

umh-core/pkg/fsmv2/config/childspec.go
```diff
@@ -379,118 +408,81 @@ type ChildInfo struct {
...
+		}
+
+		if !c.IsStopped {
+			view.AllStopped = false
 		}
 	}
 
-	return true
+	return view
 }
```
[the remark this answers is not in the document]
JeremyTheocharis: Fixed in e97331af8 — NewChildrenView copies the input slice before precomputing the aggregates.

umh-core/pkg/fsmv2/internal/validator/worker.go
```diff
@@ -478,13 +479,18 @@ func checkChildSpecValidation(filename string) []Violation {
...
-			violations = append(violations, Violation{
-				File:    filename,
-				Line:    pos.Line,
-				Type:    "MISSING_CHILDSPEC_VALIDATION",
-				Message: "DeriveDesiredState() should validate ChildrenSpecs before returning",
-			})
+			// A parent that declares a top-level RenderChildren passes the
+			// no-children-in-DeriveDesiredState check, even if DeriveDesiredState
+			// has no range-over-Children loop.
+			if !dirHasRenderChildren(filepath.Dir(filename)) {
```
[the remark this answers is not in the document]
JeremyTheocharis: Declining — the call site scans the worker's own directory, and FSMv2 doesn't nest worker packages under a parent, so there's no false-pass surface. The recursive walk matches the documented "any file under the directory" convention.

umh-core/pkg/fsmv2/workers/communicator/state/state_recovering.go
```diff
@@ -32,19 +32,29 @@ type RecoveringState struct {
...
 
+	aliveChildren, err := communicator.RenderChildren(snap.Config, true)
+	if err != nil {
+		aliveChildren = nil
+	}
+
+	stopChildren, err := communicator.RenderChildren(snap.Config, false)
+	if err != nil {
+		stopChildren = nil
+	}
```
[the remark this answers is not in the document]
JeremyTheocharis: Deferred — the nil-on-render-error semantic is owned by ENG-5115 (nil = keep last rendered children; blocked on ENG-5088). Communicator's sites stay as-is until then.

umh-core/pkg/fsmv2/workers/communicator/state/state_recovering.go
```diff
@@ -32,19 +32,29 @@ type RecoveringState struct {
...
+
 	if snap.IsShutdownRequested {
-		return fsmv2.Transition(&StoppedState{}, fsmv2.SignalNone, nil, "Shutdown requested during recovering state", nil)
+		return fsmv2.Transition(&StoppedState{}, fsmv2.SignalNone, nil, "Shutdown requested during recovering state", stopChildren)
 	}
```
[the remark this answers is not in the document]
JeremyTheocharis: Declining — the shutdown branch already precedes the business-logic transitions; the render calls only feed stopChildren into it. This file is a parked ENG-5115 site, no restructuring on this PR.

umh-core/pkg/fsmv2/workers/example/exampleparent/children.go
```diff
@@ -0,0 +1,60 @@
...
+		specs = append(specs, config.ChildSpec{
+			Name:       fmt.Sprintf("child-%d", i),
+			WorkerType: childWorkerType,
+			UserSpec: config.UserSpec{
+				Config:    cfg.ChildConfig,
+				Variables: childVariables,
+			},
+			ChildStartStates: []string{"TryingToStart", "Running"},
+			Enabled:          enabled,
+		})
```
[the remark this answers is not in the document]
JeremyTheocharis: Declining — NewChildSpec only sets UserSpec.Config; it can't carry the per-child DEVICE_ID in UserSpec.Variables, and ChildConfig is already a raw string (yaml.Marshal would double-encode it). The literal is correct here.

umh-core/pkg/fsmv2/workers/transport/pull/state/state_stopped.go
```diff
@@ -36,7 +36,11 @@ func (s *StoppedState) Next(snapAny any) fsmv2.NextResult[any, any] {
 		return fsmv2.Transition(s, fsmv2.SignalNeedsRemoval, nil, "shutdown requested, signaling removal", nil)
 	}
 
-	if snap.ParentMappedState == config.DesiredStateRunning && !snap.IsShutdownRequested {
+	if snap.IsDisabled {
+		return fsmv2.Transition(s, fsmv2.SignalNone, nil, "disabled by supervisor, staying stopped", nil)
```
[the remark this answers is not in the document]
JeremyTheocharis: Declining for this stack — #2589 deletes ParentMappedState from this file (Stopped→Running gates on ShouldStop()), so the suggested field is gone one PR up. The disabled branch's only input is the IsDisabled bit the message already states.

umh-core/pkg/fsmv2/workers/transport/push/state/state_stopped.go
```diff
@@ -35,6 +35,10 @@ func (s *StoppedState) Next(snapAny any) fsmv2.NextResult[any, any] {
 		return fsmv2.Transition(s, fsmv2.SignalNeedsRemoval, nil, "shutdown requested, signaling removal", nil)
 	}
 
+	if snap.IsDisabled {
+		return fsmv2.Transition(s, fsmv2.SignalNone, nil, "disabled by supervisor, staying stopped", nil)
```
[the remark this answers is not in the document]
JeremyTheocharis: Declining for this stack — #2589 deletes ParentMappedState from this file (Stopped→Running gates on ShouldStop()), so the suggested field is gone one PR up. The disabled branch's only input is the IsDisabled bit the message already states.

umh-core/pkg/fsmv2/workers/transport/state/state_stopped.go
```diff
@@ -36,19 +36,42 @@ type StoppedState struct {
 func (s *StoppedState) Next(snapAny any) fsmv2.NextResult[any, any] {
 	snap := fsmv2.ConvertWorkerSnapshot[snapshot.TransportDesiredState, snapshot.TransportStatus](snapAny)
 
+	// Stopped emits children with enabled=false (resident, not despawned).
+	// On IsShutdownRequested, SignalNeedsRemoval drives removal so the children arg is ignored.
```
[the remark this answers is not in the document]
JeremyTheocharis: Declining — the emit-aliveChildren-now block documents the one-tick-early enable trick and is referenced from ENG-5134; it stays as deliberate non-obvious-mechanism documentation.

umh-core/pkg/fsmv2/workers/transport/state/children_helpers.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	return specs
+}
+
+func childrenStopped(cfg snapshot.TransportDesiredState) []config.ChildSpec {
+	specs, err := snapshot.RenderChildren(cfg, false)
+	if err != nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
JeremyTheocharis: Partly addressed in 4bcd65e03 (error-level log + corrected comment); the signature-change/propagation is the ENG-5115 semantic fix, deferred.

umh-core/pkg/fsmv2/workers/example/exampleparent/state/children_helpers.go
```diff
@@ -0,0 +1,38 @@
...
+// exampleparent's RenderChildren builds its specs literally and cannot fail
+// today; the guard and log are future-proofing for when it can. The semantic
+// fix (nil = keep last rendered children) is tracked in ENG-5115.
+func childrenAlive(cfg exampleparent.ExampleparentConfig) []config.ChildSpec {
+	specs, err := exampleparent.RenderChildren(cfg, true)
+	if err != nil {
+		logger.For("exampleparent").Errorf("RenderChildren(enabled=true) failed, all resident children will be despawned (ENG-5115): %v", err)
+
+		return nil
+	}
```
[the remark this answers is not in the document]
JeremyTheocharis: Same ENG-5115 scope — this guard is unreachable today (exampleparent's RenderChildren builds specs literally and can't fail) and logs at error level when it ever fires; propagation lands with ENG-5115.

PR #2589, opened 2026-05-31
L5b: remove the legacy child-control mechanism

JeremyTheocharis:
## The stack

> **FSMv2 migration-API stack, PR 1 of 6.** A worker acts on its *snapshot* (its
> 1. **#2589 ← you are here**: delete the dead legacy stop path so a worker's lifecycle depends on its own state, not a parent side channel.
>

### This PR: a worker's stop decision reads its own state, not the parent's

The framework still had an older mechanism (`ParentMappedState` / `ChildStartStates` /
`computeMappedState`) that let a worker's *stop* decision depend on its parent's FSM
state name. It was already dead (always false in practice) but still present in the code,
so the next contributor could route new behavior through it, and it kept a worker's
lifecycle tied to a parent side channel instead of the worker's own state. This PR
deletes it. A worker now stops for exactly two reasons: shutdown or disabled.

---

# L5b: remove the legacy child-control mechanism

## Problem

Child workers decided "should I stop?" through two channels: the explicit intent fields (`IsShutdownRequested`, `IsDisabled`) and a legacy parent-FSM-state-name pathway (`ChildStartStates` -> `computeMappedState` -> `ParentMappedState`, read as a third term in `ShouldStop()`). No worker emits `ChildStartStates` anymore and `application` forces `Enabled=true`, so that third term was always false. Dead but present, it invited the next contributor to route a new stop signal through the wrong channel.

## What we're shipping

1. **`fix(fsmv2/supervisor)`: preserve `Disabled` across per-tick re-derivation.** The child supervisor carried only `ShutdownRequested` forward across `DeriveDesiredState`, so a resident-disabled child's `Disabled` bit was clobbered to false each tick and survived only by tick-timing accident. Commit 2 changes that timing, so this lands first and makes resident-disable deterministic. Mirrors the existing `ShutdownRequested` preservation.

2. **`refactor(fsmv2/L5b)`: delete the legacy mechanism end to end** — `ParentMappedState`, `ChildStartStates`, `computeMappedState`/`applyStateMapping`, the supervisor + collector wiring, and the local transport copies. `ShouldStop()` becomes `return s.IsShutdownRequested || s.IsDisabled`. Folds in the dead `WrappedDesiredState.State`/`GetState()`.

After the deletion the legacy channel is **unrepresentable at the state-file API**: `WorkerSnapshot` carries no parent-state field, so a state file cannot route a stop on parent state. We rely on that structural absence plus review — an earlier draft added a name-ban architecture test, but it only caught re-introduction of the six specific dead identifiers (a failure mode the deletion already made implausible) while doing nothing against a new-named parallel channel, so it was dropped as a brittle proxy.

## Review disposition (2026-06-13)

A third commit addresses review feedback, both halves rooted in commit 2 widening `ShouldStop()` to `shutdown || disabled`:

3. **`refactor(fsmv2)`: idiomatic state returns + truthful stop reasons.**
   - **Truthful stop reasons.** Because `ShouldStop()` now covers shutdown *and* disable, the old `shutdown=%t` reason strings misreported a disable-driven stop as `shutdown=false`. Added `WorkerSnapshot.StopReason()` (`"shutdown=%t disabled=%t"`) and used it at the stop / stop-complete transition sites. (Closes the CodeRabbit "include the disable cause" cluster.)
   - **Removed the non-idiomatic "Exhaustive Transition Coverage" architecture rule** (`ValidateExhaustiveTransitionCoverage` + its arch-test block + registry entry). Go already guarantees every path returns, so the rule's only effect was to mandate a trailing `return s, SignalNone, nil` catch-all — which, once commit 2 made the parent-state branch redundant, became provably-unreachable dead code. That dead catch-all is exactly what CodeRabbit flagged as an "unreachable branch": the finding was real, but the fix is to remove the rule that forced the dead code, not to keep it. State `Next()` methods are rewritten with guard clauses + a definite final transition; the example stopped states gain an explicit `IsDisabled` branch. (Same spirit as the name-ban test dropped above: a syntactic arch rule that fought idiom rather than guarding a real invariant.)

Changelog: `skip-changelog-guard` — FSMv2 transport is default-off (`USE_FSMV2_TRANSPORT=false`), not user-facing, matching #2588.

## Scope

- umh-core `pkg/fsmv2` only. Behavior preserved: the deleted term was inert, and `childStartStates` in application YAML was never emitted by any config writer or the Go-built template, so the lenient parser ignoring a stale key changes nothing.
- Wire-shape change: `mapped_parent_state` drops from `/debug/fsmv2` (write-only field, zero readers anywhere).
- NOT in scope: the `application` root typing / `UserSpec.Config` removal, transport auth, the `IsShutdownRequested` rename. Each is its own later PR.
- CI note: the supervisor package suite runs ~280s; allow at least 360s so it does not time out.
- Was stacked on #2588 (now **MERGED**); rebased onto `staging`. **L5d (#2590) stacks on this PR.**

### DIFF
--- a/umh-core/pkg/fsmv2/config/childspec.go
+++ b/umh-core/pkg/fsmv2/config/childspec.go
@@ -46,11 +46,11 @@ import (
 //   - Disabled: Inherited from this type. Set by the supervisor's disable-mapping pass
 //     from the parent's ChildSpec.Enabled; child workers stay resident in Stopped.
 //
-// The desired lifecycle state ("running"/"stopped") is read from the user spec
-// via BaseUserSpec.GetState() in DeriveDesiredState and stored directly on the
-// outer DesiredState type (not on BaseDesiredState). Callers that used to set
-// BaseDesiredState{State: x} should instead set the State field on the embedding
-// struct (e.g., MyDesiredState{State: x, ...}).
+// The desired lifecycle state ("running"/"stopped") lives on the user spec, not
+// on the DesiredState. Embed BaseUserSpec in your UserSpec config type to inherit
+// GetState(); DeriveDesiredState reads and validates it. As stated above, do not
+// add a State field to your DesiredState — the supervisor drives lifecycle through
+// ShutdownRequested and Disabled and never populates a custom State field.
 //
 // Correct ShouldBeRunning() implementation:
 //
--- a/umh-core/pkg/fsmv2/config/childspec.go
+++ b/umh-core/pkg/fsmv2/config/childspec.go
@@ -43,7 +43,8 @@ import (
 //
 // Correct lifecycle control:
 //   - ShutdownRequested: Inherited from this type. Set by supervisor for graceful shutdown.
-//   - ParentMappedState: For child workers only. Injected by supervisor from parent's ChildStartStates.
+//   - Disabled: Inherited from this type. Set by the supervisor's disable-mapping pass
+//     from the parent's ChildSpec.Enabled; child workers stay resident in Stopped.
 //
 // The desired lifecycle state ("running"/"stopped") is read from the user spec
 // via BaseUserSpec.GetState() in DeriveDesiredState and stored directly on the
@@ -51,17 +52,15 @@ import (
 // BaseDesiredState{State: x} should instead set the State field on the embedding
 // struct (e.g., MyDesiredState{State: x, ...}).
 //
-// Correct ShouldBeRunning() implementations:
+// Correct ShouldBeRunning() implementation:
 //
-//	// Root/leaf workers (no parent):
 //	func (s *MyDesiredState) ShouldBeRunning() bool {
 //	    return !s.ShutdownRequested
 //	}
 //
-//	// Child workers (have parent):
-//	func (s *MyDesiredState) ShouldBeRunning() bool {
-//	    return !s.ShutdownRequested && s.ParentMappedState == config.DesiredStateRunning
-//	}
+// Child workers do not gate on a parent-supplied field: the supervisor's
+// disable-mapping pass sets Disabled from the parent's ChildSpec.Enabled, and
+// state files route lifecycle through snap.ShouldStop().
 //
 // See ValidateNoCustomLifecycleFields in pkg/fsmv2/internal/validator/snapshot.go for enforcement.
 type BaseDesiredState struct {
@@ -166,31 +165,13 @@ func (u UserSpec) Clone() UserSpec {
 //   - Supervisor handles "how to make it exist"
 //   - Children run independently in their own FSMs
 //
-// # Child Start States
+// # Child Lifecycle Gating
 //
-// ChildStartStates coordinates parent state with child lifecycle.
-//
-// ChildStartStates specifies which parent FSM states cause children to run.
-// When the parent is in a listed state, children run. Otherwise, they stop.
-//
-// INVARIANT: If a parent state checks ChildrenHealthy/ChildrenUnhealthy to gate
-// its own transitions, that state MUST be listed in ChildStartStates. Otherwise,
-// children are mapped to "stopped" and can never satisfy the health check,
-// creating a permanent deadlock.
-//
-// Example use case: Children should only run when parent is in "Running" or "TryingToStart" states.
-//
-// Format:
-//
-//	ChildStartStates: []string{"Running", "TryingToStart"}
-//
-// When to use:
-// - Parent lifecycle controls child lifecycle (children run only in certain parent states)
-// - Simple "run when parent is active" patterns
-//
-// When not to use:
-// - Passing data between states (use VariableBundle instead)
-// - Triggering actions (use signals instead)
+// A child runs while the parent sets ChildSpec.Enabled=true for it. The
+// supervisor's disable-mapping pass translates Enabled into the child's
+// Disabled bit, and state files route lifecycle through snap.ShouldStop().
+// Setting Enabled=false leaves the child resident in Stopped (not despawned);
+// flipping it back to true resumes the child on the next tick.
 //
 // # Dependency Inheritance
 //
@@ -206,7 +187,7 @@ func (u UserSpec) Clone() UserSpec {
 //	    Name:       "mqtt-connection",
 //	    WorkerType: "mqtt_client",
 //	    UserSpec:   UserSpec{Config: "url: tcp://localhost:1883"},
-//	    ChildStartStates: []string{"Running", "TryingToStart"}, // Child runs when parent is active
+//	    Enabled:    true, // Parent wants this child running
 //	}
 //
 // Example - Benthos managing connections and data flows:
@@ -217,25 +198,20 @@ func (u UserSpec) Clone() UserSpec {
 //	        Name:       "modbus-connection",
 //	        WorkerType: "modbus_client",
 //	        UserSpec:   UserSpec{Config: "address: 192.168.1.100:502"},
-//	        // Empty ChildStartStates = child always runs (follows parent's DesiredState.State)
+//	        Enabled:    true,
 //	    },
 //	    {
 //	        Name:       "source-flow",
 //	        WorkerType: "benthos_flow",
 //	        UserSpec:   UserSpec{Config: "input: {...}"},
-//	        ChildStartStates: []string{"Running"}, // Child only runs when parent is Running
+//	        Enabled:    true,
 //	    },
 //	}
 type ChildSpec struct {
-	Dependencies map[string]any `json:"dependencies,omitempty"     yaml:"dependencies,omitempty"` // Additional deps to merge with parent's deps (child overrides parent)
-	UserSpec     UserSpec       `json:"userSpec"                   yaml:"userSpec"`               // Raw user config (input to DeriveDesiredState)
-	Name         string         `json:"name"                       yaml:"name"`                   // Unique name for this child (within parent scope)
-	WorkerType   string         `json:"workerType"                 yaml:"workerType"`             // Type of worker to create (registered worker factory key)
-	// Deprecated. Only the application worker still consumes ChildStartStates
-	// (via computeMappedState in reconciliation.go). Other workers leave it
-	// empty. Remove together with computeMappedState when the application
-	// worker moves to RenderChildren.
-	ChildStartStates []string `json:"childStartStates,omitempty" yaml:"childStartStates,omitempty"` // Parent FSM states where child should run (empty = always run)
+	Dependencies map[string]any `json:"dependencies,omitempty" yaml:"dependencies,omitempty"` // Additional deps to merge with parent's deps (child overrides parent)
+	UserSpec     UserSpec       `json:"userSpec"               yaml:"userSpec"`               // Raw user config (input to DeriveDesiredState)
+	Name         string         `json:"name"                   yaml:"name"`                   // Unique name for this child (within parent scope)
+	WorkerType   string         `json:"workerType"             yaml:"workerType"`             // Type of worker to create (registered worker factory key)
 
 	// Enabled is the parent's per-tick request for this child to run. Zero
 	// value (false) means "stopped but resident" — children stay in Stopped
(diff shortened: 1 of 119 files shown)


umh-core/pkg/fsmv2/supervisor/reconciliation.go
```diff
@@ -825,14 +825,21 @@ func (s *Supervisor[TObserved, TDesired]) tick(ctx context.Context) (err error)
 		}
 	}
 
-	// Preserve ShutdownRequested: shutdown is a supervisor operation that overrides DeriveDesiredState
+	// Preserve ShutdownRequested and Disabled: both are supervisor operations
```
JeremyTheocharis, line 836: Evaluated: yes. This preserve copy-back goes away if we take the control fields (`ShutdownRequested`/`Disabled`) off the worker's desired *type* and write them with a field-scoped save, so the config re-derive can't clobber them. Out of scope for this PR — it touches the same supervisor/store code as the cascade. Tracked as tech debt in ENG-5095.

JeremyTheocharis:
**Review disposition (force-pushed `741b104b0`, rebased onto `staging` after #2588 merged).**

The CodeRabbit cluster on this PR splits three ways:

1. **"Include the disable cause in the stop reason" (×~10).** Addressed: `ShouldStop()` widened to `shutdown || disabled`, so the bare `shutdown=%t` strings lied on a disable-driven stop. Added `WorkerSnapshot.StopReason()` → `"shutdown=%t disabled=%t"` and used it at every stop / stop-complete site.

2. **"Unreachable branch / effectively unconditional" (examplepanic disconnected, pull `state_stopped`, …).** The finding was real — but the dead code existed only because the `Exhaustive Transition Coverage` architecture rule *mandated* a trailing `return s, SignalNone, nil` catch-all. Go already guarantees every path returns, so once commit 2 made the parent-state branch redundant that catch-all became provably unreachable. Fix: removed the non-idiomatic rule and rewrote the states with guard clauses + a definite final transition (same rationale as the name-ban arch test already dropped in this PR). Net −142 lines.

3. **"Add a changelog entry" (Major).** Declining: `skip-changelog-guard` is applied — FSMv2 transport is default-off (`USE_FSMV2_TRANSPORT=false`), not user-facing, matching #2588.

The standing `reconciliation.go` preserve-copy-back thread stays deferred to ENG-5095 (control-state overlay), as noted there.

See the updated PR description's "Review disposition" section for detail.

umh-core/pkg/fsmv2/supervisor/doc.go
```diff
@@ -422,7 +421,7 @@
 //
 //   - Check IsShutdownRequested() as first condition in state.Next()
 //   - Make all actions idempotent (check if work already done)
-//   - Use ChildStartStates to coordinate child lifecycle (not data passing)
+//   - Use ChildSpec.Enabled to coordinate child lifecycle (not data passing)
 //   - Pass data to children via VariableBundle, not direct method calls
```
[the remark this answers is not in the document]
JeremyTheocharis: Addressed in `0e89549b0` — the `StopReason()` sweep + idiomatic-return rewrite already covers this. Resolving.

umh-core/pkg/fsmv2/workers/example/examplepanic/state/state_disconnected.go
```diff
@@ -32,12 +31,12 @@ func (s *DisconnectedState) Next(snapAny any) fsmv2.NextResult[any, any] {
...
 	}
 
-	if !snap.IsShutdownRequested && snap.ParentMappedState == config.DesiredStateRunning {
+	if !snap.ShouldStop() {
 		return fsmv2.Transition(&TryingToConnectState{}, fsmv2.SignalNone, nil,
-			fmt.Sprintf("parentState=%q: should be running, transitioning to TryingToConnect", snap.ParentMappedState), nil)
+			"should be running, transitioning to TryingToConnect", nil)
 	}
 
 	return fsmv2.Transition(s, fsmv2.SignalNone, nil, "Connection lost, will retry", nil)
```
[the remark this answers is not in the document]
JeremyTheocharis: Addressed in `0e89549b0` — the `StopReason()` sweep + idiomatic-return rewrite already covers this. Resolving.

umh-core/pkg/fsmv2/supervisor/reconciliation.go
```diff
@@ -825,14 +825,21 @@ func (s *Supervisor[TObserved, TDesired]) tick(ctx context.Context) (err error)
...
-		// Check if existing state had shutdown requested via interface
+		// Check the existing state's lifecycle flags via interface.
 		if ds, ok := any(existingDesiredTyped).(fsmv2.DesiredState); ok {
 			if ds.IsShutdownRequested() {
 				desiredDoc[FieldShutdownRequested] = true
 			}
+
+			if ds.IsDisabled() {
+				desiredDoc[FieldDisabled] = true
+			}
```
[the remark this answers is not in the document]
JeremyTheocharis: Intentional `skip-changelog-guard`: FSMv2 transport is a refactor here, not a user-facing change in this PR, matching #2588. Not adding a CHANGELOG entry.

umh-core/pkg/fsmv2/config/childspec.go
```diff
@@ -43,25 +43,24 @@ import (
...
 //	    return !s.ShutdownRequested
 //	}
 //
-//	// Child workers (have parent):
-//	func (s *MyDesiredState) ShouldBeRunning() bool {
-//	    return !s.ShutdownRequested && s.ParentMappedState == config.DesiredStateRunning
-//	}
+// Child workers do not gate on a parent-supplied field: the supervisor's
+// disable-mapping pass sets Disabled from the parent's ChildSpec.Enabled, and
+// state files route lifecycle through snap.ShouldStop().
```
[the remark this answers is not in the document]
JeremyTheocharis: The example is correct as written — `ShouldBeRunning()` deliberately ignores `Disabled`. The only production `ShouldBeRunning()` (`workers/transport/snapshot/snapshot.go:127-133`) gates on `!ShutdownRequested && GetState()==running` and never consults `Disabled`; disable is routed independently through the supervisor's disable-mapping pass and `snap.ShouldStop()` at the state-file level (`workers/transport/state/state_stopped.go:51`). The doc prose right below the example (childspec.go:61-63) already states this. Adding `&& !s.Disabled` would make the doc diverge from the implementation, so leaving the example as-is. Resolving.

umh-core/pkg/fsmv2/config/childspec.go
```diff
@@ -43,25 +43,24 @@ import (
...
 // Correct lifecycle control:
 //   - ShutdownRequested: Inherited from this type. Set by supervisor for graceful shutdown.
-//   - ParentMappedState: For child workers only. Injected by supervisor from parent's ChildStartStates.
+//   - Disabled: Inherited from this type. Set by the supervisor's disable-mapping pass
+//     from the parent's ChildSpec.Enabled; child workers stay resident in Stopped.
 //
 // The desired lifecycle state ("running"/"stopped") is read from the user spec
 // via BaseUserSpec.GetState() in DeriveDesiredState and stored directly on the
 // outer DesiredState type (not on BaseDesiredState). Callers that used to set
 // BaseDesiredState{State: x} should instead set the State field on the embedding
```
led0nk: This guidance is stale after the PR. It tells authors to "set the State field on the embedding struct (e.g. MyDesiredState{State: x, ...})", but this PR deleted WrappedDesiredState.State / GetState(), and ValidateNoCustomLifecycleFields now forbids a custom lifecycle field — following it reintroduces a banned stringly-typed field and breaks the build/validator. The ShouldBeRunning() example just below was updated; this paragraph (and the surrounding block) still describes the deleted State model. Rewrite to the Disabled/ShutdownRequested model only.

umh-core/pkg/fsmv2/workers/transport/push/worker_test.go
```diff
@@ -209,7 +209,7 @@ var _ = Describe("PushWorker", func() {
 			Expect(err).ToNot(HaveOccurred())
 			typedRunning, okRunning := desired.(*fsmv2.WrappedDesiredState[snapshot.PushDesiredState])
 			Expect(okRunning).To(BeTrue())
-			Expect(typedRunning.GetState()).To(Equal("running"))
+			Expect(typedRunning).NotTo(BeNil())
```
led0nk: Tautological assertion under a state-claiming title. DeriveDesiredState returns the same &WrappedDesiredState[...]{} for `state: stopped` and `state: running` (only ParseUserSpec errors differ), so NotTo(BeNil()) after the comma-ok BeTrue() asserts nothing about state, and the differing `state:` inputs flow nowhere. The It title implies a state check that does not happen — a state-derivation regression would pass. Rename to what it verifies (error-free parse of a valid spec) or assert a surviving field on the derived state. Same pattern in pull/worker_test.go and transport/worker_test.go.

umh-core/pkg/fsmv2/workers/transport/pull/worker_test.go
```diff
@@ -216,7 +216,7 @@ var _ = Describe("PullWorker", func() {
 			Expect(err).ToNot(HaveOccurred())
 			typedRunning, okRunning := desired.(*fsmv2.WrappedDesiredState[snapshot.PullDesiredState])
 			Expect(okRunning).To(BeTrue())
-			Expect(typedRunning.GetState()).To(Equal("running"))
+			Expect(typedRunning).NotTo(BeNil())
```
led0nk: Same tautological pattern as push/worker_test.go: NotTo(BeNil()) after the comma-ok check asserts nothing; `state: stopped` vs `running` inputs are not distinguished by DeriveDesiredState. Rename or assert a real field.

umh-core/pkg/fsmv2/workers/transport/worker_test.go
```diff
@@ -213,7 +213,7 @@ relayURL: "https://relay.example.com"`,
 				Expect(err).ToNot(HaveOccurred())
 				transportDesired, ok := desired.(*fsmv2.WrappedDesiredState[snapshot.TransportDesiredState])
 				Expect(ok).To(BeTrue())
-				Expect(transportDesired.GetState()).To(Equal("stopped"))
+				Expect(transportDesired).NotTo(BeNil())
```
led0nk: Same tautological pattern: the differing spec inputs flow nowhere and NotTo(BeNil()) is vacuous after the successful type assertion. Rename to reflect the parse-success check or assert a meaningful derived field.

led0nk:
README.md:468 (not in this diff, noting here): the field list "ChildSpec fields (Enabled, ChildStartStates, UserSpec, Variables)" is stale — this PR removes ChildStartStates from ChildSpec, and Variables is not a direct field (it lives under UserSpec.Variables). Actual fields: Enabled, UserSpec, Name, WorkerType, Dependencies. Worth folding into the same doc-sweep as the childspec.go godoc fix.

umh-core/pkg/fsmv2/config/childspec.go
JeremyTheocharis: Fixed in 8187c8565. Rewrote the paragraph to the actual post-L5b model: desired lifecycle state lives on the user spec — embed `BaseUserSpec` to inherit `GetState()`, which `DeriveDesiredState` reads and validates (worker_base.go:144). `WrappedDesiredState` carries no `State` field, and the godoc no longer tells authors to add one, so it is consistent with the "do not add a State field" rule directly above. Resolving.

umh-core/pkg/fsmv2/workers/transport/push/worker_test.go
JeremyTheocharis: Agreed. The worker returns an empty `&WrappedDesiredState[PushDesiredState]{}` for every valid spec (worker.go:138) — push is a leaf child with no lifecycle State of its own, gated by the parent via Disabled — so there is no derived field to assert. Renamed both `It`s to the parse-success check they actually perform and dropped the vacuous `NotTo(BeNil())`. Fixed in 8187c8565. Resolving.

umh-core/pkg/fsmv2/workers/transport/pull/worker_test.go
JeremyTheocharis: Same fix as push (8187c8565): pull is likewise a leaf child carrying no lifecycle State, so renamed both `It`s to the parse-success check and dropped the vacuous `NotTo(BeNil())`. Resolving.

umh-core/pkg/fsmv2/workers/transport/worker_test.go
JeremyTheocharis: Fixed in 8187c8565. The parent never populates `Config.State` (that field is deprecated/never-populated — see snapshot.go), but it does carry the credentials through, so renamed both `It`s and now assert the surviving fields: `Config.RelayURL`/`InstanceUUID`/`AuthToken` propagate on the running spec, and a stopped spec parses without requiring them. A drop-on-derive regression now fails. Resolving.

led0nk (approved):
looks good now

PR #2590, opened 2026-06-02
feat(fsmv2/L5d): carry transport auth on the typed snapshot, not parent deps

JeremyTheocharis:
## The stack

> **FSMv2 migration-API stack, PR 2 of 6.** A worker acts on its *snapshot* (its
> 2. **#2590 ← you are here**: put each transport worker's auth on its own spec, not a side channel into the parent.
>

### This PR: give each transport worker its own auth

A push or pull worker reads its JWT from the parent supervisor's in-memory state
directly, not from its own spec. So the spec does not fully describe the worker: two
workers with identical specs can behave differently depending on what the parent happens
to be holding, a worker cannot be run or tested on its own, and the token it acts on can
disagree with the rest of the config it was handed. This PR has the parent stamp the
token into each worker's own config, so its spec is its complete input.

**Prerequisite because:** the migration API only controls a worker if setting its spec
sets *everything* the worker needs; auth was the last input still arriving through a side
channel.

---

## Problem

Push and pull get the JWT by calling `GetJWTToken()` on the parent's deps. This isn't a live bug, but:

- **It breaks the framework's model.** Every other worker takes its inputs from its own snapshot; this is the only worker that reaches into another worker's deps. As a result the token isn't in the child's persisted state, so you can't tell from CSE or logs why a child pushed or stalled.
- **The child can't be tested on its own.** Exercising the token gate means building and mutating the parent's deps (`parentDeps.SetJWT(...)`).
- **The child can't work without the parent's internals.** It holds a pointer into the parent's mutable deps instead of receiving the token as its own config.

## What we're shipping

A typed `AuthSession` (token, expiry, backend UUID) the parent writes onto each child's spec in `RenderChildren`; the child reads it from its own snapshot and calls `AuthSession.IsUsable(1m)` to decide whether to send. `GetJWTToken`, `GetAuthenticatedUUID`, and `IsTokenValid` are deleted from parent and children. The token is still written to the parent's deps by the auth action; only how the children receive it changes. (`AuthSession` is in `types` because `snapshot` can't import `push`/`pull`.)

## Behavior change

Children read a snapshot copy of the token instead of a live parent read, so a refreshed token reaches a child one or two reconciles late. The supervisor re-applies each child's spec every reconcile (`updateUserSpec`), so already-running children get the new token. On expiry that's covered: the child's 1-minute buffer is shorter than the parent's 10-minute refresh. On revocation the child uses the stale token for a tick or two (a few auth failures, no messages dropped).

## Scope

- **In:** transport package + one `cmd/main.go` reader.
- **Deferred (ENG-4799):** dropping `TransportUserSpec.InstanceUUID`; a separate change, never sent on the wire.
- **ENG-4405:** the token is now also in both child configs (CSE-synced), not just parent status, so the secret-tier scrub must cover all three (TODO updated).
- Stacks on `feature/wapi-L5b` (#2589); lands after #2588. Draft until then.

### DIFF
--- a/umh-core/pkg/fsmv2/workers/transport/dependencies.go
+++ b/umh-core/pkg/fsmv2/workers/transport/dependencies.go
@@ -129,26 +129,6 @@ func (d *TransportDependencies) SetJWT(token string, expiry time.Time) {
 	d.authSession.Expiry = expiry
 }
 
-// GetJWTToken returns the stored JWT token.
-// Kept for child-deps delegation (push/pull still call parentDeps.GetJWTToken).
-// Removed in a later task once child deps no longer delegate.
-func (d *TransportDependencies) GetJWTToken() string {
-	d.mu.RLock()
-	defer d.mu.RUnlock()
-
-	return d.authSession.Token
-}
-
-// GetJWTExpiry returns the stored JWT expiry time.
-// Kept for child-deps delegation (push/pull still call parentDeps.GetJWTExpiry).
-// Removed in a later task once child deps no longer delegate.
-func (d *TransportDependencies) GetJWTExpiry() time.Time {
-	d.mu.RLock()
-	defer d.mu.RUnlock()
-
-	return d.authSession.Expiry
-}
-
 // GetAuthSession returns the current auth bundle under the read lock. Parent COS
 // uses this single locked read; per-field reads would race the SetJWT /
 // SetAuthenticatedUUID writers.
@@ -326,16 +306,6 @@ func (d *TransportDependencies) SetAuthenticatedUUID(uuid string) {
 	defer d.mu.Unlock()
 
 	d.authSession.InstanceUUID = uuid
-}
-
-// GetAuthenticatedUUID returns the stored UUID from backend authentication.
-// Kept for child-deps delegation (push still calls parentDeps.GetAuthenticatedUUID).
-// Removed in a later task once child deps no longer delegate.
-func (d *TransportDependencies) GetAuthenticatedUUID() string {
-	d.mu.RLock()
-	defer d.mu.RUnlock()
-
-	return d.authSession.InstanceUUID
 }
 
 // GetResetGeneration returns the current reset generation counter.
--- a/umh-core/pkg/fsmv2/workers/transport/dependencies.go
+++ b/umh-core/pkg/fsmv2/workers/transport/dependencies.go
@@ -58,7 +58,7 @@ var AuthFailureRateConfig = failurerate.Config{
 
 // TransportDependencies provides transport and channel access for transport worker actions.
 type TransportDependencies struct {
-	jwtExpiry         time.Time
+	authSession       types.AuthSession
 	lastAuthAttemptAt time.Time
 
 	transport types.Transport
@@ -67,8 +67,6 @@ type TransportDependencies struct {
 	authFailureRate *failurerate.Tracker
 	inboundChan     chan<- *types.UMHMessage
 	outboundChan    <-chan *types.UMHMessage
-	jwtToken        string
-	instanceUUID    string
 
 	failedAuthToken    string
 	failedRelayURL     string
@@ -127,24 +125,38 @@ func (d *TransportDependencies) SetJWT(token string, expiry time.Time) {
 	d.mu.Lock()
 	defer d.mu.Unlock()
 
-	d.jwtToken = token
-	d.jwtExpiry = expiry
+	d.authSession.Token = token
+	d.authSession.Expiry = expiry
 }
 
 // GetJWTToken returns the stored JWT token.
+// Kept for child-deps delegation (push/pull still call parentDeps.GetJWTToken).
+// Removed in a later task once child deps no longer delegate.
 func (d *TransportDependencies) GetJWTToken() string {
 	d.mu.RLock()
 	defer d.mu.RUnlock()
 
-	return d.jwtToken
+	return d.authSession.Token
 }
 
 // GetJWTExpiry returns the stored JWT expiry time.
+// Kept for child-deps delegation (push/pull still call parentDeps.GetJWTExpiry).
+// Removed in a later task once child deps no longer delegate.
 func (d *TransportDependencies) GetJWTExpiry() time.Time {
 	d.mu.RLock()
 	defer d.mu.RUnlock()
 
-	return d.jwtExpiry
+	return d.authSession.Expiry
+}
+
+// GetAuthSession returns the current auth bundle under the read lock. Parent COS
+// uses this single locked read; per-field reads would race the SetJWT /
+// SetAuthenticatedUUID writers.
+func (d *TransportDependencies) GetAuthSession() types.AuthSession {
+	d.mu.RLock()
+	defer d.mu.RUnlock()
+
+	return d.authSession
 }
 
 // RecordError increments consecutive errors and records when degraded mode started.
@@ -313,15 +325,17 @@ func (d *TransportDependencies) SetAuthenticatedUUID(uuid string) {
 	d.mu.Lock()
 	defer d.mu.Unlock()
 
-	d.instanceUUID = uuid
+	d.authSession.InstanceUUID = uuid
 }
 
 // GetAuthenticatedUUID returns the stored UUID from backend authentication.
+// Kept for child-deps delegation (push still calls parentDeps.GetAuthenticatedUUID).
+// Removed in a later task once child deps no longer delegate.
 func (d *TransportDependencies) GetAuthenticatedUUID() string {
 	d.mu.RLock()
 	defer d.mu.RUnlock()
 
-	return d.instanceUUID
+	return d.authSession.InstanceUUID
 }
 
 // GetResetGeneration returns the current reset generation counter.
--- a/umh-core/pkg/fsmv2/workers/transport/pull/action/pull.go
+++ b/umh-core/pkg/fsmv2/workers/transport/pull/action/pull.go
@@ -31,6 +31,16 @@ const PullActionName = "pull"
 // PullAction pulls inbound messages from the backend via long-poll and delivers
 // them to the inbound channel. It manages a pending-message buffer for partial
 // deliveries and applies backpressure when the channel nears capacity.
+//
+// Precondition: JWTToken is non-empty. The dispatching state validates it:
+// RunningState and DegradedState only emit a PullAction when
+// snap.Status.HasTransport && snap.Status.HasValidToken is true
+// (state_running.go, state_degraded.go), so
+// Execute never runs with an empty token in production. Do not construct a
+// PullAction outside that gate. There is intentionally no in-action
+// revalidation: an action reaching Execute with an empty token would mean the
+// dispatch gate regressed, and that should surface rather than be silently
+// turned into a no-op.
 type PullAction struct {
 	JWTToken string
 }
--- a/umh-core/pkg/fsmv2/workers/transport/pull/action/pull.go
+++ b/umh-core/pkg/fsmv2/workers/transport/pull/action/pull.go
@@ -31,7 +31,9 @@ const PullActionName = "pull"
 // PullAction pulls inbound messages from the backend via long-poll and delivers
 // them to the inbound channel. It manages a pending-message buffer for partial
 // deliveries and applies backpressure when the channel nears capacity.
-type PullAction struct{}
+type PullAction struct {
+	JWTToken string
+}
 
 // Execute runs one pull cycle: delivers pending messages, checks backpressure,
 // pulls new messages from the backend, and delivers them to the inbound channel.
@@ -45,10 +47,6 @@ func (a *PullAction) Execute(ctx context.Context, depsAny any) error {
 	pullDeps, ok := depsAny.(snapshot.PullDependencies)
 	if !ok {
 		return errors.New("invalid dependencies type: expected PullDependencies")
-	}
-
-	if !pullDeps.IsTokenValid() {
-		return errors.New("token not valid, skipping pull")
 	}
 
 	t := pullDeps.GetTransport()
@@ -142,7 +140,7 @@ func (a *PullAction) Execute(ctx context.Context, depsAny any) error {
 	}
 
 	// Phase 3: Pull from backend
-	jwtToken := pullDeps.GetJWTToken()
+	jwtToken := a.JWTToken
 
 	pullStart := time.Now()
 	messages, err := t.Pull(ctx, jwtToken)
--- a/umh-core/pkg/fsmv2/workers/transport/pull/snapshot/snapshot.go
+++ b/umh-core/pkg/fsmv2/workers/transport/pull/snapshot/snapshot.go
@@ -29,7 +29,6 @@ type PullDependencies interface {
 	GetInboundChan() chan<- *types.UMHMessage
 	GetInboundChanStats() (capacity int, length int)
 	GetTransport() types.Transport
-	GetJWTToken() string
 	RecordTypedError(errType types.ErrorType, retryAfter time.Duration)
 	RecordSuccess()
 	RecordError()
@@ -43,8 +42,6 @@ type PullDependencies interface {
 
 	IsBackpressured() bool
 	SetBackpressured(v bool)
-
-	IsTokenValid() bool
 
 	GetResetGeneration() uint64
 	CheckAndClearOnReset() bool
--- a/umh-core/pkg/fsmv2/workers/transport/pull/snapshot/snapshot.go
+++ b/umh-core/pkg/fsmv2/workers/transport/pull/snapshot/snapshot.go
@@ -56,7 +56,8 @@ type PullDependencies interface {
 
 // PullDesiredState represents the target configuration for the pull worker.
 type PullDesiredState struct {
-	State string `json:"state" yaml:"state"`
+	AuthSession types.AuthSession `json:"auth_session,omitempty" yaml:"auth_session,omitempty"`
+	State       string            `json:"state" yaml:"state"`
 	config.BaseDesiredState
 }
 
--- a/umh-core/pkg/fsmv2/workers/transport/pull/snapshot/snapshot.go
+++ b/umh-core/pkg/fsmv2/workers/transport/pull/snapshot/snapshot.go
@@ -59,7 +59,9 @@ type PullDesiredState struct {
 }
 
 // GetState returns the desired lifecycle state, defaulting to "running" if empty.
-// For pull workers this is propagated from the parent's user spec.
+// Pull is a leaf child whose lifecycle is gated by the parent transport worker via
+// the Disabled bit, not by its own State. The transport worker therefore leaves State
+// unset when constructing PullDesiredState, and GetState defaults to running.
 func (s *PullDesiredState) GetState() string {
 	if s.State == "" {
 		return config.DesiredStateRunning
--- a/umh-core/pkg/fsmv2/workers/transport/pull/snapshot/snapshot.go
+++ b/umh-core/pkg/fsmv2/workers/transport/pull/snapshot/snapshot.go
@@ -57,7 +57,7 @@ type PullDependencies interface {
 // PullDesiredState represents the target configuration for the pull worker.
 type PullDesiredState struct {
 	AuthSession types.AuthSession `json:"auth_session,omitempty" yaml:"auth_session,omitempty"`
-	State       string            `json:"state" yaml:"state"`
+	State       string            `json:"state"                  yaml:"state"`
 	config.BaseDesiredState
 }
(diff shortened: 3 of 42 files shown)


umh-core/pkg/fsmv2/workers/transport/pull/snapshot/snapshot.go
```diff
@@ -56,7 +53,8 @@ type PullDependencies interface {
 
 // PullDesiredState represents the target configuration for the pull worker.
 type PullDesiredState struct {
-	State string `json:"state" yaml:"state"`
+	AuthSession types.AuthSession `json:"auth_session,omitempty" yaml:"auth_session,omitempty"`
+	State       string            `json:"state"                  yaml:"state"`
```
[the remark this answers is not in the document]
JeremyTheocharis: Acknowledged. The bearer token being carried in CSE sync payloads is a pre-existing, tracked concern — see the `TODO(security)` in `workers/transport/snapshot/snapshot.go` and **ENG-4405** (CSE secret-tier: persist locally, exclude from delta sync). To be precise: L5d does *widen* the persisted surface — the token, already CSE-synced via the parent transport status, is now also carried on the push/pull child desired-states — but it crosses no new trust boundary (same CSE sync) and introduces no new exposure class. The scrub belongs in ENG-4405's secret-tier work, which must cover the parent status and both child copies, not this refactor. Tracking there.

umh-core/pkg/fsmv2/workers/transport/push/snapshot/snapshot.go
```diff
@@ -56,7 +51,8 @@ type PushDependencies interface {
 
 // PushDesiredState represents the target configuration for the push worker.
 type PushDesiredState struct {
-	State string `json:"state" yaml:"state"`
+	AuthSession types.AuthSession `json:"auth_session,omitempty" yaml:"auth_session,omitempty"`
+	State       string            `json:"state"                  yaml:"state"`
```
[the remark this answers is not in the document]
JeremyTheocharis: Acknowledged. The bearer token being carried in CSE sync payloads is a pre-existing, tracked concern — see the `TODO(security)` in `workers/transport/snapshot/snapshot.go` and **ENG-4405** (CSE secret-tier: persist locally, exclude from delta sync). To be precise: L5d does *widen* the persisted surface — the token, already CSE-synced via the parent transport status, is now also carried on the push/pull child desired-states — but it crosses no new trust boundary (same CSE sync) and introduces no new exposure class. The scrub belongs in ENG-4405's secret-tier work, which must cover the parent status and both child copies, not this refactor. Tracking there.

umh-core/pkg/fsmv2/workers/transport/snapshot/children.go
```diff
@@ -18,28 +18,25 @@ import (
...
 
-	pushSpec, err := config.NewChildSpec("push", "push", config.BaseUserSpec{}, enabled)
+	pushSpec, err := config.NewChildSpec("push", "push", carrier, enabled)
 	if err != nil {
 		return nil, fmt.Errorf("transport RenderChildren: push: %w", err)
 	}
 
-	pullSpec, err := config.NewChildSpec("pull", "pull", config.BaseUserSpec{}, enabled)
+	pullSpec, err := config.NewChildSpec("pull", "pull", carrier, enabled)
 	if err != nil {
```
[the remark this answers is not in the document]
JeremyTheocharis: Acknowledged. The bearer token being carried in CSE sync payloads is a pre-existing, tracked concern — see the `TODO(security)` in `workers/transport/snapshot/snapshot.go` and **ENG-4405** (CSE secret-tier: persist locally, exclude from delta sync). To be precise: L5d does *widen* the persisted surface — the token, already CSE-synced via the parent transport status, is now also carried on the push/pull child desired-states — but it crosses no new trust boundary (same CSE sync) and introduces no new exposure class. The scrub belongs in ENG-4405's secret-tier work, which must cover the parent status and both child copies, not this refactor. Tracking there.

umh-core/pkg/fsmv2/workers/transport/types/types.go
```diff
@@ -25,6 +25,39 @@ import (
...
+// TransportDependencies, not enforced by the struct. Children read the session from
+// their own snapshot instead of reaching into parent dependencies.
+//
+// Defined here (a leaf package imported by snapshot, push, and pull) to avoid an
+// import cycle: RenderChildren (in the snapshot package) stamps it and push/pull
+// parse it, but snapshot cannot import push/pull.
+type AuthSession struct {
+	Expiry       time.Time `json:"expiry,omitempty"       yaml:"expiry,omitempty"`
+	Token        string    `json:"token,omitempty"        yaml:"token,omitempty"`
+	InstanceUUID string    `json:"instanceUUID,omitempty" yaml:"instanceUUID,omitempty"`
```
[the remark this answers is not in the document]
JeremyTheocharis: Acknowledged. The bearer token being carried in CSE sync payloads is a pre-existing, tracked concern — see the `TODO(security)` in `workers/transport/snapshot/snapshot.go` and **ENG-4405** (CSE secret-tier: persist locally, exclude from delta sync). To be precise: L5d does *widen* the persisted surface — the token, already CSE-synced via the parent transport status, is now also carried on the push/pull child desired-states — but it crosses no new trust boundary (same CSE sync) and introduces no new exposure class. The scrub belongs in ENG-4405's secret-tier work, which must cover the parent status and both child copies, not this refactor. Tracking there.

umh-core/pkg/fsmv2/workers/transport/pull/worker.go
```diff
@@ -128,14 +134,16 @@ func (w *PullWorker) DeriveDesiredState(spec interface{}) (fsmv2.DesiredState, e
...
+	// stamped by the parent transport worker.
+	parsed, err := config.ParseUserSpec[PullUserSpec](renderedSpec)
+	if err != nil {
 		return nil, err
 	}
 
-	return &fsmv2.WrappedDesiredState[snapshot.PullDesiredState]{}, nil
+	return &fsmv2.WrappedDesiredState[snapshot.PullDesiredState]{
+		Config: snapshot.PullDesiredState{AuthSession: parsed.AuthSession},
+	}, nil
```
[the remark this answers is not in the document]
JeremyTheocharis: Intentional — not a dropped field. Pull/push are leaf children whose lifecycle is gated by the parent transport worker via the `Disabled` bit (disable-mapping pass), not by their own `State`. Pre-L5d these workers returned an empty `WrappedDesiredState{}` for exactly this reason; L5d adds `AuthSession` but correctly continues to leave `State` unset, so `GetState()` defaults to running. Adding `State: parsed.State` would introduce a self-gating path these children must not have. `ae567d941` restores a godoc note on `GetState()` documenting this so it's no longer a surprise. Resolving.

umh-core/pkg/fsmv2/workers/transport/push/action/push.go
```diff
@@ -27,7 +27,10 @@ import (
 
 const PushActionName = "push"
 
-type PushAction struct{}
+type PushAction struct {
```
led0nk, line 41: PushAction.JWTToken/InstanceUUID are bare exported strings and the in-action IsTokenValid() prechecks were removed — token validity now lives only in the state Next() methods behind snap.Status.HasValidToken. Honored today, but a future caller doing &PushAction{} with an empty token burns a real HTTP round-trip and then calls RecordSuccess on it, diluting the failure-rate window — violates the package invariant (Record* only after real transport ops). Make the empty-token state unrepresentable: carry the typed AuthSession (or a non-empty-token newtype) on the action and validate at the top of Execute, returning early without any Record* call when the token is empty. Same shape in pull/action/pull.go.

umh-core/pkg/fsmv2/workers/transport/pull/action/pull.go
```diff
@@ -31,7 +31,9 @@ const PullActionName = "pull"
 // PullAction pulls inbound messages from the backend via long-poll and delivers
 // them to the inbound channel. It manages a pending-message buffer for partial
 // deliveries and applies backpressure when the channel nears capacity.
-type PullAction struct{}
+type PullAction struct {
```
led0nk, line 44: Same as push/action/push.go: PullAction.JWTToken is a bare string with the in-action validity precheck removed. A future &PullAction{} with an empty token would burn a transport round-trip + Record*. Carry/validate a typed AuthSession at the top of Execute rather than relying on call-site discipline across the state files.

umh-core/pkg/fsmv2/workers/transport/types/types.go
```diff
@@ -25,6 +25,39 @@ import (
...
+// AuthSession groups the auth fields that travel together from parent transport status
+// to push/pull child config: the JWT bearer token, its expiry, and the backend-confirmed
+// instance UUID. Atomic token+expiry update is provided by the SetJWT setter on
+// TransportDependencies, not enforced by the struct. Children read the session from
+// their own snapshot instead of reaching into parent dependencies.
+//
+// Defined here (a leaf package imported by snapshot, push, and pull) to avoid an
+// import cycle: RenderChildren (in the snapshot package) stamps it and push/pull
+// parse it, but snapshot cannot import push/pull.
+type AuthSession struct {
```
led0nk, line 37: AuthSession exposes Token/Expiry/InstanceUUID as public mutable fields and the atomic invariant is enforced only by the SetJWT setter, not the struct. SetJWT writes Token+Expiry together; SetAuthenticatedUUID writes UUID under a separate lock, so a token-set-but-UUID-empty state is representable. Not a live race (per-worker COS/actions run single-threaded on the reconcile loop and GetAuthSession reads the bundle under one RLock), but the illegal state is representable. The backend returns Token+UUID+ExpiresAt together (AuthResponse) — unexport the fields and construct only via a NewAuthSession(authResp) so Token/Expiry cannot diverge and UUID cannot be set on an empty-token session.

umh-core/pkg/fsmv2/workers/transport/snapshot/snapshot.go
```diff
@@ -149,44 +145,57 @@ func (f FailedAuthConfig) IsEmpty() bool {
...
-	LastAuthAttemptAt   time.Time        `json:"last_auth_attempt_at,omitempty"`
-	FailedAuthConfig    FailedAuthConfig `json:"failed_auth_config,omitempty"`
-	JWTToken            string           `json:"jwt_token,omitempty"`
-	AuthenticatedUUID   string           `json:"authenticated_uuid,omitempty"`
-	LastErrorType       types.ErrorType  `json:"last_error_type"`
-	LastRetryAfter      time.Duration    `json:"last_retry_after,omitempty"`
-	TotalMessagesPushed int64            `json:"total_messages_pushed"`
-	TotalMessagesPulled int64            `json:"total_messages_pulled"`
-	ConsecutiveErrors   int              `json:"consecutive_errors"`
+	AuthSession         types.AuthSession `json:"auth_session"`
```
led0nk, line 165: This nests the previously-flat CSE keys (jwt_token/jwt_expiry/authenticated_uuid) into auth_session. The supervisor serializes observed state to CSE between ticks and reloads via LoadObservedTyped; after an upgrade, state written by the prior version carries the old flat keys, which will not deserialize into auth_session — so the parent reads an empty AuthSession on the first post-upgrade tick and re-authenticates once. Tolerable, but please confirm no CSE delta-sync/migration path assumes the old keys, and treat the one-time post-upgrade re-auth as intended.

umh-core/pkg/fsmv2/workers/transport/snapshot/snapshot_test.go
```diff
@@ -102,6 +136,73 @@ var _ = Describe("TransportStatus", func() {
 	})
 })
 
+var _ = Describe("RenderChildren", func() {
```
led0nk, line 139: RenderChildren is covered only in isolation here (calling it twice and asserting tokens differ proves it is a pure function of its status arg). The actual production path is untested end to end: the supervisor re-applying the changed UserSpec.Config to a resident push/pull child each tick so the child re-reads cfg.AuthSession.IsUsable(time.Minute) without a respawn. The 9-minute buffer gap (parent 10m IsTokenExpired vs child 1m IsUsable) is the only thing keeping a child off a token the parent already rotated — if the re-apply path silently regressed, this isolation test would not catch it. Add a supervisor/worker+children integration test: spawn transport with resident children, mutate parent AuthSession.Token, run reconcile ticks, assert the child observed state reflects the new token without a respawn.

umh-core/pkg/fsmv2/workers/transport/dependencies.go
```diff
@@ -313,15 +305,7 @@ func (d *TransportDependencies) SetAuthenticatedUUID(uuid string) {
 	d.mu.Lock()
```
led0nk, line 305: Stale-UUID on re-auth: this setter is called (from transport/action/authenticate.go) only when authResp.InstanceUUID != "", so a re-auth that returns an empty UUID leaves the PREVIOUS UUID in authSession.InstanceUUID (stale-but-present) — nothing clears it. Low impact since backend auth is JWT-based, but it is an unflagged state-divergence path distinct from the empty-UUID case. Either clear the UUID on an empty re-auth response, or assert the backend always returns one.

umh-core/pkg/fsmv2/workers/transport/push/action/push.go
JeremyTheocharis: **Background:** the push action sends queued messages to the relay and records each attempt into the failure-rate window, which is meant to count only real transport calls. The token it uses is validated upstream: the Running and Degraded states only emit a `PushAction` when `snap.Status.HasTransport && snap.Status.HasValidToken` is true (`state_running.go`, `state_degraded.go`). That dispatch gate is the single precondition, and the in-action `IsTokenValid()` check was removed on purpose as a redundant second copy of it.

You're right, though, that nothing at the action itself says so, which is the call-site-discipline worry: a future state that built a `PushAction` outside that gate wouldn't be caught. I'd rather not re-add a runtime check for it, because a silent early return would hide the bug you'd actually want to see (an action reaching `Execute` with an empty token means the dispatch gate regressed, and that should surface, not become a no-op tick). Instead I'm making the precondition explicit at the action: a doc comment on `PushAction` saying `JWTToken` is non-empty by contract, validated by the dispatching state, and that the action shouldn't be constructed outside that gate, with a note on why there's deliberately no in-action recheck. Same on `PullAction`. That removes the implicit-discipline part without weakening the gate or masking a regression.

umh-core/pkg/fsmv2/workers/transport/pull/action/pull.go
JeremyTheocharis: Same as `push/action/push.go`: the dispatch gate (`HasTransport && HasValidToken` in the Running and Degraded states) is the precondition, and I'm adding the same doc comment on `PullAction` making that contract explicit rather than re-adding a runtime check. Full reasoning on the `push/action/push.go` thread.

umh-core/pkg/fsmv2/workers/transport/types/types.go
JeremyTheocharis: **Context for anyone reading this later:** `AuthSession` is a serialized value, not an in-memory-only struct. It's produced once on the parent (the authenticate action calls `SetJWT(token, expiry)` from the backend `AuthResponse`), then it travels: `GetAuthSession()` → `RenderChildren` stamps it into `ChildAuthUserSpec` → marshaled to YAML into each child's `UserSpec.Config` → the push/pull child unmarshals it from its own snapshot every tick and builds the action from there. The `json`/`yaml` tags exist for exactly that round trip.

**How I read your request:** construct the session only via `NewAuthSession(authResp)` and unexport the fields, so token+expiry can't diverge and a UUID can't sit on an empty-token session. The goal (no partial or empty session ever gets used) is the right one and I'm taking it. Two notes on the specifics first: token+expiry already move together today (`SetJWT`, one lock); the only decoupled field is `InstanceUUID`, and that decoupling is deliberate, because the backend returns it on some auth responses and omits it on others.

**Why I didn't put the invariant on the type:** a constructor only enforces an invariant if every path that produces the value runs through it, and for a serialized value one path never can. The child rebuilds the session by unmarshaling its config each tick, and the YAML/JSON decoders construct the struct field-by-field via reflection. They never call a constructor. So `NewAuthSession` would guard the hand-written-literal path (which production doesn't use) and miss the deserialization path (which it does, every tick on the child side), while also forcing custom marshalers on a value whose whole purpose is to marshal. On top of that, the empty session is a state we require: before the parent authenticates, the session is the zero value, and `IsTokenExpired()` treats a zero expiry as "not yet authenticated, nothing to refresh" on purpose. A constructor that rejected empty tokens would reject that valid state, so "empty-token unrepresentable" isn't reachable at the type anyway. The cross-worker read concern is already closed by other means: `d.authSession` is unexported, and children read auth from their own snapshot rather than from parent deps.

**What I'm doing instead:** the invariant is already enforced at dispatch. The Running and Degraded states only emit a push or pull action when `snap.Status.HasTransport && snap.Status.HasValidToken` is true, so an empty or partial session never reaches a transport call. On the `push.go`/`pull.go` threads I'm making that precondition explicit with a doc comment at the action, so it no longer rests on call-site discipline. I'm also handling the stale-UUID-on-empty-re-auth path from your other comment.

umh-core/pkg/fsmv2/workers/transport/snapshot/snapshot.go
JeremyTheocharis: **Background:** the parent transport worker writes its observed state (including the auth session) to CSE between ticks and reloads it the next tick via `LoadObservedTyped`. L5d changes how the auth fields are keyed in that document: the old flat keys (`jwt_token` / `jwt_expiry` / `authenticated_uuid`) are now nested under `auth_session`.

Confirmed on the upgrade path, and thanks for walking through it. I checked who reads the old flat keys, and nothing outside this package does: no MC frontend or backend code references them, and there's no CSE delta-sync or migration step that maps them. So the only effect after an upgrade is the one you described. State written by the previous version still has the flat keys, won't load into `auth_session`, and the parent sees an empty session on its first tick and authenticates once. That's intended and clears itself. On the security side, the Sentry scrubber denylists `token` (the json tag on the nested field), so the token is still redacted; `jwt_token` just becomes an unused entry in that list. I've added a code comment on the field noting the rename costs a one-time re-auth on upgrade.

umh-core/pkg/fsmv2/workers/transport/snapshot/snapshot_test.go
JeremyTheocharis: **Background:** `RenderChildren` stamps the parent's current auth session into each push/pull child's config every tick. When the parent rotates the token, the supervisor re-applies the changed config to the already-running child, and the child picks up the new token without being torn down and recreated. That re-apply path is the one that runs in production.

You're right that the current test only checks `RenderChildren` in isolation (call it twice, confirm the token changes), which proves it's a pure function of its input but never runs the end-to-end re-apply path through the supervisor. I've added a supervisor-level integration test: spawn transport with resident push/pull children, rotate the parent's token with `SetJWT`, run reconcile ticks, and assert three things: each child's observed `HasValidToken` stays true throughout the rotation (no auth gap), the rotated token lands in the child's desired config, and the children stay the same instances and are never marked for removal (re-apply, not respawn). It runs through the production seam (parent COS, `RenderChildren`, `reconcileChildren`, `updateUserSpec`); breaking the resident-child re-apply makes it fail. The timing margin you noted (parent refreshing at 10 minutes via `IsTokenExpired`, child accepting down to 1 minute via `IsUsable`) is what keeps a child off a token the parent already replaced.

umh-core/pkg/fsmv2/workers/transport/dependencies.go
JeremyTheocharis: **Background:** `InstanceUUID` is the backend-assigned identity of the instance. It's set after authentication from the auth response, and the same instance keeps the same UUID across re-authentications.

You're right this is an unflagged path, so let me make the intent explicit. The setter only runs when the auth response includes a UUID, so an empty one on re-auth leaves the previous value in place. That is what we want: an omitted UUID on re-auth means "unchanged," not "this instance no longer has one," so keeping the previous value is correct and clearing it would be the bug, since it would drop a valid identity on a routine token refresh. What was missing is that this was implicit. I've added a comment at the caller saying an empty `InstanceUUID` on re-auth means "keep the existing one." The alternative, asserting the backend always sends it, ties us to behavior we don't control, so I'd rather document the keep.

JeremyTheocharis:
also @led0nk 

#1/#2/#4/#6 → doc commit 8c30e8a18; #5 → test commit aa9b1eb57; #3 defended in-thread."

led0nk (approved):

PR #2598, opened 2026-06-10
feat(fsmv2): add a worker's desired ChildrenSpecs to WorkerSnapshot

JeremyTheocharis:
## The stack

> **FSMv2 migration-API stack, PR 3 of 6.** A worker acts on its *snapshot* (its
> 3. **#2598 ← you are here**: put the config-declared child list on the snapshot so the migration API can read it.
>

### This PR: put the config-declared child list where the API can read it

A worker runs off its snapshot, but the snapshot did not include the list of children
declared for it in config. The migration API (#2599) has each application worker compute
its children as the union of that config-declared list with the registry's dynamic ones,
and it computes that union from inside the worker, off the snapshot. So the declared list
has to be on the snapshot for #2599 to read it. This PR puts it there, write-only:
nothing reads it until #2599.

---

> **Stack:** based on `feature/wapi-L5d` (#2590); `eng-5088-migration-api` stacks on this.

## Problem

This PR opens the FSMv1→FSMv2 migration-API stack. Context: the FSMv2 internal refactor (the cascade behind #2588/#2589/#2590, in review below this stack) is deliberately paused so FSMv2 can earn production usage first. This stack is that usage push.

The target consumer is the Probe worker (ENG-4839): one FSMv2 worker per network connection, created and removed as users edit config.yaml, replacing the Nmap connection check. That requires FSMv1 — which owns config.yaml — to start, reconfigure, and stop FSMv2 workers at runtime. Today an FSMv2 worker runs only if it is hardcoded in the application root's YAML children template.

Two known debts sit underneath. First, the application's children are declared as raw YAML, making it the only type-agnostic worker; typing them is ENG-5097. Second, the supervisor resolves children from two sources: what a state's `Next()` returns, with a fallback to the `DeriveDesiredState`-computed list when a state returns nil. The fallback was left transitional by the children-rendering cutover; its deletion is ENG-5115, unblocked once the migration-API PR (stacked directly on this one) has every worker rendering from `Next()`.

The specific gap this PR fixes: the application's states must take over child-rendering (emit "the declared children plus dynamically registered ones"), but the snapshot passed to `Next()` does not carry the declared children, so a state cannot re-emit them.

## What this ships

`WorkerSnapshot` gains a `ChildrenSpecs` field carrying the worker's declared children, filled in `ConvertWorkerSnapshot`. Nothing reads it yet; no behavior change; the stacked migration-API PR is the first reader. This is deliberate scaffolding: it lives exactly as long as YAML-declared children do — ENG-5097 deletes it together with the raw-YAML carrier, once states read declared children from typed config directly.

## What this does NOT ship

- **NOT any reader of the field**: the stacked migration-API PR is the first one.
- **NOT the typed replacement** (ENG-5097): typed child structure and the raw-YAML carrier deletion.

Contributes the minimal slice of ENG-5097. **Do not close ENG-5097.**

### DIFF
--- a/umh-core/pkg/fsmv2/api.go
+++ b/umh-core/pkg/fsmv2/api.go
@@ -482,10 +482,18 @@ func (d *WrappedDesiredState[TConfig]) GetChildrenSpecs() []config.ChildSpec {
 // WorkerSnapshot is the typed snapshot passed to State.Next(). It eliminates
 // unsafe type assertions previously required in every state file.
 type WorkerSnapshot[TConfig any, TStatus any] struct {
-	CollectedAt         time.Time
-	Config              TConfig
-	Status              TStatus
-	ChildrenView        config.ChildrenView
+	CollectedAt  time.Time
+	Config       TConfig
+	Status       TStatus
+	ChildrenView config.ChildrenView
+	// ChildrenSpecs is the worker's own DESIRED child set, copied verbatim from
+	// WrappedDesiredState.ChildrenSpecs (the legacy DeriveDesiredState path). It is
+	// NOT the effective reconciled set: for parents that declare children via the
+	// Next() Children override it does not reflect the override and may be stale or
+	// empty. It is also distinct from the observed ChildrenView above, which reports
+	// the children that actually exist. nil means "no opinion" (parent does not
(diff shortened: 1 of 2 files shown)


JeremyTheocharis:
@coderabbitai review

led0nk (approved):

PR #2600, opened 2026-06-10
Bridge edits: surface render failures and fail fast (ENG-5103)

JeremyTheocharis:
Screenshot 2026-06-11 at 15 56 07

## Problem

When a bridge edit produces a config that umh-core cannot render (template → YAML → unmarshal), the await loop retries every second for the full 30s timeout, then rolls back. The render error is logged server-side but never shown: the user sees `read DFC config not yet applied. Status reason:` with an empty reason, 30 times. ENG-5103 hit this at a customer site: 657 errors saying `yaml: line 25: did not find expected key`, with no way to see what line 25 contained, because the rendered output was discarded on failure. The render bug itself is still unreproduced; this PR makes the next occurrence diagnosable from its own error output.

## What we're shipping

1. **Render errors carry the failing region of the rendered output** (`RenderTemplate`): the reported line ± 3 lines, line-numbered, with a marker. Handles yaml's position quirks (0-based marks, multi-region `TypeError`s, errors without a position); lines truncate at 160 bytes.
2. **Fail fast**: edits that keep failing to render abort after 3 identical consecutive failures (~3s instead of 30s), roll back, and return `ErrConfigFileInvalid` naming the render failure as root cause.
3. **Per-tick visibility**: each progress reply names the render failure (`; Render failed: <first line>`); the full snippet arrives once, in the terminal message.
4. **Correctness fixes the fail-fast required**:
   - Persist and verify now share one user-variable merge. The verification render previously missed the edit's template variables and connection IP/PORT, so a valid edit could fail its render until the FSM propagated the edit; with the 3-strike abort that would have rolled back valid edits, and a connection-only edit could report success against stale values.
   - The observed-config presence guard was dead code (typed-nil interface); it is now a concrete per-side nil check, so ticks where Benthos is restarting skip the render instead of counting bogus strikes.
   - The generic `rollout_failed` Sentry event is suppressed only on the new render-failure paths (which fire their own events); the previous error-code-based gate silently downgraded the pre-existing timeout and config-error paths.

## What we're NOT shipping (known issues)

- **The render bug fix**: trigger unreproduced; item 1 is the instrument for the next occurrence.
- **`RenderTemplate` redesign** (text/template over serialized YAML is unsafe by construction): architecture work.
- **Validate-render before persisting**: first increment under ENG-4900.
- **Redaction of values in the snippet**: decided against, in favor of diagnosability. The decision covers logs, the MC status reason, and full-snippet Sentry event payloads (the commit message records it).
- **When the rollback itself fails**, the new render path returns `ErrRetryRollbackTimeout` while the pre-existing config-error path returns `ErrConfigFileInvalid`, and that branch's message and Sentry event omit the render root cause its rollback-success sibling carries. Changes frontend-visible behavior of an untouched path: reviewer's call.
- **Add-DFC-via-edit with a broken template** never populates the observed config, so fail-fast cannot engage and the edit burns the 30s timeout without a root cause (deliberate presence-guard tradeoff; the incident class, redeploy of an existing DFC, is fully covered).
- **Streak identity is full-string equality** including the snippet: a template producing per-render-varying output degrades fail-fast back to the 30s timeout, where the root cause is still named.
- Pre-existing: `deploy-dataflowcomponent` parallel-run test flake; Sentry fingerprint grouping for snippet-bearing events (should key off event name); "Save Anyway" on the failed-edit dialog.

## Testing

- Specs drive the real `Execute()`; behaviors were verified red against mutated implementations (streak reset, error-code collapse, merge removal, overlay order, presence hardcoding, per-tick append: each killed by a named spec).
- **Live verification** on a real container + production MC UI: both render-failure classes triggered through the UI, fail-fast at 3.0s (vs 30s), 3 error lines per failed edit (vs 657 in the incident), rollback verified on disk, snippet greppable per-line in the agent log. Evidence: [verification comment](https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/2600#issuecomment-4678802614).


JeremyTheocharis:
## Live verification on a real container + Management Console UI

Ran the branch image (`sha-0d5ddb0`) as a fresh instance registered to the production Management Console and drove bridge edits through the UI, to verify the three behaviors end to end.

### How the render failure was triggered

The original incident's root trigger (what made that specific config render into broken YAML) is still unknown and this PR deliberately does not chase it. For verification we manufactured failures of the same two classes, at the same code sites:

1. **Template-execution failure**: edit the bridge's read flow to reference `{{ .doesnotexist }}` — a variable that exists nowhere. The verification render fails identically every tick (`missingkey=error`).
2. **YAML-unmarshal failure (the incident's class)**: seed the bridge's variables with a multiline value, `inject: "x\nbroken: ["` — harmless while unreferenced, the bridge stays healthy. Then make an ordinary edit that splices `{{ .inject }}` into a multi-line block scalar. After substitution the embedded newline puts `broken: [` at column 0, which terminates the block scalar and corrupts the rendered YAML — `yaml: line N: did not find expected ',' or ']'`, deterministic on every tick, exactly the incident's error family.

### Raw server log, class 1 (3 identical strikes at 1s ticks, then abort + rollback)

```
07:36:19.980  [ERROR] failed to render desired read DFC config: failed to build runtime config:
              failed to execute template for type ...: template: pc:8:32: executing "pc" at
              <.doesnotexist>: map has no entry for key "doesnotexist"
07:36:20.978  [ERROR] failed to render desired read DFC config: ... (byte-identical)
07:36:21.976  [ERROR] failed to render desired read DFC config: ... (byte-identical)
07:36:21.978  [WARN]  edit_protocol_converter_render_failure_rolled_back ...
07:36:21.978  [ERROR] Error executing action: Failed during rollout: protocol converter
              'eng5103-bridge' was rolled back to its previous configuration after repeated render
              failures: ... Please fix the configuration and try editing again
```

### Raw server log, class 2 (the snippet, as it lands in /data/logs/umh-core/current)

```
07:42:27.981  [ERROR] [actions/edit-protocolconverter.go:1032] failed to render desired read DFC
              config: ... yaml: line 10: did not find expected ',' or ']'
07:42:27.981  rendered output around the reported line (yaml line numbers can be off by one):
07:42:27.981       7 |             interval: 1s
07:42:27.981       8 |             mapping: |-
07:42:27.981       9 |                 root = "hello"
07:42:27.981  >   10 |                 # x
07:42:27.981      11 | broken: [
07:42:27.981      12 |     pipeline:
07:42:27.981      13 |         processors:
```

Each snippet line is its own timestamped log line (verified via `od -c`: real `0x0A` newlines), so the region is greppable. Note the `>` marker sits one line above the offending `broken: [` — the off-by-one the snippet header warns about, behaving as documented.

### What the user sees in the dialog

Per-tick progress lines carry the first line of the error from the first failing tick (`... DFC config not yet applied. State: idle, Status reason: ; Render failed: ... yaml: line 10: ...`), and the terminal "Bridge update failed" message carries the full error including the numbered snippet.

### Results

| Measure | Before this PR | Observed |
|---|---|---|
| Time from save to rollback | 30s timeout | **3.008s / 3.015s** (server-log timestamps) |
| Render-error log lines per failed edit | 657 in the original incident | **3** (the strike count) |
| User-visible cause during the wait | empty status reason | per-tick `Render failed:` line |
| Cause in the terminal message | generic timeout text | root cause + line-numbered snippet |
| Rollback | after 30s | verified on disk both classes; unrelated config (incl. the seeded variable) untouched |

A benign edit on a healthy bridge (control) completed normally with zero render errors and no rollback.

umh-core/pkg/config/templating_test.go
```diff
@@ -0,0 +1,123 @@
...
+// in their place produces a yaml.TypeError (valid YAML, failing decode).
+// omitempty keeps the zero-valued fields out of the marshalled template so
+// the injected keys do not collide with them.
+type typedRenderFixture struct {
+	A     string `yaml:"a"`
+	Port  int    `yaml:"port,omitempty"`
+	Count int    `yaml:"count,omitempty"`
+}
+
+var _ = Describe("RenderTemplate", func() {
```
rpaland, line 35: thought: With regard to the screenshot and shown logs

> for type dataflowcomponentserviceconfig.DataflowComponentServiceConfig

is an internal Go type and not relevant for the user imo and should therefor not be inside the UI. Could maybe replace with readDFC / writeDFC.

rpaland (changes requested):
LGTM - Just that thought for the UI logs

umh-core/pkg/config/templating_test.go
rpaland: See PR Image. Couldn't attach to it

JeremyTheocharis: Good catch — fixed, and went a bit further per our error-UX standard. `RenderTemplate` no longer emits the internal Go type, and the message is now plain-language with the technical detail kept at the end.

Before:
`…was rolled back…: failed to unmarshal rendered template back to type dataflowcomponentserviceconfig.DataflowComponentServiceConfig: yaml: line 10: …`

After:
> Bridge 'X' was restored to its previous working configuration because the new read flow config isn't valid YAML. Fix the highlighted line and try again.

followed by the raw `yaml: line N: …` error and the failing-region snippet. The flow is named read/write depending on which side failed, and the per-tick progress replies are de-jargoned the same way (`read flow config not yet applied … Render failed: yaml: line N: …`).

JeremyTheocharis:
Addressing the two CodeRabbit notes:

- **Collector barrier** (`edit-protocolconverter_test.go`): the capstone spec already waits on `finalFailureReply()` before inspecting `replies`, so the collector goroutine has drained by then — verified deterministic over repeated runs. No change needed.
- **Inline comments** (`templating_test.go` + the new specs): removed the redundant narrative comments; kept a couple of one-liners that document genuinely non-obvious mechanics (the yaml.v3 0-based line off-by-one, and the column-0 block-scalar termination trick) where dropping them would invite a silent breakage.

Also rebased onto current `staging`, which surfaced the write-flow schema rework — updated the write-DFC test fixture to the new structured shape (assertions unchanged); the render snippet + fail-fast are confirmed still correct for write flows.

rpaland (approved):

PR #2603, opened 2026-06-15
fix(fsmv2): drain the worker tree on SIGTERM instead of timing out (ENG-4971)

JeremyTheocharis:
## The stack

> **FSMv2 migration-API stack, PR 5 of 6.** A worker acts on its *snapshot* (its
> 5. **#2603 ← you are here**: shut down by draining instead of timing out (the bug building the proof found).
>

### This PR: shut down without losing data (the bug the proof found)

Driving a real supervisor through create, update, and shutdown is what exposed this: on
SIGTERM it did not drain its workers, it timed out (about 34 seconds against Docker's
10-second grace) and dropped in-flight work. The proof test flaked about a quarter of the
time, and that flake *was* the production bug. This PR makes shutdown an orderly drain
sized to the worker tree, plus several related defects the same investigation uncovered.

**Why it's here:** it was not planned as part of the API. Running the proof found a real
production defect before a customer did, so it merges under the proof.

---

## Background
On SIGTERM, umh-core shuts down gracefully: the FSMv2 supervisor walks its worker tree to Stopped and reaps each worker before the process exits, inside docker's 10s stop grace. Workers form a tree (application, then child supervisors, then leaf workers); each level drains its children before itself.

## Bug
The drain did not drain. On SIGTERM, workers rode their full timeout budget instead of stopping, so the process was SIGKILLed with work still in flight. The symptoms were shutdown-time data loss and a burst of `graceful_shutdown_timeout` warnings. The integration test that exercises a real drain flaked about 25% with a family of teardown warnings. Root-causing found five distinct defects, all on the production shutdown path:

1. The tick loop was cancelled by the same context that triggered shutdown, so the drain ran against a dead loop and every phase waited out its full timeout.
2. The drain budget was a flat per-level timeout that ignored tree depth, so deep trees could not finish in time.
3. A worker that was mid-action when shutdown landed never re-evaluated its shutdown check, so it never walked to Stopped.
4. An unhealthy child subtree (open infrastructure circuit breaker) suppressed the parent's worker tick, so the parent never reaped.
5. A data race overwrote the shutdown request: the per-tick desired-state re-derivation did a non-atomic read-modify-write that raced the shutdown setter and reset `ShutdownRequested` to false, so the worker never observed shutdown.

## Fix
- **Shutdown drains a live loop.** The tick loop runs on a context detached from the shutdown trigger; cancellation becomes an orderly drain against a loop that is still ticking.
- **The budget scales with tree height.** Each level sizes its budget to its subtree depth (production base 2s, so 8s at depth 4, inside the 10s grace).
- **A worker always re-evaluates shutdown.** Shutdown now bypasses the action-observation gate, the stale-data freshness gate, and the infrastructure circuit breaker, so the worker reaches its shutdown check and walks to Stopped.
- **The shutdown flag is race-free.** A per-supervisor mutex serializes the desired-state read-modify-write against the shutdown and disable setters, which now own those bits exclusively.
- **Expected teardown warnings are quieter.** Events that occur during a correct cancel-triggered teardown (collector cancellation, a parent reaping ahead of an already-drained child) drop to debug, which cuts Sentry noise; genuine hangs still warn.

## Tests
Deterministic tests pin each reap path (mid-action, open-circuit, and the flag-clobber race, the last under the race detector). The integration scenario that exercises a real drain went from about 25% flaky to 0 failures in 30 runs.

## Notes
- Stacked on the FSMv2 migration-API foundation; review and merge after the base PRs land.
- Two follow-ups filed separately: a single predicate so every pre-tick gate honors shutdown by construction, and a correction to the misleading "CSE never removes fields" comment on the store write path.


JeremyTheocharis:
@coderabbitai review

umh-core/pkg/fsmv2/supervisor/internal/collection/collector.go
```diff
@@ -239,9 +239,19 @@ func (c *Collector[TObserved]) Stop(ctx context.Context) {
...
 
 	if c.state != collectorStateRunning {
-		c.config.Logger.SentryWarn(deps.FeatureForWorker(c.config.Identity.WorkerType), c.config.Identity.HierarchyPath, "collector_stop_skipped",
-			deps.Reason("not_running"),
-			deps.String("current_state", c.state.String()))
+		// Stopping a collector that never ran or already stopped is benign during a
+		// ctx-cancel teardown (a worker reaped before its collector started, or a
+		// double-stop). Log it at Debug then; an unexpected skip while the ctx is
+		// still live stays a warn.
+		if errors.Is(ctx.Err(), context.Canceled) {
```
led0nk: This Debug downgrade is gated on errors.Is(ctx.Err(), context.Canceled), but the primary shutdown teardown caller passes a non-cancellable context: lifecycle.go:481 sets shutdownCtx := context.Background() and :488 calls collector.Stop(shutdownCtx). Background().Err() is permanently nil, so on the normal supervisor shutdown path this downgrade NEVER fires — a double-stop/already-stopped collector still emits SentryWarn, exactly the noise this PR set out to silence. The downgrade only takes effect on the reconciliation.go:1140 reap path. Pass s.ctx (the Phase-4-cancelled detached context) into the Phase-4 Stop loop, or gate the downgrade on a shutdown flag (e.g. !s.started.Load()) rather than ctx cancellation.

umh-core/pkg/fsmv2/supervisor/internal/collection/collector.go
```diff
@@ -302,8 +322,18 @@ func (c *Collector[TObserved]) CollectFinalObservation(ctx context.Context) erro
...
 
 	if err != nil {
-		c.config.Logger.SentryWarn(deps.FeatureForWorker(c.config.Identity.WorkerType), c.config.Identity.HierarchyPath, "collector_final_observation_failed",
-			deps.Err(err))
+		// A context cancellation here is the expected shutdown race: the
+		// supervisor's Phase-4 teardown cancels the collector ctx after the drain
+		// already completed, so the final observation is best-effort and no data
+		// is lost. Log it at Debug to keep Sentry quiet; any other failure stays a
+		// warn.
+		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
```
led0nk: This downgrades BOTH context.Canceled AND context.DeadlineExceeded to Debug, but the deadline source is collector.go:315 (context.WithTimeout(ctx, ObservationTimeout)) — a DeadlineExceeded here means the observer genuinely overran ObservationTimeout, i.e. a real stuck collector. CollectFinalObservation runs from the reconciliation.go:1131 reap on EVERY worker removal (routine despawns and restart cascades, not just process shutdown), where the tick ctx is live, so a true hang is now fully silent in Sentry — and on a restart/crash loop a repeatedly-timing-out collector goes completely dark. Restrict the Debug downgrade to errors.Is(err, context.Canceled) only; keep DeadlineExceeded at SentryWarn (mirroring the Stop stopTimer.C branch at :286).

umh-core/pkg/fsmv2/supervisor/lifecycle.go
```diff
@@ -351,6 +454,23 @@ func (s *Supervisor[TObserved, TDesired]) Shutdown() {
...
 
+	// Budget re-check after the Phase-4 joins: a child drain that a
+	// tick-goroutine reap entered first early-returns in Phase 2 (see the
+	// child.Shutdown comment above) and its real spend completes inside the
+	// tickLoop, invisible to the Phase-3 timer — only the total elapsed
+	// measured here catches it. An overrun that reaches this point without a
+	// drain-loop warn never came from this level's own workers exhausting an
+	// available window, so per the event-name rule above childDrainElapsed it
+	// reports as budget exhaustion.
+	if totalDrainElapsed := time.Since(drainStart); !budgetWarned && totalDrainElapsed > drainBudget {
```
led0nk: This post-Phase-4 re-check (!budgetWarned && totalDrainElapsed > drainBudget) runs after the Phase-4 joins (<-s.tickLoopDone at :451, s.metricsWg.Wait() at :455). When workers drain cleanly via the remaining==0 ticker branch (:392) near the budget boundary, budgetWarned stays false and the join overhead can tip totalDrainElapsed past drainBudget — firing a spurious graceful_shutdown_budget_exhausted SentryWarn AND s.drainTimedOut.Store(true), so DrainOutcomeClean() returns false for a genuinely clean drain (corrupting the signal this PR introduces). Compare against the actual Phase-3 deadline (the joins are bookkeeping, not drain time), or add a small slack to the post-join comparison.

umh-core/pkg/fsmv2/supervisor/internal/collection/collector.go
```diff
@@ -260,8 +270,18 @@ func (c *Collector[TObserved]) Stop(ctx context.Context) {
...
 			deps.String("result", "success"))
 	case <-ctx.Done():
-		c.config.Logger.SentryWarn(deps.FeatureForWorker(c.config.Identity.WorkerType), c.config.Identity.HierarchyPath, "collector_stopped",
-			deps.String("result", "context_cancelled"))
+		// Phase-4 teardown cancels this ctx and races the collector goroutine's
+		// own exit; the goroutine drains right after. This is the expected
+		// shutdown cancellation (the mirror of collector_stop_skipped), so log it
+		// at Debug. A non-cancellation ctx error would be unexpected and stays a
+		// warn.
+		if errors.Is(ctx.Err(), context.Canceled) {
```
led0nk, line 272: These three new context.Canceled/DeadlineExceeded -> Debug vs SentryWarn splits (here, :246, :330) have zero behavioral coverage. collector_test.go uses NewNopFSMLogger() everywhere so no test observes log severity, and the one before-Start Stop() test passes context.Background(), never exercising the cancelled-ctx Debug branch. The severity split is the entire purpose of the top commits and is the untested half — an inverted or dropped errors.Is condition (e.g. the :246 Background() defect) would re-emit a fleet-wide SentryWarn on every graceful shutdown and no test would fail. Add specs with a severity-capturing logger: cancelled ctx -> assert Debug; DeadlineExceeded on a live ctx -> assert SentryWarn; plus a real-Shutdown test asserting no collector_stop_skipped SentryWarn.

umh-core/pkg/fsmv2/supervisor/internal/collection/collector.go
JeremyTheocharis: Good catch, fixed. You're right that with `context.Background()` on the shutdown path the cancel-based downgrade never fired, so every graceful shutdown still raised the warn.

Rather than thread the cancelled context into the Phase-4 `Stop()` loop (that context also drives the `select` that waits for the collector goroutine, so reusing it for the severity decision would change the wait behaviour as a side effect), I dropped the context check on the not-running branch entirely: `collector_stop_skipped` now logs at Debug unconditionally. A `Stop()` on a not-running collector is benign for every caller the code documents (reaped before its collector started, double-stop, or background-context shutdown), so there is no caller where it should page. This also clears the same spurious warn on the reap path you noted (`reconciliation.go`, live ctx).

One consequence worth flagging: a not-running `Stop()` on a genuinely unexpected live path is now Debug too. Telling "expected teardown" apart from "anomalous live stop" cleanly wants a `StopReason` argument on `Stop()` (shutdown vs reap) rather than overloading the context, which I have filed as a follow-up rather than grow this PR. Severity is now pinned by a test across background, cancelled, deadline-exceeded, and live contexts.

umh-core/pkg/fsmv2/supervisor/internal/collection/collector.go
JeremyTheocharis: Agreed, done. Restricted the Debug downgrade to `context.Canceled` only; `context.DeadlineExceeded` now stays at SentryWarn, mirroring the `stopTimer.C` branch at `:286`. Since `CollectFinalObservation` runs from the reap on every worker removal with a live context, a deadline overrun there is a genuinely stuck collector and needs to stay visible.

I also rewrote the comment above the check: it described the supervisor's Phase-4 teardown cancelling the collector context, but the actual caller wraps the live reconcile context in a local `WithTimeout` and cancels it only after the call returns, so `context.Canceled` here comes from the reconcile context being cancelled on process shutdown. Both arms are now pinned by tests (Canceled to Debug, DeadlineExceeded to SentryWarn).

umh-core/pkg/fsmv2/supervisor/internal/collection/collector.go
```diff
@@ -302,8 +322,18 @@ func (c *Collector[TObserved]) CollectFinalObservation(ctx context.Context) erro
...
+		// already completed, so the final observation is best-effort and no data
+		// is lost. Log it at Debug to keep Sentry quiet; any other failure stays a
+		// warn.
+		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+			c.config.Logger.Debug("collector_final_observation_failed",
+				deps.Err(err))
+		} else {
+			c.config.Logger.SentryWarn(deps.FeatureForWorker(c.config.Identity.WorkerType), c.config.Identity.HierarchyPath, "collector_final_observation_failed",
+				deps.Err(err))
+		}
```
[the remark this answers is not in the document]
JeremyTheocharis: Done. The downgrade is now restricted to `context.Canceled`; `context.DeadlineExceeded` falls through to SentryWarn so a final observation that overran `ObservationTimeout` stays visible. A test pins both arms.

umh-core/pkg/fsmv2/supervisor/lifecycle.go
JeremyTheocharis: Agreed on the false positive, fixed with the slack option. I kept the post-join measurement deliberately: a child drain that a tick-goroutine reap enters first early-returns in Phase 2 and finishes inside the tickLoop, which the Phase-3 timer never sees, so total-elapsed measured after the joins is the only thing that catches those overruns. Moving the comparison to the Phase-3 deadline would lose them.

Instead the post-join re-check now compares against `drainBudget` plus one ticker interval of slack. A genuine overrun (whole seconds at production budgets) still warns; pure join bookkeeping on an otherwise clean drain no longer trips `graceful_shutdown_budget_exhausted` or corrupts `DrainOutcomeClean()`. I pulled the check into a small method so it can be unit-tested with fixed durations: a live test of that exact window races the reap ticker against the budget timer at the same instant and is inherently flaky, so the slack arithmetic is pinned directly instead.

umh-core/pkg/fsmv2/supervisor/internal/collection/collector.go
JeremyTheocharis: Agreed, this was the untested half and the reason the `:246` defect slipped through. Added a severity-capturing `FSMLogger` test double and wrote the severity specs test-first, so each one fails on the pre-fix code:

- not-running `Stop()` logs `collector_stop_skipped` at Debug across background, cancelled, deadline-exceeded, and live contexts;
- `CollectFinalObservation` logs `collector_final_observation_failed` at Debug on `context.Canceled` and at SentryWarn on `context.DeadlineExceeded`.

These are in the same commits as the two fixes.

umh-core/pkg/fsmv2/supervisor/drain_budget_cascade_test.go
```diff
@@ -0,0 +1,815 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package supervisor internal test for the graceful-drain budget cascading
+// contract (pkg/fsmv2/CLAUDE.md §"Graceful Shutdown Cascading"): each level
+// samples base × subtree height at Shutdown entry and its child drains spend
+// from that same budget, so per-level budgets do not sum — a chain of depth N
+// drains within N×base total (depth 1 = base, depth 2 = 2×base, depth 3 =
+// 3×base).
+package supervisor
```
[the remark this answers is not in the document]
JeremyTheocharis: Done. Converted the cluster to `package supervisor_test`. The five files plus `lifecycle_shutdown_test.go` share unexported test fixtures both ways (one defines `shutdownTestSyncBuffer`/`shutdownHonoringState` that the others use, and uses helpers the others define), so they had to move together. The handful of genuinely-internal reads (worker count, infra-circuit state, the post-join budget check) now go through small `TestXxx` accessors on the supervisor; every assertion is unchanged.

led0nk (approved):
checked in on your replies again, looks pretty clean then now

PR #2610, opened 2026-06-23
Demote bridge config dumps to DEBUG, report config divergence in bridge status

JeremyTheocharis:
## Problem

The bridge reconcile loop compares each bridge's observed config to its desired config every 100 ms. When they differ it re-applies and writes the full rendered config (connection settings, data-flow component configs) to the agent log at INFO.

On one customer instance with 27 bridges, the bridges re-applied every tick and these INFO dumps reached ~8,000 lines per second, a 10 MB log file every 7 seconds, on a box already burning 4 to 6 of 8 cores. Once an instance climbed into high CPU it stayed there instead of recovering. The working theory, not yet proven, is that the logging is part of what keeps it there: under load the reconcile can't finish in its time budget, an unfinished reconcile keeps re-applying and dumping full configs, and writing that much log costs the CPU the reconcile needed to catch up.

The line that would explain why the re-apply kept firing, the config diff, was at DEBUG. The only way to read it is to raise the log level, which turns the dumps back on.

## What we're shipping

- **A bounded config-diff helper.** `BoundDiff(diff, capRunes)` in the protocol-converter config package caps a diff string at 400 runes, rune-safe (the cut never splits a multibyte character), and appends a truncation marker with the total rune count. An empty diff returns a named sentinel (`configs differ but field diff is empty (divergence outside compared fields)`) instead of a blank line, so the silent-mismatch class that hid the original cause is reported explicitly. Pure, no state, no logging.

- **Full config dumps moved from INFO to DEBUG.** In both the protocol-converter and stream-processor services, the `Connection config: %+v` / `Dataflow component config: %+v` dumps on the create and update paths, plus the per-retry `Updating`, `Updated`, and `Removing` lines, now log at DEBUG. The once-per-create lifecycle lines (`Adding …`, `… added to manager`) stay at INFO, and `Adding` now logs only on a real create instead of on every idempotent retry (it moved below the already-exists check). `Removing` is demoted rather than made fire-once because the removal path retries on genuine pending state, not idempotent no-ops; a single removal stays visible at DEBUG.

- **Divergence surfaced in the bridge status, not just logs.** When a bridge's observed config differs from desired, a bounded summary of the diff shows up in that bridge's status reason in the Management Console, so an operator can see *what* is being re-applied without enabling debug logging. It's appended after (not replacing) the operational reason Step 3 already set (e.g. "active", "starting"), so both are visible. The text is cleared every tick and only set when there is a real divergence, so it never grows unbounded or leaks across ticks.

- **One heartbeat WARN per bridge per minute.** While config diverges, on ticks where `snapshot.Tick % 600 == 0` (600 ticks at 100 ms = one minute), one WARN is logged with the bridge ID and the bounded diff. It recurs rather than firing once, so a tail-kept log export taken hours into an incident still contains a recent WARN. Divergent windows that span no multiple of 600 emit no WARN.

- **Stream processors get the demotions too**, since they have the same dump sites. The status surfacing and WARN are protocol-converter only (the proven-hot path).

- **FF: none.** Additive. The only behavior change is the status text on a bridge that is actively re-applying, which now names the diverging config. No schema change, no field any automation parses, and a converged bridge's status is unchanged.

## What we're NOT shipping

- **NOT fixing the re-apply loop.** Why a bridge's observed config never converges to desired is still open under ENG-5090. This removes one of the things feeding the loop (the log flood) and makes the cause visible; it does not stop the loop. Whether a stuck instance recovers on its own is unverified until this ships.

- **NOT changing FSM states or transitions.** Only the status-reason string is composed; the state machine is untouched. The one new field is a tick-scoped string on `ObservedState`, reset every observation, nothing that persists or accumulates across ticks.

- **NOT generalized log de-noising.** No fingerprinting, no suppression counters, no throttle infrastructure, no touching the existing external-changes warning line. That is a separate effort.

- **NOT touching benthos child logs.** Only umh-core's own reconcile logging changes; the benthos process logs are untouched.

- **NOT credential redaction.** ENG-3483 stays open; moving dumps to DEBUG shrinks the exposure, it does not close it.

## Tests

A reconcile-lifecycle test drives a bridge through ~1,300 sustained-divergence ticks, two distinct deadline-injection points (before and after the status-info store), convergence, and diverge-into-removal, asserting the status reason names the diverging config while diverged, is clean once converged, and stays bounded across removal retries. Reverting the composition placement, the per-tick reset, or the removal clear makes it fail. Per-edge tests pin the `BoundDiff` contract (truncation, rune-safety, empty-diff sentinel), the reset-before-deadline-guard ordering, and the WARN cadence.

## Empirical verification

Verified the demotion in a container on real images, before/after, with an identical trigger (a data-flow component whose benthos stays unhealthy, so the reconcile re-applies every tick — the same path that produced the customer flood).

**RED** = `ghcr.io/united-manufacturing-hub/umh-core:v0.44.24` (pre-PR, the four dump lines are INFO).
**GREEN** = `ghcr.io/united-manufacturing-hub/umh-core:sha-cef1501` (this PR).

Same config, same bundled benthos, same unhealthy-DFC trigger in both runs.

On a single re-applying bridge over ~50s at `LOGGING_LEVEL=INFO`:

| Line (was INFO) | RED, at INFO | GREEN, at INFO |
|---|---|---|
| `Updating protocolconverter %s` | 515 | 0 |
| `Updated protocolconverter config in manager` | 515 | 0 |
| `Connection config: %+v` | 515 | 0 |
| `Dataflow component config: %+v` | 515 | 0 |

Same line, same trigger, now zero at INFO. The flood (the customer case had 27 bridges re-applying, each emitting several such lines per tick) is gone from the INFO log.

To confirm the lines were **demoted, not removed**, the GREEN image was re-run at `LOGGING_LEVEL=DEBUG`. The same four lines fire at DEBUG (571/571/572/572 over the window), from `protocolconverter.go:552` — the same sites, shifted by the `ctx.Err()`-guard reordering in this PR:

- RED:    `[INFO]  protocolconverter/protocolconverter.go:548  Updating protocolconverter vibration-sensor-pc`
- GREEN:  `[DEBUG] protocolconverter/protocolconverter.go:552  Updating protocolconverter vibration-sensor-pc`

Under the production INFO level, DEBUG output is suppressed, so the dumps disappear from the log without losing the re-apply behavior. The trigger (the unhealthy DFC re-applying) is unchanged between the two runs — the only difference is the log level.

The two features that make the cause visible without DEBUG were also observed in the GREEN run (absent in RED):

- the **WARN heartbeat** (`re-applying config: bridge … config diverged: …`) fires while a bridge is diverging, bounded (one per snapshot interval, recurring, so a late log export still has a recent entry);
- the bounded **config-diff** is logged at DEBUG (`Configuration differences: …`, 589 occurrences in the DEBUG run vs 0 in RED), confirming the divergence-detection path runs.

Status-reason surfacing in the Management Console UI was not exercised here (no MC in the test container) and remains covered by the reconcile-lifecycle test above.

Refs ENG-5108, ENG-5090.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ finalization/
 *.result.txt
 .worktrees/
 .self-review/
+vsdd/
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -5,6 +5,10 @@
 ### New Features
 - Preview: an instance can now store a connection to an external TimescaleDB or PostgreSQL historian database (host, port, database, login, and TLS mode), which can be created, viewed, updated, and removed. Storing the connection does not yet route any data to it
 
+### Improvements
+
+- The full bridge and stream-processor configuration is no longer written to the umh-core log at INFO on every internal re-apply (moved to DEBUG), removing a source of log flooding and CPU pressure when a config keeps being re-applied. While a bridge's observed config differs from its desired config, its status in the Management Console now shows a bounded summary of what differs, and the umh-core log carries one warning per bridge per minute, so the cause is visible without enabling debug logging
+
 ## [0.44.28]
 
 ### New Features
@@ -43,8 +47,6 @@
 - Editing a bridge whose new configuration keeps failing to render now fails after three identical render failures (a few seconds) and rolls back to the previous configuration automatically, naming the render failure as the root cause; while the edit is still retrying, each progress update also names the render failure. Previously such an edit retried for the full 30-second timeout before rolling back, and progress updates gave no reason
 
 ![Bridge update failed dialog showing the failing region of the rendered config and the automatic rollback message](./changelog-images/2026-06-19-bridge-render-error.png)
-
-- The full bridge and stream-processor configuration is no longer written to the umh-core log at INFO on every internal re-apply (moved to DEBUG), removing a source of log flooding and CPU pressure when a config keeps being re-applied. While a bridge's observed config differs from its desired config, its status in the Management Console now shows a bounded summary of what differs, and the umh-core log carries one warning per bridge per minute, so the cause is visible without enabling debug logging
 
 ### Fixes
 
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -44,6 +44,8 @@
 
 ![Bridge update failed dialog showing the failing region of the rendered config and the automatic rollback message](./changelog-images/2026-06-19-bridge-render-error.png)
 
+- The full bridge and stream-processor configuration is no longer written to the umh-core log at INFO on every internal re-apply (moved to DEBUG), removing a source of log flooding and CPU pressure when a config keeps being re-applied. While a bridge's observed config differs from its desired config, its status in the Management Console now shows a bounded summary of what differs, and the umh-core log carries one warning per bridge per minute, so the cause is visible without enabling debug logging
+
 ### Fixes
 
 - OPC UA: reading or subscribing to large tag sets (roughly a thousand or more) over an encrypted connection no longer drops the connection with an `EOF` error
--- a/umh-core/pkg/config/protocolconverterserviceconfig/bounddiff.go
+++ b/umh-core/pkg/config/protocolconverterserviceconfig/bounddiff.go
@@ -1,0 +1,54 @@
+// Copyright 2025 UMH Systems GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build test
+
+// BoundDiff has no production caller yet; the test build tag above keeps it
+// out of production binaries until the first non-test caller is added.
+
+package protocolconverterserviceconfig
+
+import (
+	"strconv"
+	"strings"
+	"unicode/utf8"
+)
+
+// BoundDiff caps a config-diff string at capRunes runes, rune-safe, appending a
+// truncation marker when the diff exceeds the cap. capRunes values <= 0 are
+// treated as 0.
+func BoundDiff(diff string, capRunes int) string {
+	if capRunes < 0 {
+		capRunes = 0
+	}
+
+	total := utf8.RuneCountInString(diff)
+	if total <= capRunes {
+		return diff
+	}
+
+	var b strings.Builder
+	i := 0
+	for _, r := range diff {
+		if i >= capRunes {
+			break
+		}
+		b.WriteRune(r)
+		i++
+	}
+	b.WriteString(" …[truncated, ")
+	b.WriteString(strconv.Itoa(total))
+	b.WriteString(" chars total]")
+	return b.String()
+}
--- a/umh-core/pkg/config/protocolconverterserviceconfig/bounddiff.go
+++ b/umh-core/pkg/config/protocolconverterserviceconfig/bounddiff.go
@@ -12,11 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build test
-
-// BoundDiff has no production caller yet; the test build tag above keeps it
-// out of production binaries until the first non-test caller is added.
-
 package protocolconverterserviceconfig
 
 import (
--- a/umh-core/pkg/config/protocolconverterserviceconfig/bounddiff.go
+++ b/umh-core/pkg/config/protocolconverterserviceconfig/bounddiff.go
@@ -38,6 +38,7 @@ func BoundDiff(diff string, capRunes int) string {
 	if diff == "" {
 		return EmptyDiffSentinel
 	}
+
 	if capRunes < 0 {
 		capRunes = 0
 	}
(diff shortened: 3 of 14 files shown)


JeremyTheocharis:
Raw log excerpts backing the "Empirical verification" section in the PR description. Both runs used an identical config and identical trigger (a data-flow component whose benthos stays unhealthy, so the reconcile re-applies every tick). Only the image differs.

- **RED** = `ghcr.io/united-manufacturing-hub/umh-core:v0.44.24` (pre-PR; four dump lines are `INFO`)
- **GREEN** = `ghcr.io/united-manufacturing-hub/umh-core:sha-cef1501` (this PR; four dump lines are `DEBUG`)

All lines below are unedited; only the per-tick repetition is elided.

---

### RED — one re-apply tick, at `LOGGING_LEVEL=INFO`

```
2026-06-24 06:11:14.297140512   [INFO]	[protocolconverter/protocolconverter.go:548:548]	[ProtocolConverterServicevibration-sensor-pc]			Updating protocolconverter vibration-sensor-pc
2026-06-24 06:11:14.297142721   [INFO]	[protocolconverter/protocolconverter.go:640:640]	[ProtocolConverterServicevibration-sensor-pc]			Updated protocolconverter config in manager
2026-06-24 06:11:14.297143887   [INFO]	[protocolconverter/protocolconverter.go:641:641]	[ProtocolConverterServicevibration-sensor-pc]			Connection config: [{FSMInstanceConfig:{Name:protocolconverter-vibration-sensor-pc DesiredFSMState:up} ConnectionServiceConfig:{NmapServiceConfig:{Target:127.0.0.1 Port:8080}}}]
2026-06-24 06:11:14.297354054   [INFO]	[protocolconverter/protocolconverter.go:642:642]	[ProtocolConverterServicevibration-sensor-pc]			Dataflow component config: [{FSMInstanceConfig:{Name:read-protocolconverter-vibration-sensor-pc DesiredFSMState:active} DataFlowComponentServiceConfig:{BenthosConfig:{Input:map[generate:map[count:0 interval:1s mapping:root = random_int(min:10, max:20) ... ]} ...]}} ...]
```

This tick repeats ~10×/s for as long as the DFC stays unhealthy (515 occurrences of each of the four lines over ~50 s on this single bridge). The full rendered connection + dataflow-component config is written to the INFO log on every tick.

---

### GREEN — the same re-apply tick, at `LOGGING_LEVEL=DEBUG`

Same four lines, same sites — now at `DEBUG`, and shifted to `protocolconverter.go:552` (the `ctx.Err()`-guard reordering in this PR moves the `Updating`/`Removing` lines below the guard):

```
2026-06-24 06:20:41.260368261   [DEBUG]	[protocolconverter/protocolconverter.go:552:552]	[ProtocolConverterServicevibration-sensor-pc]			Updating protocolconverter vibration-sensor-pc
2026-06-24 06:20:41.260415469   [DEBUG]	[protocolconverter/protocolconverter.go:640:640]	[ProtocolConverterServicevibration-sensor-pc]			Updated protocolconverter config in manager
2026-06-24 06:20:41.260416552   [DEBUG]	[protocolconverter/protocolconverter.go:641:641]	[ProtocolConverterServicevibration-sensor-pc]			Connection config: [{FSMInstanceConfig:{Name:protocolconverter-vibration-sensor-pc DesiredFSMState:up} ConnectionServiceConfig:{NmapServiceConfig:{Target:127.0.0.1 Port:8080}}}]
2026-06-24 06:20:41.260484344   [DEBUG]	[protocolconverter/protocolconverter.go:642:642]	[ProtocolConverterServicevibration-sensor-pc]			Dataflow component config: [{FSMInstanceConfig:{Name:read-protocolconverter-vibration-sensor-pc DesiredFSMState:stopped} DataFlowComponentServiceConfig:{BenthosConfig:{Input:map[generate:map[count:0 interval:1s mapping:root = random_int(min:10, max:20) ... ]} ...]}} ...]
```

Under the production `LOGGING_LEVEL=INFO`, `DEBUG` output is suppressed, so these four lines (and the full config they carry) no longer appear in the log. Re-running GREEN at DEBUG is what produces the lines above — confirming the lines were demoted, not removed. The same flood fires in GREEN (571/571/572/572 over the window); only the level changed.

---

### GREEN — the cause made visible without DEBUG

Two things that did not exist in RED now surface while a bridge is diverging:

The bounded config-diff, logged at `DEBUG` (`protocolconverter/actions.go:364`):

```
2026-06-24 06:20:41.260363969   [DEBUG]	[protocolconverter/actions.go:364:364]	[vibration-sensor-pc]			Configuration differences: Connection: {0x7e1f2c8adb20} vs {0x7e1f2c8adb40}
```

The heartbeat WARN, composed into the bridge's status reason on the snapshot interval (visible at INFO, recurring, so a late log export still has a recent entry):

```
2026-06-24 06:14:20.963751293   [INFO]	[cmd/main.go:449:449]	[SnapshotLogger]			  └─ ⚠️ vibration-sensor-pc: starting_connection → active | starting: connection is degraded; re-applying config: ReadDFC: Input config differences:
2026-06-24 06:14:25.964606337   [INFO]	[cmd/main.go:449:449]	[SnapshotLogger]			  └─ ⚠️ vibration-sensor-pc: starting_dfc → active | starting: starting - waiting for benthos to be up: waiting for config to be loaded: uptime 3 s (< 5 s threshold); re-applying config: ReadDFC: No significant differences
```

Neither line exists in the RED run.

FalconTube (changes requested):
This PR does two things
1. Report the divergence of the Config to the User -> Very good and I think we should keep that
2. Address the logging for a very specific part of the software only -> I would propose to do this differently

I would propose to do
- staggered backoff when lots of the same messages come in
- combine that with fingerprinting
- clean up the logs by moving more stuff to `debug`
- make it easier to read the logs by formatting them properly

Basically all of this can be done through our central logger and will therefore affect all current and following log usages.

I gave it in an initial shot on this branch here https://github.com/united-manufacturing-hub/united-manufacturing-hub/tree/improve-logging-wip

It does more stuff than this PR here in a much smaller package.
So please let me know, if you also find this more viable and then we can iterate on this

FalconTube (approved):
Good to merge.<br>As discussed, I will open separate logging PR, which adds

1. structured logging
2. exponential backoff logging for some parts

PR #2613, opened 2026-06-24
CPU health: two-rule verdict model + p95/p99/steal/pressure wire + stable degraded state

JeremyTheocharis:
# CPU health: two-rule verdict model + p95/p99/steal/pressure wire + stable degraded state

## Problem

A container flashes Degraded while the customer's hypervisor shows 40% average CPU, so the customer concludes UMH is broken and refuses to add CPUs. The evidence that would win that conversation (throttling, steal time, burst peaks) is partly in the status message already but never shown to anyone. (Eng-3500 flicker case, VE-230 throttling case.)

The deeper issue: umh-core's spikes create CPU pressure while average CPU stays low. The new metrics let us measure every improvement (benthos_monitor, FSMv2) against pressure, not against averages.

## How to review this

The PR is large; do not read it linearly. The suggested path:

1. `pkg/cpuhealth/doc.go` carries the whole model in ~90 lines (the two-rule ceiling, the four sub-latches, the hold-on-missing discipline). Start there.
2. `docs/production/cpu-health.md` is the operator-facing behavior (status table, remediation).
3. The scenario names in `pkg/cpuhealth/decide_test.go` and `pkg/service/container_monitor/sampler_full_sample_wire_test.go` are the edge-case index; each test comment states the contract it pins.
4. `pkg/models/status_message.go` is the wire seam; the consumer half is united-manufacturing-hub/ManagementConsole#3308 (the five Storybook scenarios there render this wire visually).
5. The verdict logic itself is `pkg/cpuhealth/decide.go` (a state machine over samples; `Decide` is the only mutator of `WindowState`).

A live container smoke test on a real cgroup v2 kernel passed all four scenarios (limit mode, no-limit, throttle under load with zero flapping, PSI present).

### The two-rule ceiling model

The verdict's ceiling follows the limit. No limit: host headroom (host busy mean vs host cores, degrade when less than one core is free; time waiting on disk I/O counts as schedulable idle, not busy). Limit set: container usage vs the quota (degrade on headroom exhaustion or throttle). This fixes the quota-scoped-capacity false-fire (container has a limit but host is busy) and the no-limit/no-host-stats blind spot (silent never-degrade).

### The saturation sub-latches

Four sub-latches each pin a specific saturation cause (limit saturation, host full, no-limit host headroom, no-host-stats fallback). The no-host-stats fallback fires at sustained 70% container usage of the machine's cores when host stats are unavailable. The status message reports them via the structured breakdown.

### The hold-on-missing discipline

Every latch is a Schmitt trigger (fires high, recovers low, no flap at the boundary). The same discipline covers missing data: a failed read is not evidence of health. Ring appends are gated on readability flags, and the container monitor holds the last verdict (state, message, basis, CgroupCores, usage) across a sampler outage instead of flapping to healthy. A held stale-degraded verdict keeps blocking bridge creation for the duration of the outage; that is the deliberate conservative choice.

### The wire

The status message now carries p95/p99 peaks, steal time, CPU pressure (PSI), and a structured host-vs-limit breakdown (`verdictBasis`). The flat `throttleRatio`/`isThrottled` mirrors moved into `verdictBasis.throttle` (no deployed consumer reads the old fields). The peaks appear in the MC tooltip (united-manufacturing-hub/ManagementConsole#3308), not the status text.

### The message (ComposeMessage)

Mode-aware, plain-language. States whether usage is measured against the host or the container's CPU limit. Reports pressure, throttling, and steal when those signals are available. Says plainly when host stats are unavailable.

### The docs

`docs/production/cpu-health.md` is restructured around the two-rule model. Leads with "CPU health is based on whether the container has a CPU limit." The status table matches the code's headlines. The 70% no-host-stats fallback is disclosed. `pkg/cpuhealth/doc.go` carries the maintainer-facing model.

## Also fixes

- The degraded state no longer flaps on transient `/proc/stat` outages and no longer reverts to a self-contradictory healthy headline on a sustained outage.
- Attribution smoothed on 60s means, not instantaneous spikes.
- The CPU panel no longer shows a garbled +Inf percentage on tiny Kubernetes CPU quotas.
- Hosts busy with disk I/O no longer count that wait as CPU load, so I/O-heavy boxes with spare compute are not falsely degraded.
- A single steal spike right after startup no longer fires a false degraded verdict (the steal latch needs a fuller window).
- Customer name scrubbed from a test comment.

## Scope

CPU only. No feature flag on the umh-core side (additive status-message fields plus a display fix; the MC panel is flag-gated on its side). No memory or disk. No graph. No log-flood detection. No per-process breakdown (ENG-3323 closed; ENG-5264 is the future follow-up). cgroup v1 hosts keep the pre-existing limitation (the sampler reads v2 paths only; CPU health is dark there and fails open).

## Deferred

- ENG-5264: per-process attribution (host, workload, software, unknown via a per-PID reader).
- ENG-5265: host-aware bridge-admission capacity (replace `(cores - 1) * 5`).
- Structural refactor (Decide extraction, Signals split, prune-loop dedupe) as a follow-up ticket.

See also for deferred items: https://linear.app/united-manufacturing-hub/issue/ENG-5384/cpu-health-rollout-tech-debt-and-deferred-hardening-evaluate-and-split

## Cross-link

united-manufacturing-hub/ManagementConsole#3308 is the MC panel that consumes this wire. Both PRs ship together; this one lands first so staging carries the emitter before the consumer.

Co-Authored-By: Claude <noreply@anthropic.com>


FalconTube:
Spoke with @Aaron99B about this PR today, he will take care of reviewing it.<br>Can re-assign me if needed

umh-core/pkg/cpuhealth/doc.go
Aaron99B, line 1: I have problems understanding this document. There are a lot of things that are not defined at the beginning but just appear e.g. NrPeriodsAvailable, what is a ring append, etc. I think it would be nice to have some brief definitions at the beginning of the document to make following it easier

umh-core/pkg/cpuhealth/doc.go
```diff
@@ -0,0 +1,100 @@
...
+//
+// High CPU usage alone is not degradation: a container that uses its full
+// quota while making progress is busy, not sick. The verdict therefore keys
+// on starvation evidence rather than utilization: CFS throttling, PSI
+// pressure, hypervisor steal time, and the host running out of spare cores.
+//
+// # The two-rule ceiling
+//
+// One question selects the rule: does the container have a CPU limit (a
+// cgroup quota)? [Decide] evaluates the limit rule first, so a limit-set
```
Aaron99B, line 25: What is this decide? Why is it in square brackets?

umh-core/pkg/cpuhealth/doc.go
```diff
@@ -0,0 +1,100 @@
...
+// sample never reaches the no-limit arms.
+//
+//   - No limit: the ceiling is the host. Headroom is LogicalCpus minus the
+//     60s mean of host-busy cores minus a one-core reserve
+//     (cpuReserveCores); the saturation latch fires when headroom drops
```
Aaron99B, line 30: Also constants like this just appear here without definition

umh-core/pkg/cpuhealth/doc.go
```diff
@@ -0,0 +1,100 @@
...
+//     below 0, that is, when the host as a whole has less than about one
+//     core free.
+//   - Limit set: the ceiling is the quota. Headroom is the quota minus the
+//     container's own 60s-avg core usage minus a fractional reserve
+//     (Thresholds.LimitReserveFraction of the quota); the latch fires when
```
Aaron99B, line 35: I am not familiar with the concept of a latch in the Schmitt trigger context. Maybe introduce this briefly as well?

umh-core/pkg/cpuhealth/doc.go
```diff
@@ -0,0 +1,100 @@
...
+//     sustained usage enters the reserve band. Sustained throttling
+//     degrades independently of headroom.
+//
+// # Causes and sub-latches
+//
+// The emitted saturation state is the OR of four internal latches, each
+// pinning a specific cause so the operator message and the wire basis can
+// name it:
+//
+//   - limitSaturationFired: limit mode, the container's sustained usage is
```
Aaron99B, line 45: I think for these different states with different attributes, a table would be more comprehensible

umh-core/pkg/cpuhealth/doc.go
```diff
@@ -0,0 +1,100 @@
...
+//     inside the quota's reserve band.
+//   - hostFullFired: limit mode, the host itself is full (host headroom
+//     below 0). A limit is a ceiling, not a reservation, so a full host
+//     stacks as a cause on a limited container. Evaluated only when
+//     /proc/stat is readable; holds across an outage.
```
Aaron99B, line 50: What is this /proc/stat?

umh-core/pkg/cpuhealth/doc.go
```diff
@@ -0,0 +1,100 @@
...
+//     core free.
+//   - Limit set: the ceiling is the quota. Headroom is the quota minus the
+//     container's own 60s-avg core usage minus a fractional reserve
+//     (Thresholds.LimitReserveFraction of the quota); the latch fires when
+//     sustained usage enters the reserve band. Sustained throttling
+//     degrades independently of headroom.
+//
+// # Causes and sub-latches
+//
+// The emitted saturation state is the OR of four internal latches, each
```
Aaron99B, line 41: if it is an OR of 4 latches, that means all of the 4 latches can happen at the same time?

umh-core/pkg/cpuhealth/doc.go
Aaron99B: I see that in the doc there are a lot of references to the code, but if I use the doc for the initial read I am missing quite some context where all teh variabels and methods that are mentioned are coming from. It would be nice to briefly outline the cpuhealth's structure here as well

umh-core/docs/production/cpu-health.md
```diff
@@ -0,0 +1,56 @@
+# CPU Health
+
+CPU health is based on one question: does this container have a CPU limit? (a Docker `--cpus` or Kubernetes CPU limit). The answer decides what UMH measures against.
+
+With no limit, the whole machine is the ceiling. UMH watches how busy the machine is (60-second average) and reports degraded when less than about one core is free. At that point everything on the box, UMH included, starts waiting for CPU (the same waiting the pressure signal measures). Next to the machine total, the view shows the container's own usage, so you can see roughly what is UMH vs. everything else.
```
Aaron99B, line 5: What is the box?

Aaron99B, line 5: I would just exclude this: "the whole machine is the ceiling"

umh-core/pkg/cpuhealth/decide.go
```diff
@@ -0,0 +1,1157 @@
...
+
+// Cause is a single degradation reason with an associated numeric value.
+type Cause struct {
+	Kind  CauseKind
+	Value float64
+}
+
+// Thresholds holds the tunable cutoffs for the Decide verdict.
+type Thresholds struct {
+	HighUsageFraction float64
```
Aaron99B, line 58: when respecting the naming schema for the other threshholds, shouldn't this be SaturationHigh?

umh-core/pkg/cpuhealth/decide.go
```diff
@@ -0,0 +1,1157 @@
...
+		PressureHigh:                0.20,
+		PressureRecover:             0.12,
+		StealHigh:                   0.10,
+		StealRecover:                0.06,
+		LimitReserveFraction:        0.10,
+		LimitReserveRecoverFraction: 0.05,
+	}
+}
+
+// throttleWindow is the sliding window length over which the throttle ratio
```
Aaron99B, line 99: I would reduce the amount of comments. A comment like this is not useful. The variable is and should be self-explanatory

umh-core/pkg/cpuhealth/decide.go
```diff
@@ -0,0 +1,1157 @@
...
+//
+// CgroupCores is the legacy non-pointer quota representation. The production
+// sampler sets Quota, not CgroupCores; the fallback is kept for callers and
+// tests that still populate it.
+type Sample struct {
+	Timestamp   time.Time
+	Quota       *float64
+	UsageCores  float64
+	CgroupCores float64
+	NrPeriods   int64
```
Aaron99B, line 162: Does Nr mean number here? Then I would also spell it out for readabiliy

umh-core/pkg/models/status_message.go
```diff
@@ -120,14 +120,97 @@ type Container struct {
...
+type Cause struct {
+	Kind  CauseKind `json:"kind"`
+	Value float64   `json:"value"`
+}
+
 type CPU struct {
-	Health         *Health `json:"health"`
-	TotalUsageMCpu float64 `json:"totalUsageMCpu"` // Total usage in milli-cores (1000m = 1 core)
-	CoreCount      int     `json:"coreCount"`      // Number of CPU cores
+	Health *Health `json:"health"`
```
Aaron99B, line 133: Why is the health optional?

umh-core/pkg/models/status_message.go
```diff
@@ -120,14 +120,97 @@ type Container struct {
...
+	// omitempty emits a real 0 (non-nil pointer) when the metric is
+	// fetchable (the ring holds >= 2 entries) and omits it when un-fetchable
+	// (nil pointer: the first ticks before the ring has 2 entries).
+	// Observability-only. They do not change the verdict.
+	AvgMCpu *float64 `json:"avgMCpu,omitempty"`
+	P95MCpu *float64 `json:"p95MCpu,omitempty"`
+	P99MCpu *float64 `json:"p99MCpu,omitempty"`
+	// State is always emitted (no omitempty), even when healthy. Attribution
+	// and Causes are set only when State == "degraded".
+	State          string  `json:"state"`                 // "healthy" | "degraded"
```
Aaron99B, line 145: Don't we have this already in the health?

Aaron99B:
When looking at the percentages of CPU used now for my instance, the percentage is changing quite fast. Shouldn't we also smooth it more?

umh-core/pkg/cpuhealth/decide.go
```diff
@@ -0,0 +1,1157 @@
...
+		StealHigh:                   0.10,
+		StealRecover:                0.06,
+		LimitReserveFraction:        0.10,
+		LimitReserveRecoverFraction: 0.05,
+	}
+}
+
+// throttleWindow is the sliding window length over which the throttle ratio
+// is computed.
+const throttleWindow = 60 * time.Second
```
Aaron99B, line 101: I think this variable and also the other should be self-explanatory. Meaning the naming should be throttleRatioSlidingWindow. Then we could remove the comment

umh-core/pkg/cpuhealth/decide.go
```diff
@@ -0,0 +1,1157 @@
...
+// cpu.max = "max <period>", and dividing by it would poison the fraction
+// with +Inf/NaN. When Quota is nil, Decide falls back to a positive
+// CgroupCores, else treats the sample as uncapped (healthy, no fraction).
+//
+// CgroupCores is the legacy non-pointer quota representation. The production
+// sampler sets Quota, not CgroupCores; the fallback is kept for callers and
+// tests that still populate it.
+type Sample struct {
+	Timestamp   time.Time
+	Quota       *float64
```
Aaron99B, line 159: What quota?

umh-core/pkg/cpuhealth/decide.go
```diff
@@ -0,0 +1,1157 @@
...
+	// even though PSI is known to exist on the host. When PsiReadable is
+	// false, Decide holds the pressure latch instead of clearing it on
+	// PressureAvg60=0, mirroring the hold-on-missing discipline on the
+	// hostBusy and throttle rings.
+	PsiReadable bool
+	// HostBusyCoresAvailable is the readability flag for /proc/stat's
+	// host-busy signal: it distinguishes "read succeeded, host idle" from
+	// "/proc/stat unreadable" when HostBusyCores is 0. Set true by
+	// readProcStat only on a successful read and parse.
+	HostBusyCoresAvailable bool
```
Aaron99B, line 219: Here also an example for renaming. Why not call it Readable instead of Available?

umh-core/pkg/cpuhealth/decide.go
```diff
@@ -0,0 +1,1157 @@
...
+	ts          time.Time
+	nrPeriods   int64
+	nrThrottled int64
+}
+
+// WindowState holds the per-signal 60s sample ring and the per-signal
+// flip-latch that Decide mutates in place. The sliding 60s window is the
+// debounce; the asymmetric (Schmitt) recover band is the only extra mechanism.
+type WindowState struct {
+	throttleRing               []throttlePoint
```
Aaron99B, line 234: I guess these rings here are always ring buffers. Let's also call them like this. Further, why do we use an array for them and not a dedicated RingBuffer datatype?

umh-core/pkg/cpuhealth/decide.go
```diff
@@ -0,0 +1,1157 @@
...
+	stealRing                  []stealPoint
+	usageRing                  []usagePoint
+	hostBusyRing               []hostBusyPoint
+	throttleFired              bool
+	pressureFired              bool
+	stealFired                 bool
+	saturationFired            bool
```
Aaron99B, line 241: Is this really independently computed of the others or could it be a function of the other booleans?

umh-core/pkg/cpuhealth/decide.go
```diff
@@ -0,0 +1,1157 @@
...
+	// when headroom > LimitReserveRecoverFraction × quota. Initial value 0.05.
+	// TODO: calibrate.
+	LimitReserveRecoverFraction float64
+}
+
+// DefaultThresholds returns the canonical thresholds (HighUsageFraction 0.70,
+// SaturationRecover 0.60, ThrottleHigh 0.05, ThrottleRecover 0.03, PressureHigh
+// 0.20, PressureRecover 0.12, StealHigh 0.10, StealRecover 0.06,
+// LimitReserveFraction 0.10, LimitReserveRecoverFraction 0.05).
+func DefaultThresholds() Thresholds {
```
Aaron99B, line 84: I would also move these things in dedicated structs such that one can directly recognise and access the high and low values for each

umh-core/pkg/cpuhealth/decide.go
```diff
@@ -0,0 +1,1157 @@
...
+type throttlePoint struct {
+	ts          time.Time
+	nrPeriods   int64
+	nrThrottled int64
+}
+
+// WindowState holds the per-signal 60s sample ring and the per-signal
+// flip-latch that Decide mutates in place. The sliding 60s window is the
+// debounce; the asymmetric (Schmitt) recover band is the only extra mechanism.
+type WindowState struct {
```
Aaron99B, line 233: Every warning signal here follows the same three-part rule:

* turn the warning **on** when the number crosses a bad threshold
* turn it **off** when it falls back past a good threshold
* **if we couldn't read the number at all, leave the warning exactly as it was**

That third part is the discipline in `doc.go:71-81`, and it's the one that matters.

The rule is typed out by hand seven times: lines 543, 568, 628, 805, 821, 848, 870. Six copies include the third part. The steal copy doesn't — that's the bug I flagged at :586.

That's my case for extracting this in *this* PR rather than the deferred refactor ticket: the duplication isn't a style issue, it's how the bug stayed hidden for 28 commits. Written once, with "did we manage to read it?" as a required argument, forgetting it stops being possible:

```go
// pkg/cpuhealth/latch.go
type gate struct {
    readable  bool // false → leave the warning as it is
    warm      bool // false → not enough history yet, turn it off
    blockFire bool // true → may turn off but not on (steal's min-samples rule)
}
func (l *latch) fireAbove(v, fireMark, clearMark float64, g gate)
func (l *latch) fireBelow(v, fireMark, clearMark float64, g gate)
```

You can't call it without answering the readability question. The zero value leaves the warning alone, which is the safe default. Unexported, so nothing on the wire changes — `Signals` still mirrors plain bools out.

The seven copies have also drifted in ways that are hard to see 300 lines apart: four different implementations of the "leave it alone" clause, three that turn the warning off when history is short versus one that holds instead, and one using `>=` where the others use `>`.

umh-core/pkg/cpuhealth/decide.go
```diff
@@ -0,0 +1,1157 @@
...
+	signals.HostBusyCoresAvailable = sample.HostBusyCoresAvailable
+
+	signals.LimitedVisibility = deadZone
+
+	// Saturation: mode-aware switch over the four sub-latches (see the
+	// Signals docs for each latch's fire and clear marks). The limitMode
+	// case MUST come first: a limit-set sample falling into the no-limit
+	// branches would measure host busyness against the quota-sized capacity
+	// and fire falsely. Below the fire condition the verdict is healthy.
+	switch {
```
Aaron99B, line 795: There are two rulebooks here — limit-mode and no-limit — each with its own pair of saturation flags. When an instance moves between them, the old rulebook's flags must be cleared, or you'd report "this instance is at its CPU limit" for an instance with no limit set.

That clearing is 11 separate statements across the four branches (802-803, 844-846, 867-868, 886-889). It's correct today — I traced all four, no leak. Two things concern me anyway.

**1.** `= false` **means four different things here and they look identical.** "The rulebook changed" (11×), "a better rule superseded this fallback" (846 only), "not enough history yet" (813, 830, 856, 881), and "the condition genuinely went away" — the last being the only real signal.

**2. The most important behavior in this switch is a line that isn't there.** Branch 3 clears `limitSat` and `hostFull` but deliberately leaves `noLimitHostFired` alone, and that silence *is* the hold-on-missing behavior you documented at 836-843. Someone making the four branches look consistent would add `st.noLimitHostFired = false` and silently bring back the flapping this PR fixes. Only `decide_test.go:4779` would object.

One function replaces all 11:

```go
func (st *WindowState) enterMode(m ceilingMode) {
    if st.satMode == m { return }
    st.satMode = m
    // clear all four saturation latches, in one place
}
```

**The mode needs three values (**`limit` **/** `noLimit` **/** `unknown`**), not one per branch.** Branches 2 and 3 are the same rulebook — one just has host stats temporarily missing. Treat them as separate modes and every `/proc/stat` blip clears the flag, which is the flapping again. So *which flag to update* depends on readability (4 cases), but *which flags are meaningful at all* does not (3 cases), and only the second should drive clearing. Three values puts what 836-843 explains in prose into the type instead.

**The live problem this gives a home to:** on a mode change the flags reset but the rings don't, so up to 60s of readings taken against the old ceiling keep feeding the new rules. That's the wart your test documents at `decide_test.go:1885-1891` — setting a CPU limit can produce a false degraded verdict for up to a minute, and degraded blocks bridge creation. There's no single place today that means "the ceiling rule changed," which is why that has nowhere to be fixed.

umh-core/pkg/cpuhealth/decide.go
```diff
@@ -0,0 +1,1157 @@
...
+// ring.
+type stealPoint struct {
+	ts    time.Time
+	steal float64
+}
+
+// usagePoint is one timestamped usage-fraction observation in the WindowState
+// usage ring, used by the no-host-stats saturation latch (NoHostStatsSaturationFired).
+type usagePoint struct {
+	ts       time.Time
```
Aaron99B, line 258: Do we need to store the absolute number and the fraction? can't we just store the absolute number and then compute the fraction when necessary? Then we could also combine this struct with stealPoint and hostBusyPoint into an Event struct

umh-core/pkg/cpuhealth/decide.go
```diff
@@ -0,0 +1,1157 @@
...
+	// saturation cause Value when NoHostStatsSaturationFired is true; 0
+	// otherwise. It deliberately does not use saturationAvg, whose
+	// Quota/CgroupCores denominator is 0 in no-limit mode.
+	NoHostStatsSaturationFraction float64
+	// HostHeadroomCores is the host-scope headroom: LogicalCpus −
+	// hostBusyMean − cpuReserveCores. In no-limit mode it equals
+	// HeadroomCores; in limit mode it is the host-full latch's decision
+	// variable and the saturation cause Value when host-full dominates. NOT
+	// clamped: negative on a full host. Computed whenever LogicalCpus > 0.
+	HostHeadroomCores float64
```
Aaron99B, line 384: Here the variable name is different to the name in the comment. I would remove the comments and make the variable names more self-explanatory

umh-core/pkg/cpuhealth/decide.go
```diff
@@ -0,0 +1,1157 @@
...
+// decides: AttributionHost when the non-container share of host-busy
+// (HostBusyCores60sMean − AvgUsageCores) exceeds the container's own share,
+// else AttributionUnknown. On a /proc/stat outage tick with a held
+// NoLimitHostFired latch the split cannot run, so Attribution is forced to
+// Host for that tick.
+//
+// Quota semantics (nil versus &0 versus positive) are documented on
+// [Sample]. HostContentionFired is reserved and always false: a neighbor
+// filling the box is expressed as saturation plus AttributionHost.
+func Decide(st *WindowState, sample Sample, thresholds Thresholds) (Verdict, Signals) {
```
Aaron99B, line 469: I cannot understand this function with all the complexity inside which make it very likely to introduce some errors. The complexity inside it should be broken down into smaller helper functions to make it more testable and safe

umh-core/pkg/cpuhealth/decide.go
Aaron99B: Further, these ring buffers appear more like sliding windows to me than actual ring buffers

Aaron99B:
I think there might be potential to reduce the complexity in this PR. I let Claude generate a report for this

# CPU health: complexity report

**Summary.** ~2,700 lines of production code carry ~11,000 lines of tests. That ratio is a cross-product, not thoroughness: two ceiling modes × four saturation sub-latches × three readability flags × warm/cold rings. Most of that state space exists because one verdict answers two different questions. Separating them removes the two-rule ceiling, the mode-transition machinery, `Attribution`, and one of the four sub-latches.

## Where the complexity sits

| Unit | Lines |
|---|---|
| `Decide` (single function) | ~570 |
| `Signals` struct + docs (40 fields) | ~150 |
| `sampler.go` | 457 |
| `message.go` | 332 |
| `container_monitor.go` (changed) | ~470 |

## Root cause: health and capacity are merged

Two separable questions share one verdict:

- **Health** — is this instance starved? Container-scoped: throttling, pressure, steal, quota headroom.
- **Capacity** — can this machine take another bridge? Machine-scoped: host headroom.

`hostFullFired` and `noLimitHostFired` are capacity questions expressed as health. What follows from that merge:

- **The two-rule ceiling** — one verdict needing two denominators.
- **The mode-transition machinery** — 11 clearing statements (802-803, 844-846, 867-868, 886-889), a deliberate *missing* line in branch 3 that carries the hold-on-missing behaviour, and a leak invariant that `decide.go:961`, `causeDetails`, `BlockReason`, and `protocolconverter.go:1175` all depend on.
- **`Attribution`** — needed only to explain whether "you" or "the host" caused a merged verdict.
- **The 70%-of-cores fallback latch** — guesses at host capacity when `/proc/stat` is unreadable, presented as a health signal.

Two of the four latches are also arithmetically identical: in no-limit mode `signals.HeadroomCores` (:772) and `hostHeadroomCores` (:779) are the same expression with the same fire and clear marks.

**Behavioural consequence.** On a shared host where other software uses 7.5 of 8 cores, UMH reports itself degraded and blocks bridges even at 0.2 cores of its own usage. Setting a CPU limit doesn't help — `BlockReason` states it: *"the machine is full and this instance's CPU limit cannot help."* No configuration avoids it short of disabling `enableResourceLimitBlocking`, which also disables memory and disk
protection.

**Cost of deferring.** This PR ships `attribution` and four sub-latch booleans; MC #3308 mirrors those types. After both land, changing them is a coordinated two-repo migration. Same for `Cause.value` on the saturation kind, which carries cores or a 0-1 fraction depending on which sub-latch fired (:957-976).

## Simplifications

| # | Change | Removes | Size |
|---|---|---|---|
| 1 | Sample at 1s, not per 100ms tick | rings 600→60 entries; `len>=2` becomes 2s not 200ms; 20s as documented; the "p99 = ring max" window | ~5 lines |
| 2 | Delete the fraction family (`UsageFraction`, `Avg/P95/P99UsageFraction`) | write-only ione sort per tick; 27 test refs | deletion |
| 3 | Extract a `latch` type | 7 hand-written copies of the hysteresis rule; the steal copy's missing readability gate (:586) is the bug | −115/+70 |
| 4 | Collapse `hostFull`/`noLimitHost` into one latch + ceiling label | 11 mode-clears → 1 t |
| 5 | Cause copy as a data table | text written 3× (`causeDetails`, `BlockReason`, docs) and already diverged on the no-host-stats remediation | moderate |
| 6 | Drop the printed arithmetic in the healthy message | ~70 lines of rounding, re-rounding, negative-zero and `+Inf` guards existing so one sentence adds up | −50 lines |
| 7 | Split health from capacity; `IsResourceLimited` consumes capacity directly | the two-rule ceiling, mode transitions, `Attribution`, the fallback latch, and the shared-host blocking behaviour | large — but a net deletion |                                                                                                                                                                                                   Items 1–3 are safe and small. 4–6 fit ENG-5384. Item 7 is a net reduction in this PR's diff, the two host latches and their wire fields rather than adding anything.

## Not recommended for change
                                                                                                                                                                                                                 The hysteresis bands, the readability-flag discipline, `Sample.Quota`'s nil/zero/positive seming, and the wire's `{value, threshold, fired, applies}` per cause. Each has a documentedfailure mode behind it, and `applies` distinguishing "measured and fine" from "unmeasurable h
                                                                                                                                                                                                                Post as a top-level PR comment. The two facts most likely to be contested are the :772/:779 et-full belongs in a health verdict — the counter to the latter is that IsResourceLimitedalready exists as a separate admission-control path, so #7 makes an existing boundary honest
                                                                                                                                                                                                                
                                                                                                                                                                                                                What do you say about it?

JeremyTheocharis:
Superseded by the decomposed stack: #2679 (shared judging library), #2686 (cpuhealth on that library), #2696 (CPU monitor worker), #2697 (verdict seam). Closing in favour of those.

PR #2616, opened 2026-06-26
[ENG-5258] Standalone Deployment return UUID

rpaland:
## What
After deploying a standalone flow the MC can't redirect correctly to it's edit page, because the UUID of the deployed bridge did not correctly return from umh-core.

There is also the ManagementConsole [PR 3281](https://github.com/united-manufacturing-hub/ManagementConsole/pull/3281) which will hold the Changelog entries and screenshots.

### DIFF
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Deploying a stand-alone flow now returns its component ID in the action reply, so the Management Console's "Continue Editing" button opens the flow's editing view instead of only closing the dialog.
+
 ## [0.44.26]
 
 ### Improvements
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -2,10 +2,6 @@
 
 ## Unreleased
 
-### Fixes
-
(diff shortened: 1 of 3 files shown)


umh-core/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Deploying a stand-alone flow now returns its component ID in the action reply, so the Management Console's "Continue Editing" button opens the flow's editing view instead of only closing the dialog.
```
JeremyTheocharis: `Deploying a stand-alone flow now returns its component ID in the action reply,` is not user facing

JeremyTheocharis (changes requested):

umh-core/CHANGELOG.md
JeremyTheocharis: also does this require a MC change as well?

rpaland: Yes screenshots and user facing message was thought to be included in the MC PR, while this was meant as a 'technical fix', but then I'll remove the changelog entry for umh-core entirely and leave it to MC

JeremyTheocharis (approved):

PR #2617, opened 2026-06-26
[ENG-5031] Timeseries-boolean data model support 

rpaland:
## Problem
UMH ships only timeseries-number and timeseries-string as default payload shapes - there is no boolean equivalent. Boolean process values (machine running, alarm active, valve open/closed) are core to manufacturing, so users who reference timeseries-boolean in a data model hit a hard deploy-time validation rejection (referenced payload shape 'timeseries-boolean' does not exist) and must fake on/off state as a 0/1 number, losing type fidelity. 

## What
Timeseries data models should support boolean.


JeremyTheocharis (changes requested):
Can you update the documentation as well?

JeremyTheocharis (approved):

PR #2619, opened 2026-06-28
feat(generator): emit per-side benthos protocol ids on the protocol-converter status row (ENG-5230)

JeremyTheocharis:
## Problem

The bridges-table throughput cell (shipped in ENG-4976 / MC#3272) cannot tell a "not available" side (a protocol that cannot have a read or write flow — e.g. Modbus has no write) from a "not configured" side. The cell already reads `bridge?.inputType` / `bridge?.outputType` and computes a disabled-reason, but the lock stays dormant because umh-core hardcodes `Dfc.Bridge = nil` for protocol-converters, so `bridge?.inputType` is `undefined` and the disabled-reason is never reached.

ENG-5230 puts the protocol id on the wire. The cell-side gate fix (swap `{#if flowHealth}` → `{#if configured}`) is the consuming frontend PR, [ManagementConsole#3285](https://github.com/united-manufacturing-hub/ManagementConsole/pull/3285) (ENG-5249). The two are order-independent: ENG-5249's cell falls back to per-side flow health when this signal is absent, so neither PR blocks the other.

## What we're shipping

umh-core populates `Dfc.Bridge` for protocol-converters (previously `nil`) with the per-side benthos plugin names:

- `InputType` = the read-side plugin name, via a new pure helper `BenthosPluginID` (`pkg/config/dataflowcomponentserviceconfig/plugin_id.go`) that recovers the top-level benthos plugin key of `BenthosConfig.Input`, descending one level through the `"input"` wrapper and skipping `label`/`processors` metadata keys.
- `OutputType` = `DataflowComponentWriteConfigInput.Destination.Protocol` (direct string field).
- `DataContract` = `""` (protocol-converters have no data contract).
- Gated: `Bridge` is `nil` when both sides are empty, so uninitialized PCs keep the prior wire behavior (`DfcBridgeInfo` fields lack `omitempty`; a non-nil empty `Bridge` would serialize as `bridge:{""}`).
- Spec-mandated debug-log: when `BenthosPluginID` returns `""` for a non-empty read input (compound/malformed), and when a configured write side has no `Destination.Protocol` (Code-without-Protocol) — the only diagnostic for an operator chasing a bridge that renders as not-configured despite a configured side.

No new wire fields: the existing `DfcBridgeInfo` struct is reused. No FSM behavior change: `isInitialized`, `AddToManager`, health, and metrics are unchanged; `buildStreamProcessorAsDfc` and the data-bridge row path are untouched.

## Why this reverses the "do not un-nil `Dfc.Bridge`" instruction

The earlier instruction not to populate `Dfc.Bridge` for protocol-converters assumed `DfcBridgeInfo.InputType`/`OutputType` are data-contract types. The frontend schema says otherwise: `status-message-dataflowcomponent.schema.d.ts:71-79` calls them "Benthos name of the input/output type", and `read-write-mapping.json` keys on benthos plugin names (`opcua`, `mqtt`, `modbus`, …). `BenthosPluginID` returns a benthos map key, which is what those fields already mean — so this populates existing fields, it does not repurpose them. No new wire fields, no contract break, no frontend poller change: `coreDfc.bridge` already flows onto `DfcRowData.bridge` via the existing `...dfc` spread. The parent `Bridge` field godoc now notes it also serves protocol-converters.

## Verification

- **Unit:** `BenthosPluginID` resolves `{"label":"x","mqtt":{...}}` → `mqtt`, descends the `input` wrapper, and returns `""` for zero or >1 plugin keys. The generator emits `Bridge: nil` when both sides are empty and a populated `Bridge` otherwise; the two debug logs fire on an empty-resolved read input and a Code-only write. See `plugin_id_test.go` and `protocolconverter_test.go`.
- **Live wire:** on a real container→prod→frontend rig, the frontend `coreStatusMessages` store showed `bridge: {dataContract:"", inputType:"generate", outputType:""}` for a labeled read input (label correctly skipped), and `bridge` absent for protocol-converters on an old-umh-core instance (backward-compat). The cell↔overview "configured" divergence is reproduced and fixed — see ManagementConsole#3285 for the before/after.

## Scope / not shipping

- No DFC lifecycle change. `AddToManager` still creates both DFCs; the unconfigured side stays present-but-stopped. This PR only *reports* configured-ness; it does not make a side genuinely uninitialized (ENG-5251).
- No steady-state cell render change. In steady state `flowHealth` is truthy and the value-branch renders, unchanged. The 4-state configured render is ENG-5249; the lock-icon visual is ENG-5250.
- No compound-input handling. `BenthosPluginID` returns `""` for >1 top-level keys (invalid benthos config; `broker`/`switch` are single-key plugins wrapping an `inputs:` array, not >1 top-level keys).
- `isInitialized` untouched (stays read-biased; write-only symmetry is ENG-5251).

## CHANGELOG

`skip-changelog-guard` — this PR is umh-core-only and has no user-visible effect alone (the visible effect is ENG-5249's frontend, which consumes the signal). The changelog entry rides with ENG-5249.

## Pre-existing bugs noticed (not this PR, separate tickets TBD)

1. `determineProtocol` (`pkg/communicator/actions/dataflow-component-parser.go:411` + `get-protocolconverter.go:182`) takes the first map key without skipping `label` and without deterministic ordering → overview "Protocol: label" for configs with sibling metadata keys. Second instance of the cell↔overview protocol-derivation inconsistency ENG-5249 unifies.
2. (hypothesis, untested) MC-UI write path emits `{outputs}` where umh-core decodes `{destination}` → `Destination.Protocol` may be empty for UI-created write flows → `bridge.outputType:""`. Scenarios A3/A4 on the live rig would confirm.

## Coordination

Do not merge blind: the plan is to validate the consuming frontend PR [ManagementConsole#3285](https://github.com/united-manufacturing-hub/ManagementConsole/pull/3285) against the live two-version rig (old umh-core + new umh-core + new frontend) before merging either. That validation passed — the before/after + cell↔overview proof lives in #3285's description. This PR can sit as a draft until reviewed.

---

### DIFF
--- a/umh-core/pkg/communicator/pkg/generator/protocolconverter.go
+++ b/umh-core/pkg/communicator/pkg/generator/protocolconverter.go
@@ -167,6 +167,17 @@ func buildProtocolConverterAsDfc(
 		writeDesiredState,
 	)
 
+	inputType := dataflowcomponentserviceconfig.BenthosPluginID(input)
+	outputType := observed.ObservedProtocolConverterSpecConfig.Config.DataflowComponentWriteServiceConfig.Destination.Protocol
+
+	var bridge *models.DfcBridgeInfo
+	if inputType != "" || outputType != "" {
+		bridge = &models.DfcBridgeInfo{
+			InputType:  inputType,
+			OutputType: outputType,
+		}
+	}
+
 	dfc := models.Dfc{
 		Type:        models.DfcTypeProtocolConverter,
 		UUID:        uuid.String(),
@@ -181,9 +192,8 @@ func buildProtocolConverterAsDfc(
 		ReadFlowHealth:  readFlowHealth,
 		WriteFlowHealth: writeFlowHealth,
 		// Metrics are added below
-		Metrics: nil,
-		// Bridge info is not applicable for protocol converters
-		Bridge:        nil,
+		Metrics:       nil,
+		Bridge:        bridge,
 		IsInitialized: isInitialized,
 	}
 
--- a/umh-core/pkg/communicator/pkg/generator/protocolconverter.go
+++ b/umh-core/pkg/communicator/pkg/generator/protocolconverter.go
@@ -170,11 +170,9 @@ func buildProtocolConverterAsDfc(
 	inputType := dataflowcomponentserviceconfig.BenthosPluginID(input)
 	outputType := observed.ObservedProtocolConverterSpecConfig.Config.DataflowComponentWriteServiceConfig.Destination.Protocol
 
-	// Bridge carries the per-side benthos plugin names so the frontend (ENG-5249)
-	// can render per-side "configured" state. DfcBridgeInfo fields deliberately
-	// lack omitempty: an unconfigured side must serialize as "" (present) so the
-	// frontend's `bridge.outputType != ""` proxy sees "not configured" rather than
-	// "absent". Do not add omitempty to DfcBridgeInfo fields.
+	// No omitempty on DfcBridgeInfo fields: an unconfigured side must serialize as
+	// "" (present). The frontend reads bridge.outputType == "" as "not configured";
+	// omitting it would read as "absent" and fall back to the wrong render.
 	var bridge *models.DfcBridgeInfo
 	if inputType != "" || outputType != "" {
 		bridge = &models.DfcBridgeInfo{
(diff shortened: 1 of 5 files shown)


FalconTube (approved):
good to go!

PR #2620, opened 2026-06-28
feat(fsmv2): benthos_monitor worker scrapes in-process (USE_FSMV2_BENTHOS_MONITOR, default off)

JeremyTheocharis:
# PR2 — benthos_monitor FSMv2 worker (in-process scrape, behind a default-off flag)

**Target:** `eng-5113-scenariov2` (the migration-API stack base; not staging).
**Label:** `skip-changelog-guard` (inert until `USE_FSMV2_BENTHOS_MONITOR` flips; no user-visible change yet).

## Problem
Each benthos instance carries a second S6 service that scrapes its metrics by forking ~14 helper processes/sec (`curl|gzip|xxd`) and round-tripping a 64KB log that umh-core re-parses. Measured on a `generate` data flow: ~80% of that bridge's added CPU is this monitoring (~12–16 mC/bridge, 58% kernel). It scales per bridge and caps bridges-per-core.

## Shipping
A drop-in FSMv2 `benthos_monitor` worker that scrapes+parses in-process (no subprocess fork tree), behind `USE_FSMV2_BENTHOS_MONITOR` (default OFF, instance-level). Two pieces:

1. **`pkg/service/benthosmetrics`** (new package) — the metrics parser (`ParseMetricsFromBytes`/`TailInt`/`extractLabel`), its value types (`Metrics`, `HealthCheck`, `BenthosMetrics`, …), the throughput windowing (`BenthosMetricsState`), and a `Scan{Metrics, HealthCheck, MetricsAvailable}` type, extracted verbatim from `pkg/service/benthos_monitor` (which keeps type aliases so all 12 existing importers compile unchanged). Plus `Observe(ctx, hc, port)` — the in-process scrape of `/ping`,`/ready`,`/version`,`/metrics`.
2. **`pkg/fsmv2/workers/benthos_monitor/`** (new worker) — on the helloworld template, registers as `"benthos_monitor"`, `CollectObservedState` calls `benthosmetrics.Observe` and publishes a `BenthosMonitorStatus{Scan, Stopped}`. State machine: `StoppedState` (initial; stays on `State=="stopped"` or `IsDisabled`, →`Starting`→`Running`), `RunningState` (→`Stopped` on shutdown), no `Degraded` (benthos health lives in the published `Scan.IsLive`).

## Scope

**In:**
- The `benthosmetrics` package (extracted, no behavior change to the parser/windowing).
- `Observe` with the full error contract: a non-ctx failure on any endpoint (transport, non-200, body-read, parse) **folds** into the Scan (the field stays zero/false, `MetricsAvailable=false`) with a nil error, and the other endpoints are still scraped independently. Only ctx cancellation propagates as a non-nil error. A down benthos → all-false Scan, nil error (an observed state, not an aborted observation). The collector's 2.2s `collectCtx` bounds the COS (no inner Observe deadline).
- The worker (COS, states, `State=="stopped"` scrape-skip with a `Stopped` field distinguishing a paused bridge from a down one).
- A supervisor-driven scenario test (legs: create→Running, benthos-down→Fresh+`IsLive=false` (the J2 foundation proof), stopped→Fresh+`Stopped=true`, recovery, port-change, idempotent Upsert) via `fsmv2client.Upsert`/`Get`/`Delete`.
- Unit tests: happy path, every /metrics failure mode, benthos-down, ctx-cancel (incl. mid-scrape), /ready+ /version independence + fold, slow-but-alive, alloc-budget bench.

**NOT in (deferred / PR3):**
- **PR3 wires the worker** via `fsmv2client` (process-scoped `Set`/`Get`) into the benthos service's FF-on read path (`GetHealthCheckAndMetrics`) + owns the per-instance throughput window. PR2's worker is inert until PR3 + the FF.
- **The full `fsmv2client.GetFresh` maxAge-classification harness** is PR3's layer (PR1 owns `GetFresh`); PR2 proves the `CollectedAt`-advancement foundation via `fsmv2client.Get`.
- **`State` vocabulary translation (`mapFrom`)** is PR3's (the monitor's `OperationalStateStopped` is `"benthos_monitoring_stopped"`, not `"stopped"`; PR3 translates to the worker's FSMv2 `"running"`/`"stopped"`). The worker speaks FSMv2 vocabulary; no v1 `benthos_monitor_fsm` import (M2-safe).

## Known limitations (follow-ups, not blockers)
- **Parser all-or-nothing + silent parse fold:** `ParseMetricsFromBytes` returns `Metrics{}` on the first malformed line (pre-existing, verbatim-moved); `Observe` folds the parse error to `MetricsAvailable=false` with no diagnostic. The operator's actionable state is carried by `IsLive`+`MetricsAvailable` (a parse failure reads "benthos up, /metrics broken, go investigate /metrics"); the parse-error *reason* is a deferred FYI (a future parser PR that makes `ParseMetricsFromBytes` `continue`-on-error, or gives `Observe` a logger — both touch the old FF-off caller `benthos_monitor.ProcessMetricsData`).
- **`TailInt` integer-only:** silently truncates plain decimal floats (`1.5`→1). Pre-existing; benthos counters are integer message counts today. Documented as the integer-only contract; the `ParseFloat` routing is a follow-up parser PR.
- **`parser_test.go`:** the moved package has no dedicated parser unit test file; the 30+ `ParseMetricsFromBytes` error branches are covered indirectly via `observe_test.go` (happy path + one parse-failure path). A focused table-driven `parser_test.go` is a follow-up.
- **`cse/storage/triangular.go` `DeriveWorkerType` camelCase landmine:** `BenthosMonitorStatus`→`"benthosmonitor"` ≠ the registered `"benthos_monitor"`. PR2 bypasses it (explicit-string `register.Worker`); PR3 must NOT use `SaveObservedTyped`/`SaveDesiredTyped` for `benthos_monitor` or it writes the wrong collection. (Pre-existing framework issue; flagged for PR3.)

## Verification
`go build ./...`, `go test ./pkg/service/benthosmetrics/... ./pkg/service/benthos_monitor/... ./pkg/fsm/benthos_monitor/... ./pkg/fsmv2/workers/benthos_monitor/... ./pkg/fsmv2/workers/benthos_monitor/state/... -count=1`, `go test ./pkg/fsmv2/ -run TestFsmv2 -ginkgo.focus "Architecture Validation"`, `go test ./pkg/fsmv2/integration/ -ginkgo.focus "benthos_monitor worker end-to-end"`, `golangci-lint run`, `make betteralign-fix`, `gofmt -l` — all green. The 3 `perfsprint` warnings on the state files match the helloworld/transport convention.

## Phase 3 fixes (2026-07-01, from PR3's self-review/da-sweep on the stacked diff)
Three follow-up commits addressing findings surfaced by PR3's Phase-3 review of the stacked diff (PR2's files):
- `9b5ddbf63` — `Observe` now logs `/metrics` regressions (non-200, body-read, parse) distinctly via `deps.FSMLogger.SentryWarn` (fields: port, status_code / error / body_len+snippet). The transport-unreachable fold stays silent (observed-down, not a regression). Return contract unchanged (still `scan, nil` — the worker stays non-degraded). This partially addresses the "Observe silent-failure" known limitation above (the bigger distinct-degraded-signal / partial-parse fallback is still deferred).
- `dc56a7e70` — `worker.go` `CollectObservedState` compares `cfg.GetState() == config.DesiredStateStopped` (typed constant) instead of the bare string `"stopped"`, matching the sibling state files. A typo is now a compile error.
- `6969b554f` — `parser.go` godoc: removed markdown `**bold**` (Go godoc doesn't render it) + first-person "We read" → "Reads".

All gates green after each (build + benthosmetrics/worker/benthos_monitor suites + 171s integration + golangci-lint 0 + gofmt). Local only — not yet pushed.

## Measured (what this worker enables)
With PR3 wiring this in-process scrape into the read path (FF on), a FF-off-vs-FF-on benchmark (N ∈ {0,5,10} `generate` bridges, per-bridge marginal CPU, slope method) measures **−72% per-bridge monitoring CPU** (19.2 → 5.3 mC/bridge; ~0.70 core at 50 bridges) and the marginal system-time fraction dropping 38% → 10.5% — i.e. the fork/exec kernel cost this worker replaces with an in-process `http.Get`+parse. Full result + method in PR3 (#2623). On the side, the per-bridge process-spawn rate drops ~14/s → ~0.

_arm64 Docker VM measurement; % and mechanism are the durable results, absolute mC is VM-relative._


umh-core/pkg/fsmv2/workers/benthos_monitor/worker.go
```diff
@@ -0,0 +1,131 @@
...
+// the benthos process is intentionally stopped, so there is nothing to scrape,
+// and the admin-pause status is carried by desired State and framework
+// lifecycle rather than by the Scan.
+//
+// Uses fsmv2.NewObservation which signals the supervisor's collector to perform
+// post-COS wrapping (CollectedAt, framework metrics, action history). A
+// canceled context is propagated as context.Canceled alongside a nil
+// observation; a down benthos is an observed state and yields a nil error.
+func (w *BenthosMonitorWorker) CollectObservedState(ctx context.Context, desired fsmv2.DesiredState) (fsmv2.ObservedState, error) {
+	select {
```
JeremyTheocharis, line 101: remove

PR #2622, opened 2026-06-29
Fsmv2 simplification and fsmv2 nmap worker

FalconTube:
What was built and why

  Context: the task

  Implement a TCP port scanner (nmap worker) inside umh-core that the connection service uses to detect
  if a remote PLC/device is reachable. It needed to be a drop-in replacement for the existing S6-based
  NmapManager.

  ---
  Phase 1: First shot (a89a51141)

  Built nmap as a full fsmv2 worker with hand-written state machine:
  - 5 port states: degraded → open/closed/filtered, stopping
  - Separate config.go, dependencies.go, worker.go, action/scan.go, state/*.go — 7 files total
  - One Fsmv2NmapInstance adapter per target (150 lines of fsmv1 shim)
  - One Fsmv2NmapManager (280 lines) per connection service instance
  - Each instance had its own supervisor → goroutine-per-target

  Problem discovered: The fsmv2 framework had friction:
  - No reusable fsmv1↔fsmv2 bridge
  - Lots of boilerplate per worker (state/action/deps files mandatory)
  - Framework used WorkerSnapshot.ParentMappedState (removed by eng-5113-scenariov2)
  - Needed register.SetDeps global for dynamic child management → multiple nmap managers would clobber
  each other

  ---
  Phase 2: Framework update (eng-5113-scenariov2 rebase)

  The eng-5113-scenariov2 branch introduced:
  - ApplicationWorker: a supervisor whose children are declared in YAML (children: [...] in UserSpec)
  - dynamicchildren + fsmv2client: registry-based dynamic child management
  - UpdateUserSpec on Supervisor: push new config → supervisor reconciles on next tick

  Key insight: dynamicchildren (global registry) still caused the multi-manager clobber problem. Used
  UpdateUserSpec + YAML children instead — one ApplicationSupervisor manages all nmap children, config
  changes push YAML spec updates.

  ---
  Phase 3: Pivot to new pattern

  Rewrote the nmap implementation using the ApplicationSupervisor + UpdateUserSpec approach:

  Fsmv2NmapManager: single supervisor, all targets as YAML children. Reconcile() diffs desired config →
  calls buildChildrenSpec() → sup.UpdateUserSpec(yaml). Children reconcile inside the supervisor on the
  next tick.

  Fsmv2NmapInstance: reads observation from shared store via StateReader.LoadObservedTyped("nmap",
  childID, &obs). All fsmv1 lifecycle methods are no-ops.

  ---
  Phase 4: Simplifications (simple, adapter packages)

  Problem 1: 7 files for a worker that just runs a function on a timer.

  Solution — pkg/fsmv2/simple:
  simple.Register[NmapConfig, NmapStatus]("nmap", 1*time.Second, scan)
  One call replaces config.go + dependencies.go + worker.go + action/scan.go + 5 state files. Framework
  handles stopped→running lifecycle, interval tracking, and result storage. The entire nmap worker is now
  nmap/nmap.go: config types + scan() function + init().

  Problem 2: every fsmv2→fsmv1 integration requires 150 lines of identical no-op boilerplate.

  Solution — pkg/fsmv2/adapter:
  - AdaptedInstance[TStatus, TDomainConfig]: generic fsmv1 FSMInstance that reads from StateReader. Only
  non-trivial methods are GetCurrentFSMState and GetLastObservedState, driven by caller-supplied
  MapState/MapObservedState callbacks.
  - WorkerManager[TDomainConfig]: generic reconcile loop (add/remove/update instances, build YAML spec,
  call UpdateSpec). Takes a WorkerManagerSpec with domain-specific callbacks. Implements the full
  NmapManagerProvider interface generically.

  fsmv2nmap shrank to: 1 factory function + 2 mapper closures (instance.go) + 1 supervisor setup + 1 YAML
  builder (manager.go).

  ---
  Phase 5: Cleanup

  - Removed suture/nmap (previous attempt using goroutine-per-target with github.com/thejerf/suture)
  - Removed vendor: thejerf/suture/v4
  - Removed toggle case "suture": from both connection.go and loop.go

  ---
  Net result

```md  ┌─────────────────────────┬───────────────┬────────────────────────┬───────────────────────────────┐
  │         Metric          │    Before     │  Before (fsmv2 first   │             After             │
  │                         │   (suture)    │         shot)          │                               │
  ├─────────────────────────┼───────────────┼────────────────────────┼───────────────────────────────┤
  │ Worker files            │ 3             │ 7                      │ 1                             │
  ├─────────────────────────┼───────────────┼────────────────────────┼───────────────────────────────┤
  │ Manager lines           │ ~200          │ 280                    │ ~70                           │
  ├─────────────────────────┼───────────────┼────────────────────────┼───────────────────────────────┤
  │ Instance lines          │ ~130          │ 150                    │ ~50 (mappers only)            │
  ├─────────────────────────┼───────────────┼────────────────────────┼───────────────────────────────┤
  │ Reusable framework      │ none          │ none                   │ simple + adapter packages     │
  │ additions               │               │                        │                               │
  ├─────────────────────────┼───────────────┼────────────────────────┼───────────────────────────────┤
  │ Goroutines per target   │ 1 (suture)    │ 1 (per-supervisor)     │ 0 (shared                     │
  │                         │               │                        │ ApplicationSupervisor)        │
  └─────────────────────────┴───────────────┴────────────────────────┴───────────────────────────────┘
```
  The simple and adapter packages are general-purpose: any future fsmv2 interval worker needs 1 file; any
  future fsmv2→fsmv1 bridge needs 2 callbacks instead of ~150 lines of boilerplate.

### DIFF
--- a/umh-core/pkg/fsmv2/adapter/instance.go
+++ b/umh-core/pkg/fsmv2/adapter/instance.go
@@ -1,0 +1,155 @@
+// Copyright 2025 UMH Systems GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package adapter bridges fsmv2 workers to the fsmv1 FSMInstance / FSMManager
+// interfaces so that fsmv2-backed components can be used as drop-in
+// replacements inside the existing control loop without writing boilerplate
+// per worker type.
+package adapter
+
+import (
+	"context"
+	"time"
+
+	publicfsm "github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsm"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsmv2"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsmv2/deps"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/service/filesystem"
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/serviceregistry"
+)
+
+// AdaptedInstance wraps a fsmv2 child worker observed via a shared StateReader
+// and satisfies publicfsm.FSMInstance.
+//
+// All lifecycle methods (CreateInstance, RemoveInstance, etc.) are no-ops
+// because fsmv2 manages its own lifecycle through the ApplicationSupervisor.
+// Callers supply MapState and MapObservedState to translate fsmv2 observations
+// into the legacy state vocabulary the rest of the system expects.
+//
+// TStatus is the worker's status type (e.g. simple.Status[NmapStatus]).
+// TDomainConfig is the domain config type that flows through the fsmv1 control
+// loop (e.g. config.NmapConfig).
+type AdaptedInstance[TStatus any, TDomainConfig any] struct {
+	sr               deps.StateReader
+	mapState         func(fsmv2.Observation[TStatus], TDomainConfig) string
+	mapObservedState func(TDomainConfig, fsmv2.Observation[TStatus]) publicfsm.ObservedState
+	domainConfig     TDomainConfig
+	workerType       string
+	childID          string
+	desiredState     string
+	minRequiredTime  time.Duration
+}
+
+// NewAdaptedInstance creates an AdaptedInstance. workerType and childID must
+// match the values used when building the ApplicationSupervisor's children spec.
+func NewAdaptedInstance[TStatus any, TDomainConfig any](
+	sr deps.StateReader,
+	workerType, childID string,
+	cfg TDomainConfig,
+	desiredState string,
+	minTime time.Duration,
+	mapState func(fsmv2.Observation[TStatus], TDomainConfig) string,
+	mapObs func(TDomainConfig, fsmv2.Observation[TStatus]) publicfsm.ObservedState,
+) *AdaptedInstance[TStatus, TDomainConfig] {
+	return &AdaptedInstance[TStatus, TDomainConfig]{
+		sr:               sr,
+		workerType:       workerType,
+		childID:          childID,
+		domainConfig:     cfg,
+		desiredState:     desiredState,
+		minRequiredTime:  minTime,
+		mapState:         mapState,
+		mapObservedState: mapObs,
+	}
+}
+
+// UpdateDomainConfig replaces the stored domain config without touching the
+// childID or mappers. Call this when config values change but the identity
+// (name/workerType) stays the same.
+func (i *AdaptedInstance[TStatus, TDomainConfig]) UpdateDomainConfig(cfg TDomainConfig, desiredState string) {
+	i.domainConfig = cfg
+	i.desiredState = desiredState
+}
+
+// --- publicfsm.FSMInstance implementation ---
+
+func (i *AdaptedInstance[TStatus, TDomainConfig]) GetCurrentFSMState() string {
+	var obs fsmv2.Observation[TStatus]
+
+	_ = i.sr.LoadObservedTyped(context.Background(), i.workerType, i.childID, &obs)
+
+	return i.mapState(obs, i.domainConfig)
+}
+
+func (i *AdaptedInstance[TStatus, TDomainConfig]) GetDesiredFSMState() string {
+	return i.desiredState
+}
+
+func (i *AdaptedInstance[TStatus, TDomainConfig]) SetDesiredFSMState(s string) error {
+	i.desiredState = s
+
+	return nil
+}
+
+func (i *AdaptedInstance[TStatus, TDomainConfig]) IsTransientStreakCounterMaxed() bool {
+	return false
+}
+
+func (i *AdaptedInstance[TStatus, TDomainConfig]) Reconcile(_ context.Context, _ publicfsm.SystemSnapshot, _ serviceregistry.Provider) (error, bool) {
+	return nil, false
+}
+
+func (i *AdaptedInstance[TStatus, TDomainConfig]) Remove(_ context.Context) error {
+	return nil
+}
+
+func (i *AdaptedInstance[TStatus, TDomainConfig]) GetLastObservedState() publicfsm.ObservedState {
+	var obs fsmv2.Observation[TStatus]
+
+	_ = i.sr.LoadObservedTyped(context.Background(), i.workerType, i.childID, &obs)
+
+	return i.mapObservedState(i.domainConfig, obs)
+}
+
+func (i *AdaptedInstance[TStatus, TDomainConfig]) GetMinimumRequiredTime() time.Duration {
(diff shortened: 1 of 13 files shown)


umh-core/pkg/fsmv2/simple/simple.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+
+	return d
+}
+
+// CollectObservedState extracts the current config from the desired state,
+// checks whether the interval has elapsed, and returns an observation that the
+// state machine uses to decide whether to dispatch a run action.
+func (w *simpleWorker[TConfig, TStatus]) CollectObservedState(ctx context.Context, desired fsmv2.DesiredState) (fsmv2.ObservedState, error) {
+	select {
```
JeremyTheocharis, line 143: can be removed

umh-core/pkg/fsmv2/simple/simple.go
```diff
@@ -0,0 +1,267 @@
...
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	default:
+	}
+
+	d := w.GetDependencies()
+
+	cfg := fsmv2.ExtractConfig[TConfig](desired)
+	d.setConfig(cfg)
```
JeremyTheocharis, line 152: double check if necessary

umh-core/pkg/fsmv2/simple/simple.go
```diff
@@ -0,0 +1,267 @@
...
+
+	return fsmv2.Transition(&stoppedState[TC, TS]{}, fsmv2.SignalNeedsRemoval, nil,
+		"stopping complete: "+snap.StopReason(), nil)
+}
+
+func (s *stoppingState[TC, TS]) String() string { return "stopping" }
+
+// --- action ---
+
+type runAction[TC any, TS any] struct{}
```
JeremyTheocharis, line 227: can be removed

umh-core/pkg/fsmv2/simple/simple.go
```diff
@@ -0,0 +1,267 @@
...
+	default:
+	}
+
+	d := w.GetDependencies()
+
+	cfg := fsmv2.ExtractConfig[TConfig](desired)
+	d.setConfig(cfg)
+
+	status := Status[TStatus]{
+		Inner:     d.getResult(),
```
JeremyTheocharis, line 155: no need to use dependencies here

umh-core/pkg/fsmv2/simple/simple.go
```diff
@@ -0,0 +1,267 @@
...
+// function on a fixed interval.
+//
+// Callers register a worker type with Register[TConfig, TStatus], supplying
+// a function called once per interval with the current config and a context.
+// The framework manages started/stopped lifecycle, interval tracking, and
+// result storage automatically.
+//
+// Example:
+//
+//	simple.Register[MyConfig, MyResult]("myworker", 5*time.Second,
```
JeremyTheocharis, line 26: maybe extend this with potential degraderd function that gets MyConfig and MyResult and returns only isDegraded, and if it is degraded also a state reason (e.g., with the reason why it is degraded)

umh-core/pkg/fsmv2/simple/simple.go
```diff
@@ -0,0 +1,267 @@
...
+//	    func(ctx context.Context, cfg MyConfig) (MyResult, error) {
```
JeremyTheocharis, line 27: maybe in the future allow giving dependencies to the register function for potential http.client or whatever

umh-core/pkg/fsmv2/adapter/instance.go
```diff
@@ -0,0 +1,155 @@
+// Copyright 2025 UMH Systems GmbH
```
JeremyTheocharis, line 1: I’d structure it as a thin FSMv1 manager that only reconciles config into `Upsert/Delete`, and let the shared FSMv2 app own execution.

Concretely:

* Keep the read bridge from [umh-core/pkg/fsmv2/fsmv2client/fsmv2client.go](<#file-umh-core/pkg/fsmv2/fsmv2client/fsmv2client.go>)
* Drop the per-manager supervisor from [umh-core/pkg/fsmv2nmap/manager.go](<#file-umh-core/pkg/fsmv2nmap/manager.go>)
* Keep a manager-local map of desired configs and adapter instances
* On reconcile:
  * diff desired configs vs current map
  * for new/changed entries: `client.Upsert(ref, spec)`
  * for removed entries: `client.Delete(ref)`
* Each adapter instance reads via `GetFresh` and maps that to FSMv1 state

Pseudo-shape:

```go
type Manager struct {
    instances map[string]fsm.FSMInstance
    configs    map[string]config.NmapConfig
}
func (m *Manager) Reconcile(...) (error, bool) {
    client := fsmv2client.GetClient()
    if client == nil {
        // either no-op, or surface "fsmv2 unavailable"
    }
    desired := snapshot.CurrentConfig.Internal.Nmap
    // remove
    for name, oldCfg := range m.configs {
        if !existsInDesired(name, desired) {
            _ = client.Delete(refFor(oldCfg))
            delete(m.instances, name)
            delete(m.configs, name)
        }
    }
    // add/update
    for _, cfg := range desired {
        ref := refFor(cfg)
        spec := specFor(cfg)
        if !exists || !equal(old, cfg) {
            _ = client.Upsert(ref, spec)
            m.configs[cfg.Name] = cfg
            m.instances[cfg.Name] = newInstance(cfg)
        }
    }
}
```

The instance side would look close to your adapter in [umh-core/pkg/fsmv2/adapter/instance.go](<#file-umh-core/pkg/fsmv2/adapter/instance.go>), except `GetCurrentFSMState` / `GetLastObservedState` should preferably use `fsmv2client.GetFresh(...)` instead of raw `LoadObservedTyped(...)`, because you want to distinguish:

* unregistered
* never observed
* stale
* fresh

That gives you cleaner FSMv1 mapping, e.g.:

* `Unregistered` → starting / absent
* `NeverObserved` → starting
* `Stale` → degraded
* `Fresh` + status → open/closed/filtered

The main pieces you’d need:

* `refFor(cfg)` → build `dynamicchildren.Ref{WorkerType: "nmap", Name: cfg.Name}`
* `specFor(cfg)` → build the same child spec payload currently produced in [umh-core/pkg/fsmv2nmap/manager.go](<#file-umh-core/pkg/fsmv2nmap/manager.go>)
* `newInstance(cfg)` → adapter instance that reads through the global client

So overall:

* PR 2621 architecture = **shared global FSMv2 runtime**
* manager becomes **control-loop compatibility + config diffing only**
* no need for `UpdateUserSpec` in [umh-core/pkg/fsmv2/supervisor/api.go](<#file-umh-core/pkg/fsmv2/supervisor/api.go>)

The tradeoff is that this manager is no longer self-contained: it now depends on startup wiring in [umh-core/cmd/main.go](<#file-umh-core/cmd/main.go>) having published the registry and client.

Possible next steps:

* I can sketch the exact `Reconcile` flow against your current [umh-core/pkg/fsmv2nmap/manager.go](<#file-umh-core/pkg/fsmv2nmap/manager.go>)
* I can show how I’d rewrite the adapter instance to use `GetFresh`
* I can outline the failure modes of the global-client approach vs the local-supervisor approach

umh-core/pkg/fsmv2/workers/nmap/nmap.go
```diff
@@ -0,0 +1,115 @@
...
+	Port uint16 `json:"port,omitempty"`
+	// IsRunning indicates whether the port is currently open.
+	IsRunning bool `json:"isRunning"`
+}
+
+// scan performs a TCP connect scan against cfg.Target:cfg.Port and returns the
+// port state. It returns an error only for context cancellation; network
+// failures are represented as "closed" or "filtered" states.
+func scan(ctx context.Context, cfg NmapConfig) (NmapStatus, error) {
+	select {
```
JeremyTheocharis, line 70: this select statement can be removed

PR #2626, opened 2026-07-01
Surface HTTP status code + upstream body on persistent connection-failure (ENG-5023)

JeremyTheocharis:
**Problem.** `persistent_push_failure` / `persistent_pull_failure` classify the error (`error_type`) but discard the HTTP status code and response body. In ENG-5275, several instances flapped emitting `server_error`, and neither Sentry nor the on-box logs showed the upstream `502` a reverse proxy in the egress path was returning. An operator could see *that* push failed, never *what* the upstream said.

**Shipping.** `status_code` (omitted when 0) and a sanitized `error_detail` (capped at `MaxErrorDetailBytes` = 256 bytes total, including the `HTTP <code> (<type>): ` prefix; clipped details end in `…`) now appear on both the `persistent_push_failure` / `persistent_pull_failure` Sentry event and the push/pull degraded-state reason string, so a proxy-injected 502 can be localized without shell access to the instance.

**Scope.** Transport push/pull only. `ExtractErrorDetails` returns the type, retry-after, status code, and a control-char-stripped, whitespace-collapsed, rune-safe-capped detail from a `*TransportError`; the push and pull actions pass these through a 4-arg `RecordTypedError`, which stores them and emits them conditionally (status_code only when >0). `newTransportError` embeds a body preview capped at the same `MaxErrorDetailBytes` const (replacing the `<200` gate that dropped larger nginx pages); since the cap applies to the whole detail string, the effective body budget is ~231 bytes after the prefix. The degraded-state reason appends `; last: <detail>` on the still-degraded paths (backoff and still-pushing/pulling), not on stop, recovering, or waiting.

A capstone test reproduces the incident shape: a fake transport returning an nginx 502 page, driven past the escalation threshold, asserts both the event (status_code=502 + the nginx body) and the degraded reason (`; last: HTTP 502 (server_error): … nginx …`).

**Not shipping (deferred).** Credential redaction is out of scope and tracked in ENG-5289. `error_detail` is sanitized for control characters and capped, but it is **not** redacted: if a server or proxy echoes the request's `Authorization` header or a credential in its error body, that content can reach Sentry and the degraded-reason logs. This is the accepted consequence of shipping the diagnostic now; ENG-5289 will add a proper secret-scanner. Three hardening findings were deferred as out of scope: allocation on large error bodies (the cap is on output; `strings.Map`/`Fields` allocate a full copy of the unbounded input first), unbounded `io.ReadAll` on upstream error bodies before the cap, and `CheckAndClearOnReset` not clearing the last-error fields on transport reset.

Push and pull mirror each other. Mutation testing was waived (no Go tooling in the repo); coverage is the TDD ladder (each behavior its own RED-first commit) plus the Phase-3 specialist review and adversarial spec-fidelity sweep.


umh-core/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Improvements
+
+- Persistent connection-failure diagnostics now include the HTTP status code and a sanitized snippet of the upstream response on both the Sentry event and the worker's degraded-state log line. A failure injected by an upstream proxy, such as a 502, can now be identified without shell access to the instance
```
JeremyTheocharis: sentry is not user facing here

JeremyTheocharis: should be reworded (if there is e.g., a change to the user logs) or if there is nothing user facing removed altogtehter

umh-core/pkg/fsmv2/workers/transport/http/http.go
```diff
@@ -158,8 +159,17 @@ func newTransportError(statusCode int, body []byte, headers http.Header, baseErr
 	retryAfter := parseRetryAfter(headers)
 
 	msg := fmt.Sprintf("HTTP %d: %s", statusCode, errType.String())
-	if len(body) > 0 && len(body) < 200 {
-		msg = fmt.Sprintf("HTTP %d (%s): %s", statusCode, errType.String(), string(body))
+
+	if len(body) > 0 {
```
JeremyTheocharis, line 163: needs to have a comment explaining what happens here

umh-core/CHANGELOG.md
JeremyTheocharis: Good catch — dropped the Sentry mention. The bullet now covers only the user-facing surface (the on-box log line an operator sees during troubleshooting): "Persistent connection failures now log the HTTP status code and a sanitized snippet of the upstream response, so the cause of a flapping connection (such as a 502 from a reverse proxy) is visible in the worker's logs." Pushed in 2595386e3.

umh-core/pkg/fsmv2/workers/transport/http/http.go
JeremyTheocharis: Added a comment above the loop: "Cap at 256 bytes and back off to a UTF-8 rune boundary so the body preview never splits a multibyte sequence." Pushed in 2595386e3.

umh-core/pkg/fsmv2/workers/transport/http/http.go
```diff
@@ -158,8 +159,19 @@ func newTransportError(statusCode int, body []byte, headers http.Header, baseErr
...
+		end := len(body)
+		if end > 256 {
```
led0nk: Dead cap: `sanitizeErrorDetail` re-caps `prefix+body` to 256 downstream (via `ExtractErrorDetails` -> `te.Error()`), and it's always stricter, so this body-only cap never changes the output. Drop it and keep a single cap in `sanitizeErrorDetail`.

umh-core/pkg/fsmv2/workers/transport/types/types.go
```diff
@@ -290,3 +325,34 @@ func CounterForErrorType(t ErrorType) depspkg.CounterName {
...
+		return r
+	}, s)
+
+	s = strings.Join(strings.Fields(s), " ")
+
+	if len(s) <= 256 {
+		return s
+	}
+
+	end := 256
```
led0nk: Effective body budget is ~231, not 256: the ~25-byte `HTTP <code> (<type>): ` prefix is included in this cap. Hoist 256 to a named const (single source of truth), and append `…` when clipped so operators can tell the detail was truncated. PR description ("first <=256 bytes of body") is off by the prefix.

umh-core/pkg/fsmv2/workers/transport/types/types.go
```diff
@@ -254,12 +261,40 @@ func (e *TransportError) Is(target error) bool {
...
+		return ErrorTypeUnknown, 0, 0, ""
+	}
+
+	var te *TransportError
+	if errors.As(err, &te) {
+		if te == nil {
+			return ErrorTypeUnknown, 0, 0, ""
+		}
+
+		return te.Type, te.RetryAfter, te.StatusCode, sanitizeErrorDetail(te.Error())
```
led0nk, line 299: Blind prefix cut, no HTML/JSON awareness. If the real cause sits past ~256 bytes — Cloudflare/WAF 5xx pages (big `<head>`/CSS before the Ray ID) or verbose JSON envelopes (Spring Boot: timestamp/status/error/path/trace before `message`) — `error_detail` keeps only boilerplate and drops the reason. Given this PR targets a reverse proxy in the egress path, is that acceptable, or should capture prefer the status line / a `message` field?

umh-core/pkg/fsmv2/workers/transport/types/types.go
JeremyTheocharis: Fixed in bf89593ba: both truncation sites now share an exported `types.MaxErrorDetailBytes = 256`, clipped output ends in `…` (total stays ≤ the cap), and the PR description now states the cap covers the whole detail string (~231-byte effective body budget after the prefix).

umh-core/pkg/fsmv2/workers/transport/http/http.go
JeremyTheocharis: The cap here isn't dead — `Message` has a second consumer that bypasses `sanitizeErrorDetail`: the wrapped error returned from the push/pull action reaches the framework's action executor (`supervisor/internal/execution/action_executor.go:240`), which stores `err.Error()` raw into `ActionResult.ErrorMsg` and persists it in action history (`deps/history.go`). Since all three call sites read the body with a plain `io.ReadAll` (no `LimitReader`), dropping this cap would put an unbounded upstream error page verbatim into `Message` and into persisted history on every failure.

The real issue was the duplicated literal — fixed in bf89593ba by hoisting both sites onto the shared `types.MaxErrorDetailBytes` const (per your other comment).

umh-core/pkg/fsmv2/workers/transport/types/types.go
JeremyTheocharis: Accepting the blind prefix cut for v1, for two reasons:

1. Classification runs on the **full** body before truncation — `classifyError(statusCode, body, headers)` sniffs the complete body for Cloudflare-challenge and proxy-block (Zscaler) signatures. So for exactly the reverse-proxy case this PR targets, the operator gets `error_type=cloudflare_challenge`/`proxy_block` plus the status code even when the detail snippet is `<head>` boilerplate. The snippet is best-effort color; status code + type are the primary signal.
2. The shipping goal (localize a proxy-injected 502 without shell access) is met by status code + type alone.

Your Spring-Boot case is real though — a JSON envelope with `message` past byte ~231 keeps only timestamp/path boilerplate. Content-aware extraction (prefer a JSON `message` field / HTML `<title>`) needs parsing heuristics I'd rather not bolt on here; happy to open a follow-up ticket alongside ENG-5289 (redaction), which already carries the deferred detail-quality work. Say the word and I'll create it.

led0nk (approved):

PR #2627, opened 2026-07-01
[ENG-5299] Stopped state on initial deployment failure

rpaland:
## What
Failed bridges now set the desired state to stopped, so they wont automatically try to redeploy and fail over and over again.

#### Read Flow Failure
Bildschirmfoto 2026-07-07 um 09 55 05


FalconTube (commented):
- In title, add "initial deployment" to make intent more clear :)
- Let's briefly discuss in team, if this the behavior we want

umh-core/pkg/communicator/actions/deploy-protocolconverter.go
```diff
@@ -244,10 +244,61 @@ func (a *DeployProtocolConverterAction) Execute() (interface{}, map[string]inter
...
+				}
+			}
+
+			if !found {
+				// Bridge no longer exists (e.g. deleted during the wait): nothing to stop.
+				return nil, nil, err
+			}
+
+			readActive := pcToStop.ProtocolConverterServiceConfig.ReadDFCDesiredState == dataflowcomponent.OperationalStateActive
+			writeActive := pcToStop.ProtocolConverterServiceConfig.WriteDFCDesiredState == dataflowcomponent.OperationalStateActive
```
FalconTube: You need to gate vs `state != StateStopped`.<br>This is because pre-write flow (which includes pre "can be stopped") the `active` state might just be empty `""`, which then indicates active.<br>So checking vs `stopped` is better here.<br><br>Super annoying and niche thing about the flows and states, not your fault/oversight!

umh-core/pkg/communicator/actions/deploy-protocolconverter.go
```diff
@@ -464,3 +521,34 @@ func (a *DeployProtocolConverterAction) waitForComponentToAppear(desiredState st
...
+
+	if !stopRead && !stopWrite {
+		return readActive, writeActive
+	}
+
+	return stopRead, stopWrite
+}
+
+// isDFCFlowRunning reports whether a dataflow component FSM state means the flow came up.
+func isDFCFlowRunning(state string) bool {
```
FalconTube: No need for this helper function here.<br>Where already have `dataflowcomponent.IsRunninState(…)`, which does the same

umh-core/pkg/communicator/actions/deploy-protocolconverter.go
```diff
@@ -464,3 +521,34 @@ func (a *DeployProtocolConverterAction) waitForComponentToAppear(desiredState st
...
+// flowsToStop decides which dataflow components to flip to stopped after a failed
+// deploy. A flow is stopped only when it was meant to run (desired active) and did
+// not reach a healthy state.
+func flowsToStop(snapshot *protocolconverter.ProtocolConverterObservedStateSnapshot, readActive, writeActive bool) (stopRead, stopWrite bool) {
+	if snapshot == nil {
+		return readActive, writeActive
+	}
+
+	info := snapshot.ServiceInfo
+	if readActive && !isDFCFlowRunning(info.DataflowComponentReadFSMState) {
```
FalconTube: `isDFCFlowRunning` includes `degraded` state. I think if we are degraded, then we want to stop it, right?<br><br>If this is correct, then we want another helper function `isDFCHealth`, which only checks for `active` or `idle` state. Then we compare against that.<br><br>I have tested in the product that it works with current implementation, but I still find it weird to follow here 😅

umh-core/pkg/communicator/actions/deploy-protocolconverter.go
```diff
@@ -464,3 +521,34 @@ func (a *DeployProtocolConverterAction) waitForComponentToAppear(desiredState st
...
 	}
 }
+
+// flowsToStop decides which dataflow components to flip to stopped after a failed
+// deploy. A flow is stopped only when it was meant to run (desired active) and did
+// not reach a healthy state.
+func flowsToStop(snapshot *protocolconverter.ProtocolConverterObservedStateSnapshot, readActive, writeActive bool) (stopRead, stopWrite bool) {
+	if snapshot == nil {
+		return readActive, writeActive
+	}
```
FalconTube, line 534: Function is hard to follow, because it re-uses inputs as outputs.<br>It does the correct thing, so just add brief comment here like this

```
// No observed state: can't isolate the failure, so stop every active flow.
if snapshot == nil {
```

umh-core/pkg/communicator/actions/deploy-protocolconverter.go
```diff
@@ -464,3 +521,34 @@ func (a *DeployProtocolConverterAction) waitForComponentToAppear(desiredState st
...
+		stopRead = true
+	}
+
+	if writeActive && !isDFCFlowRunning(info.DataflowComponentWriteFSMState) {
+		stopWrite = true
+	}
+
+	if !stopRead && !stopWrite {
+		return readActive, writeActive
+	}
```
FalconTube, line 548: Also here, make it easier to understand with short comment like

```
// No active flow looks down, yet deploy failed: stop every active flow.
if !stopRead && !stopWrite {
```

FalconTube (changes requested):

FalconTube (approved):
Looks good now.
As discussed via Slack, add a small frontend follow up PR to unify the look in overview page and table view.

PR #2632, opened 2026-07-06
Eng 5305 pr a framework prep

FalconTube:
PR-A in ENG-5305

### DIFF
--- a/umh-core/pkg/fsmv2/observation.go
+++ b/umh-core/pkg/fsmv2/observation.go
@@ -152,6 +152,17 @@ func NewObservation[TStatus any](status TStatus) ObservedState {
 type Observation[TStatus any] struct {
 	// CollectedAt is when this observation was taken.
 	CollectedAt time.Time `json:"collected_at"`
+	// Status is the developer's business data. Flattened to top level via custom MarshalJSON.
+	Status TStatus `json:"-"`
+	// State is the current FSM state name (set by supervisor via SetState).
+	State string `json:"state"`
+	// Reason is the human-readable explanation for the current health verdict,
+	// shown in logs and the frontend (e.g. "poll error: dial timeout"). The JSON
+	// tag is framework-namespaced (health_) so it does not collide with a worker's
+	// own status fields (ENG-5305).
+	Reason string `json:"health_reason"`
+	// LastActionResults contains action history (managed by supervisor).
+	LastActionResults []deps.ActionResult `json:"last_action_results,omitempty"`
 	// ChildrenView is the per-tick snapshot of children state for parent
 	// workers. The value round-trips through CSE storage between the collector
 	// goroutine and the reconciler goroutine, so the field MUST stay JSON-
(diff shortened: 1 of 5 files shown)


umh-core/pkg/fsmv2/observation.go
```diff
@@ -177,6 +182,10 @@ type Observation[TStatus any] struct {
 	ChildrenUnhealthy int `json:"children_unhealthy"`
 	// ShutdownRequested mirrors the desired state's shutdown flag.
 	ShutdownRequested bool `json:"ShutdownRequested"` //nolint:tagliatelle // Match existing API field name
+	// Degraded is the worker's health verdict for this observation: true when the
+	// target is unhealthy or could not be determined. The fsmv2 state and the
+	// fsmv1 adapter read it as the one authoritative verdict.
+	Degraded bool `json:"health_degraded"`
```
JeremyTheocharis, line 188: doesn't this leak simple worker into the normal fsmv2 api? why not simply put it into TStatus? then it remains part of simple worker

umh-core/pkg/fsmv2/observation_interval_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsmv2"
+)
+
+// The per-worker-type observation-interval registry lets a worker declare its
+// collection cadence once (via simple.Spec.Interval) so both the supervisor's
+// collector and the fsmv1 adapter (StaleAfter = 3×Interval) read one source of
+// truth. The getter returns (value, ok); the caller applies its own default so
+// this core package need not import the supervisor's DefaultObservationInterval
+// (ENG-5305, Q4/Q11).
```
JeremyTheocharis: use the self-review skill to fetch remainign engineering artifacts, or alternativelyt just the anti ai slop workflow on this. it will automatically remove "`Q4/Q11).`" and other spec artifacts

JeremyTheocharis (changes requested):

umh-core/pkg/fsmv2/observation.go
JeremyTheocharis: i think we also already talked about it

JeremyTheocharis: ok is already removed in PR B, might be cleaner to remove it directly here?

FalconTube: as discussed, will keep it here, because not very long lived change

umh-core/pkg/fsmv2/observation_interval_test.go
FalconTube: will do and push update

FalconTube: De-slopped it now

JeremyTheocharis (approved):

PR #2636, opened 2026-07-07
feat: reference the shared historian connection in bridges

FalconTube:
Lets a bridge reference the shared historian connection from `config.yaml`, so a bridge that saves to the historian no longer re-enters the TimescaleDB credentials. umh-core side of "Save to Historian" in [ENG-5098](https://linear.app/united-manufacturing-hub/issue/ENG-5098).

## What it does

- Bridge templates reference the connection as `{{ .historian.timescale.host }}`, `{{ .historian.timescale.port }}`, and the other fields, mirroring the `historian.timescale` block in `config.yaml`.
- A bridge re-renders (and re-deploys) when the shared `historian` block changes.
- A bridge that references the historian while none is configured fails with an actionable error naming the missing `historian:` section, not a raw template error.

## Hardening (from review)

- The historian password stays out of logs and the Management Console status reason: it is masked in the render-error snippet. The real value still reaches the on-disk runtime config so Benthos can connect. Complements the get-historian reply redaction in the base PR.
- Only a valid historian is injected. A present-but-invalid one is treated as absent, so a referencing bridge fails loudly instead of rendering empty credentials; a bridge that does not reference it is unaffected.
- The edit verify-render propagates a config-read failure instead of continuing with no historian. Previously that could roll a valid edit back with a misleading "invalid YAML" message.

## Base

Stacked on [ENG-5318](https://linear.app/united-manufacturing-hub/issue/ENG-5318) (`eng-5318-historian-crud-actions-in-umh-core`); rebased onto its `historian.timescale` nesting.

## For ManagementConsole

The generated historian benthos template must emit `{{ .historian.timescale.* }}` to match this shape.

## Tests

Template map contract (key set is now value-independent, locked by `TimescaleTemplateKeys`), secret redaction, present-but-invalid handling, the config-read-failure path, and the edit verify-render wiring. A `BuildRuntimeConfig` test also renders a bridge that references every `{{ .historian.timescale.* }}` field — including the `ssl*` cert paths — with no TLS certs configured, and asserts it resolves: the regression guard for the deploy timeout (re-adding `json:omitempty` makes it fail with the original `map has no entry for key "sslrootcert"`).

## Review follow-ups

- `ToTemplateMap` keeps the `json.Marshal`/`json.Unmarshal` round-trip, so a new connection field (or a future `grafana:` sub-block) reaches templates without editing the function. The optional TLS cert fields (`sslrootcert`/`sslcert`/`sslkey`) drop `json:omitempty` so an unset cert survives the round-trip as an empty key instead of vanishing — otherwise, under the default `sslmode=require`, a bridge referencing `{{ .historian.timescale.sslrootcert }}` trips `RenderTemplate`'s `missingkey=error`, renders an empty benthos config, and times out on deploy. `yaml:omitempty` stays, so `config.yaml` still omits unset certs; the `get-historian` reply now carries the empty cert fields. A `TimescaleTemplateKeys` test locks the exposed key set and fails if `omitempty` is re-added.

### DIFF
--- a/umh-core/pkg/config/historian_config.go
+++ b/umh-core/pkg/config/historian_config.go
@@ -15,6 +15,7 @@
 package config
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 )
@@ -203,4 +204,41 @@ func (t TimescaleConfig) String() string {
 
 	type redacted TimescaleConfig
 	return fmt.Sprintf("%+v", redacted(t))
+}
+
+// ToTemplateMap returns the historian settings as a nested map for template
+// rendering, mirroring the `historian.timescale` shape in config.yaml. Bridge
+// templates reference the connection as `{{ .historian.timescale.host }}`,
+// `{{ .historian.timescale.port }}`, and the other fields. It returns an empty
+// map when no timescale section is present.
+func (h HistorianConfig) ToTemplateMap() map[string]any {
+	if h.Timescale == (TimescaleConfig{}) {
+		return map[string]any{}
+	}
+
+	return toTemplateMap(h.WithDefaults())
+}
+
+// ToTemplateMap returns the timescale connection settings as a flat map keyed by
+// the config.yaml field names, for template rendering. Defaults are applied first
+// so templates always see resolved values.
+func (t TimescaleConfig) ToTemplateMap() map[string]any {
+	return toTemplateMap(t.WithDefaults())
+}
+
+// toTemplateMap serialises a config value to a map keyed by its json field names.
+// A JSON round-trip keeps the exposed keys in step with the struct tags, so a new
+// field reaches templates without editing this function.
+func toTemplateMap(v any) map[string]any {
+	b, err := json.Marshal(v)
+	if err != nil {
+		return map[string]any{}
+	}
+
+	m := map[string]any{}
+	if err := json.Unmarshal(b, &m); err != nil {
+		return map[string]any{}
+	}
+
+	return m
 }
--- a/umh-core/pkg/config/historian_config.go
+++ b/umh-core/pkg/config/historian_config.go
@@ -15,7 +15,6 @@
 package config
 
 import (
-	"encoding/json"
 	"errors"
 	"fmt"
 )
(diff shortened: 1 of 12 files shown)


umh-core/pkg/service/protocolconverter/runtime_config/runtime_config_test.go
```diff
@@ -226,12 +226,76 @@ var _ = Describe("BuildRuntimeConfig", func() {
...
 
-			result, err := runtime_config.BuildRuntimeConfig(testSpec, agentLocation, globalVars, nodeName, pcName)
+			result, err := runtime_config.BuildRuntimeConfig(testSpec, agentLocation, globalVars, nil, nodeName, pcName)
 			Expect(err).NotTo(HaveOccurred())
 			Expect(result.DataflowComponentWriteServiceConfig.BenthosConfig.Input).To(BeEmpty())
 		})
 	})
 
+	Describe("Historian variable injection", func() {
+		It("should expose historian settings as {{ .historian.* }} in templates", func() {
```
Aaron99B: Shouldn't we test for all the historian parameters here?

umh-core/pkg/service/protocolconverter/runtime_config/runtime_config_test.go
```diff
@@ -226,12 +226,76 @@ var _ = Describe("BuildRuntimeConfig", func() {
...
+					ConnectionServiceConfig: connectionserviceconfig.ConnectionServiceConfigTemplate{
+						NmapTemplate: &connectionserviceconfig.NmapConfigTemplate{
+							Target: "{{ .historian.host }}",
+							Port:   "{{ .historian.port }}",
+						},
+					},
+				},
+			}
+
+			historian := &config.HistorianConfig{
```
Aaron99B: Also here I would use all parameters except the ones for verify-full

Aaron99B:
And can you create a linear ticket for this PR and add it as a sub ticket here? https://linear.app/united-manufacturing-hub/issue/ENG-5098/rollout-historian-integration-in-management-console

umh-core/pkg/service/protocolconverter/runtime_config/runtime_config_test.go
FalconTube: This isn't really required. If one of them can be found, then all of them can also be found.
If you test all of them now, you also have to make sure, that they do that in the future :D 
So to me this just seems like more effort for no extra gain

umh-core/pkg/service/protocolconverter/runtime_config/runtime_config_test.go
FalconTube: See comment above

umh-core/pkg/service/protocolconverter/runtime_config/runtime_config_test.go
Aaron99B: Okay

umh-core/pkg/config/historian_config.go
```diff
@@ -204,3 +204,37 @@ func (t TimescaleConfig) String() string {
...
+		"port":        t.Port,
+		"database":    t.Database,
+		"username":    t.Username,
+		"password":    t.Password,
+		"sslmode":     string(t.SSLMode),
+		"sslrootcert": t.SSLRootCert,
+		"sslcert":     t.SSLCert,
+		"sslkey":      t.SSLKey,
+	}
+}
```
FalconTube, line 263: I know this function came from me initially, but better use `json.Marshal` and `json.Unmarshal` here instead of hard-coding the stuff.

Same for the `ToTemplateMap` for the entire historian config above

FalconTube (commented):
Minor edit, then good to go afterwards.
I tested the actual functionality as well, works well!

But needs to be rebased properly, at the moment it also includes the commit from the underlying branch, which is stale

umh-core/pkg/config/historian_config.go
Aaron99B: Done in 8bcb072 — both `ToTemplateMap` methods now use a `json.Marshal`/`json.Unmarshal` round-trip via a small `toTemplateMap` helper, so new fields (and a future `grafana:` sub-block on `HistorianConfig`) get exposed to templates automatically.

One behavioral note worth a look: JSON respects `omitempty`, so the optional SSL cert paths (`sslrootcert`/`sslcert`/`sslkey`) are now omitted from the map when empty, instead of being present as `""`. With `missingkey=error` that means a template referencing `{{ .historian.timescale.sslrootcert }}` on a non-TLS connection would fail the render rather than resolve to empty. No in-repo template references them, so nothing breaks today — but if you'd rather those keys always be present, say the word and I'll keep them. Numbers also decode as `float64` now; they render identically through the template (verified by the existing runtime_config render tests).

FalconTube (commented):
Ok, rebase now properly and good to merge

umh-core/pkg/config/historian_config.go
FalconTube: good to keep as is

FalconTube:
Good to merge now. 
But again, since I opened, you @Aaron99B must approve it :)

Aaron99B (approved):

Aaron99B:
## CodeRabbit autofix applied

Addressed the 🟡 Minor "remove newly added inline comments from the Go test changes" thread — removed the inline explanatory comments from the test files, keeping comments to exported godoc / non-obvious constants per the repo rule. No test logic changed.

**Files modified:**
- `umh-core/pkg/config/historian_config_test.go`
- `umh-core/pkg/config/templating_test.go`
- `umh-core/pkg/communicator/actions/edit-protocolconverter_test.go`

**Commit:** `a26c3629e`

FalconTube:
Re-reviewed and good to go now!

PR #2642, opened 2026-07-13
fix: missing datacontract no longer skips all newer contracts

led0nk:
# Description

## Origin topic

When having a data-model which uses e.g. timeseries-number as datatype for its tag:

- if you then tried to write with e.g. `timeseries-string` into the tag the error-message was very unintuitive
- it printed each tag of the data-contract + datatype
- but you had to guess the real issue here, because it only stated that the `virtual_path` is "wrong"

For reference ENG-5347

## After Fix

I implemented the fix within benthos-umh [schema-validation-error](https://github.com/united-manufacturing-hub/benthos-umh/pull/414), but when i wanted to recreate this i ran into this issue here:

- already had 2 data-contracts pre-existing
- deleted the correpsonding datamodels
- created a new data-contract ("typevalidation")
- added the datamodel ("typevalidation_v1")
- the tag appeared in topic-browser in the correct data-contract
- BUT it got a value of `true` whenever the datamodel was `timeseries-number`
- checked on metadata -> `bypass_reason` was `no schema found`

Conclusion: 

Whenever you had pre-existing datamodels, which were deleted, the whole chain of other datamodels, which should be applied to redpanda were skipped, due to a translation-error.

## Now

This PR introduces that datamodels, which cannot get translated / have missing schemas are getting skipped and do not block anymore the appliance / translation for newer datamodels.

Fixes ENG-5319

### DIFF
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- A misconfigured data contract no longer stops your other data contracts from being set up. Valid contracts are now registered and validated as usual, while the misconfigured one is skipped and reported as a warning
+
 ## [0.44.31]
 
 ### Improvements
@@ -22,16 +26,11 @@
 ## [0.44.29]
 
 ### New Features
-
 - Preview: an instance can now store a connection to an external TimescaleDB or PostgreSQL historian database (host, port, database, login, and TLS mode), which can be created, viewed, updated, and removed. Storing the connection does not yet route any data to it
 
 ### Improvements
 
 - The full bridge and stream-processor configuration is no longer written to the umh-core log at INFO on every internal re-apply (moved to DEBUG), removing a source of log flooding and CPU pressure when a config keeps being re-applied. While a bridge's observed config differs from its desired config, its status in the Management Console now shows a bounded summary of what differs, and the umh-core log carries one warning per bridge per minute, so the cause is visible without enabling debug logging
-
-### Fixes
-
-- A misconfigured data contract no longer stops your other data contracts from being set up. Valid contracts are now registered and validated as usual, while the misconfigured one is skipped and reported as a warning
 
 ## [0.44.28]
 
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -22,11 +22,16 @@
 ## [0.44.29]
 
 ### New Features
+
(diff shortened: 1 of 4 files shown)


umh-core/pkg/service/redpanda/schema_registry.go
```diff
@@ -387,7 +390,17 @@ func (s *SchemaRegistry) Reconcile(ctx context.Context, dataModels []config.Data
...
 
-	return s.ReconcileWithSchemas(ctx, expectedSubjects)
+	s.mu.Lock()
+	s.skipPrefixes = skipPrefixes
+	s.mu.Unlock()
+
+	reconcileErr := s.ReconcileWithSchemas(ctx, expectedSubjects)
+
+	if err != nil {
+		return fmt.Errorf("some data contracts could not be translated to schemas: %w", errors.Join(err, reconcileErr))
```
JeremyTheocharis, line 400: can you check with @FalconTube that this doesn't spam? i can assume this will throw a lot of error messages?

umh-core/CHANGELOG.md
```diff
@@ -3,8 +3,13 @@
 ## Unreleased
 
 ### New Features
+
 - Preview: an instance can now store a connection to an external TimescaleDB or PostgreSQL historian database (host, port, database, login, and TLS mode), which can be created, viewed, updated, and removed. Storing the connection does not yet route any data to it
 
+### Fixes
+
+- A data contract with a missing data model no longer blocks your other contracts from getting their schemas. Previously one broken contract stopped schema registration for all of them, so the UNS output skipped validation and accepted wrong data types. Valid contracts now register normally; broken ones are logged and skipped
```
JeremyTheocharis: isnt it more like "previously when you had one msiconfigured data contract, it would not validate any data contract and instead skip all of them. this is now fixed by skipping the miscongifred data contract and emitting a warning"

JeremyTheocharis (changes requested):

umh-core/CHANGELOG.md
led0nk: you're totally right, i adjusted this as this was also not very helpful from users perspective

umh-core/pkg/service/redpanda/schema_registry.go
FalconTube: As discussed in Daily on 14.07, let's just wait here until the Logging PRs of both @JeremyTheocharis and me @FalconTube are merged :) 
Then we can use the `dedup` function or just have it fall in the general reconcile logic

led0nk: this is resolved now with using the dedup-logger

FalconTube (approved):
Tres bien!

JeremyTheocharis (approved):

PR #2646, opened 2026-07-14
docs: add Type Definitions reference page for data models

robertschwarz-umh:
Adds a single Reference page listing the types a data model field accepts, resolving a contradiction between two existing docs pages.

Closes ENG-5360.

## Problem

No single docs page listed the types a data model field accepts, and two pages disagreed:

- `payload-shapes` said only `string` + `number`
- `payload-formats` said `number` / `boolean` / `string`
- Neither mentioned `integer`, which the validator accepts

Users hit `_payloadshape: number` vs `timeseries-number` mix-ups and unexplained `referenced payload shape` / `unsupported UMH type` errors.

## Changes

- **New `usage/data-modeling/type-definitions.md`** (+ SUMMARY nav): the `_type` enum (`string`/`number`/`integer`/`boolean`), how `number` (decimal) differs from `integer`, the three built-in payload shapes, examples, and the validation-error table.
- **Resolve the contradiction**: `payload-shapes` and `payload-formats` now link to the new page for the full type list. Also fixes a `datamodels` → `dataModels` key-casing bug in the payload-shapes example.
- **Changelog** entry under `## Unreleased`.

Every fact verified against `pkg/datamodel/validator.go` and `pkg/datamodel/translator.go`. Docs only, no code change.

## Open questions for review

- **Surface the verbatim `_relational` error?** The validation-error table quotes `leaf nodes must contain _payloadshape, _relational, or _refModel` verbatim (the real string, for grep-accuracy). `_relational` is a canceled FF feature (ENG-4791); policy is not to document FF-gated features. Keep verbatim, or drop the row? Team decision.
- **`work-order` shape name** is defined differently on `payload-shapes` vs the new page. Follow-up cleanup, out of scope here.

### DIFF
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Improvements
+
+- New [Type Definitions](https://docs.umh.app/usage/data-modeling/type-definitions) documentation page listing the four field types (`string`, `number`, `integer`, `boolean`), how `number` (decimal) differs from `integer`, the three built-in payload shapes, and the data model validation errors. Previously this was split across the Payload Shapes and Payload Formats pages, which had slightly different type definitions and no mention of `integer`.
+
 ## [0.44.29]
 
 ### New Features
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -2,10 +2,6 @@
 
 ## Unreleased
 
-### Improvements
-
-- New [Type Definitions](https://docs.umh.app/usage/data-modeling/type-definitions) documentation page listing the four field types (`string`, `number`, `integer`, `boolean`), how `number` (decimal) differs from `integer`, the three built-in payload shapes, and the data model validation errors. Previously this was split across the Payload Shapes and Payload Formats pages, which had slightly different type definitions and no mention of `integer`.
-
 ## [0.44.29]
 
 ### New Features
--- a/umh-core/docs/reference/data-model-type-definitions.md
+++ b/umh-core/docs/reference/data-model-type-definitions.md
@@ -8,14 +8,13 @@ A field's value type is declared at one of two levels. A data model field refere
 
 A [payload shape](payload-shapes.md) field's `_type` is one of:
 
-| `_type`   | Meaning                   | Example value |
-|-----------|---------------------------|---------------|
-| `string`  | Text                      | `"running"`   |
-| `number`  | Decimal / floating point  | `42.5`        |
-| `integer` | Whole number, no fraction | `42`          |
-| `boolean` | True / false              | `true`        |
+| `_type`   | Meaning                             | Example value |
+|-----------|-------------------------------------|---------------|
+| `string`  | Text                                | `"running"`   |
+| `number`  | Any numeric value, decimal or whole | `42.5`, `42`  |
+| `boolean` | True / false                        | `true`        |
 
-`number` and `integer` are distinct: `number` accepts a fraction, `integer` does not. Any other value is rejected: `unsupported UMH type: <type>`.
+Any other value is rejected: `unsupported UMH type: <type>`.
 
 ## Timeseries payload shapes are inherently typed
 
@@ -103,7 +102,7 @@ payloadShapes:
       orderId:
         _type: string
       quantity:
-        _type: integer
+        _type: number
       price:
         _type: number
       active:
@@ -128,7 +127,7 @@ Rules a data model field must follow:
 | Error message | Cause | Fix |
 |---------------|-------|-----|
 | `referenced payload shape '<name>' does not exist` | `_payloadshape` names a shape that is not built-in and not defined | Use a built-in, or define the shape first |
-| `unsupported UMH type: <type>` | A field `_type` is not `string`/`number`/`integer`/`boolean` | Use one of the four allowed types |
+| `unsupported UMH type: <type>` | A field `_type` is not a supported type | Use `string`, `number`, or `boolean` |
 | `field cannot have both _payloadshape and _refModel` | A leaf field sets both keys | Keep one |
 | `leaf nodes must contain _payloadshape, _relational, or _refModel` | A leaf field has no value-type key | Add `_payloadshape` or `_refModel`, or give the field subfields |
 | `non-leaf nodes (folders) cannot have _payloadshape` | A field with subfields also sets `_payloadshape` | Remove `_payloadshape` from the folder |
@@ -140,3 +139,8 @@ Rules a data model field must follow:
 - [Data Models](data-models.md) - Structure, `_refModel`, versions
 - [Data Contracts](data-contracts.md) - Enforcement at ingress
 - [Payload Formats](../unified-namespace/payload-formats.md) - UNS payload structure
+
+> ### Note: the `integer` type
+> The YAML validator also accepts `_type: integer` inside a custom payload shape defined in an instance's config file.
+> However, there is no built-in `timeseries-integer` shape and the data-model editor does not offer it, so most models never need it.
+> Use `number` for numeric values. `integer` is only relevant when a custom payload shape must reject fractional values (for example a discrete count or an ID).
--- a/umh-core/docs/reference/data-model-type-definitions.md
+++ b/umh-core/docs/reference/data-model-type-definitions.md
@@ -1,4 +1,4 @@
-# Type Definitions
+# Data Model Type Definitions
 
 > This is a reference document. Use it to look up the types that a data model field accepts and the errors the validator may raise.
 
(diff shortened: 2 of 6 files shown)


umh-core/docs/reference/data-model-type-definitions.md
```diff
@@ -0,0 +1,142 @@
...
+A field's value type is declared at one of two levels. A data model field references a type with `_payloadshape` (a named payload shape) or `_refModel` (another data model). Inside a payload shape, each field declares a scalar directly with `_type`.
+
+## Field Types (`_type`)
+
+A [payload shape](payload-shapes.md) field's `_type` is one of:
+
+| `_type`   | Meaning                   | Example value |
+|-----------|---------------------------|---------------|
+| `string`  | Text                      | `"running"`   |
+| `number`  | Decimal / floating point  | `42.5`        |
```
JeremyTheocharis: do we sctually have a difference between number and integer? that looks like it shouldnt be. it should just be number (too complex otherwise).

umh-core/docs/reference/data-model-type-definitions.md
```diff
@@ -0,0 +1,142 @@
+# Type Definitions
+
+> This is a reference document. Use it to look up the types that a data model field accepts and the errors the validator may raise.
```
JeremyTheocharis, line 3: shouldnt this then go into the referneces section?

umh-core/docs/reference/data-model-type-definitions.md
```diff
@@ -0,0 +1,142 @@
...
+          order:
+            _payloadshape: work-order
+```
+
+Rules a data model field must follow:
+
+- A leaf field references its type with `_payloadshape` or `_refModel`, never both.
+- A folder (a field with subfields) has neither `_payloadshape` nor `_refModel`.
+
+## Validation Errors
```
JeremyTheocharis, line 125: might make sense to leave in the code a reference to this page, so future error changes are refclected in ehre as well (so an LLM will auotmatically change it here, or alternatively coderabbit can check for it)

umh-core/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Improvements
+
+- New [Type Definitions](https://docs.umh.app/usage/data-modeling/type-definitions) documentation page listing the four field types (`string`, `number`, `integer`, `boolean`), how `number` (decimal) differs from `integer`, the three built-in payload shapes, and the data model validation errors. Previously this was split across the Payload Shapes and Payload Formats pages, which had slightly different type definitions and no mention of `integer`.
```
JeremyTheocharis: docs doesn't get a changelog entry, you can use the skip-changelog label :)

umh-core/docs/reference/data-model-type-definitions.md
```diff
@@ -0,0 +1,142 @@
...
+    description: Work order record
+    fields:
+      orderId:
+        _type: string
+      quantity:
+        _type: integer
+      price:
+        _type: number
+      active:
+        _type: boolean
```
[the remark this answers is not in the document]
robertschwarz-umh: Good take on consistency, but I disagree. If my intention was to show the example of `payload-shapes.md` in the `type-definitions` file, then this example would have been redundant and solved by just linking `payload-shapes.md`.

My actual intent is to showcase the availability of the `integer` parameter, and to exemplify the difference between `number` and `integer` when building data models and a payloadShape.

Human reader can see example and instinctively understand, that:
quantity: always whole pieces -> 1, 10, 20, 50, 621 (etc)
price: could be fractions -> 0.14, 0.73, 1.90 (etc)


If you insist, a better solution is to rename the custom payloadShape in `type-definitions.md` to something other than `work-order` this completely avoids the confusion/inconsistency aspect.

umh-core/docs/reference/data-model-type-definitions.md
```diff
@@ -0,0 +1,142 @@
...
+- A folder (a field with subfields) has neither `_payloadshape` nor `_refModel`.
+
+## Validation Errors
+
+| Error message | Cause | Fix |
+|---------------|-------|-----|
+| `referenced payload shape '<name>' does not exist` | `_payloadshape` names a shape that is not built-in and not defined | Use a built-in, or define the shape first |
+| `unsupported UMH type: <type>` | A field `_type` is not `string`/`number`/`integer`/`boolean` | Use one of the four allowed types |
+| `field cannot have both _payloadshape and _refModel` | A leaf field sets both keys | Keep one |
+| `leaf nodes must contain _payloadshape, _relational, or _refModel` | A leaf field has no value-type key | Add `_payloadshape` or `_refModel`, or give the field subfields |
```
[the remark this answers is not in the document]
robertschwarz-umh: This is TBD with human reviewers and part of the PR comment's "Open Questions".

umh-core/docs/reference/data-model-type-definitions.md
robertschwarz-umh: Agree 100%, thank you! I was tunnel-visioning on the `Data Modeling` portion of our docs. I'll move  it.

umh-core/docs/reference/data-model-type-definitions.md
robertschwarz-umh: I like, I'll check if and how we do it for other doc pages and build something based on that. If there's no previous docs I can use as examples, I'll try to come up with something and comment here again.

umh-core/CHANGELOG.md
robertschwarz-umh: Ah, my bad. (Un)Done here b98d1593c11ebe87934662d674eeaf8f0725df74!

umh-core/docs/reference/data-model-type-definitions.md
robertschwarz-umh: Thanks for checking this! It sort of exists [here](https://github.com/united-manufacturing-hub/united-manufacturing-hub/blob/162f84039d0a7881548c7e6244bdfbf218466cab/umh-core/pkg/datamodel/translator.go#L468
).

When creating a custom payload shape in my instance file config, I'm able to supply `integer` as a `_type`. In next step, I'm also able to use that payload shape just fine in Data Models without causing warns/degradations in the instance.

To simplify things, wdyt about removing all mentions of integer and just add an FIY about them at the very bottom of `type-definitions.dm`?

**Example**:
```markdown
[...] remove `integer` from main table
| `_type`   | Meaning                             | Example value |
|-----------|-------------------------------------|---------------|
| `string`  | Text                                | `"running"`   |
| `number`  | Any numeric value, decimal or whole | `42.5`, `42`  |
| `boolean` | True / false                        | `true`        |

[...] rest of article, no more mention of integers
[...] end of article / footer

> ### Note: the `integer` type
> The YAML validator also accepts `_type: integer` inside a custom payload shape defined in an instance's config file. 
> However, there is no built-in `timeseries-integer` shape and the data-model editor does not offer it, so most models never need it.
> Use `number` for numeric values. `integer` is only relevant when a custom payload shape must reject fractional values (for example a discrete count or an ID)
```

umh-core/docs/reference/data-model-type-definitions.md
robertschwarz-umh: Done! 58042db470805d4db4233916d5767fd557d5c55c

umh-core/docs/reference/data-model-type-definitions.md
robertschwarz-umh: Klaus and I gave it some thought, **TL;DR:**
- add code comment to `validator.go` & `translator.go`
- add CodeRabbit config to check if docs are updated when `validator.go` and/or `translator.go` are changed

@JeremyTheocharis: wdyt? 👍 or 👎?

---

**Long version:**
It looks like this would be the first time we're creating reference to public docs. I'd personally prefer a good old comment in the codebase.

Claude suggests to instead configure CodeRabbit to check changes on the validator and translator.

I think both at the same time is best: 
- a contributor working on `translator.go` or `validator.go` will not check CodeRabbit config, but they'll see the comment
- if contributor misses the comment and doesn't update, CodeRabbit still catches it

JeremyTheocharis: * add code comment to `validator.go` & `translator.go`<br>→ if you do this, you dont need to add a manual coderabbit check (and it will also not trigger deterministically on how LLMs work). but the more intuitive it is for a human to be reminded of that there is a documentation page, the higher teh chances the LLM models either during creation, during /sef-review or at least in coderabbit will catch it

robertschwarz-umh: Makes sense, ty! @Linear - create a ticket for this and assign it to me. Don't want to mix code updates with doc updates in same branch. Will create separate branch for code comments that mention the reference.

umh-core/docs/reference/data-model-type-definitions.md
robertschwarz-umh: Addressed here: fd78b7b229f37178ab571b3f097b8854954decde

JeremyTheocharis (approved):

PR #2647, opened 2026-07-15
Eng 5366 improve umh core logging structure dedup sampling noise

FalconTube:
Closes ENG-5366

Look there for detailed info

Logs look like this now
grafik

### DIFF
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -5,6 +5,9 @@
 ### Improvements
 
 - Persistent connection failures now log the HTTP status code and a sanitized snippet of the upstream response, so the cause of a flapping connection (such as a 502 from a reverse proxy) is visible in the worker's logs
+- The umh-core log is now easier to scan: each entry lines up in fixed `[level] [caller] [component]` columns
+- Repeated reconcile errors no longer flood the umh-core log: the first occurrence is logged at ERROR with a note, then demoted to DEBUG until the error changes or clears, with an ERROR summary at most once a minute reporting how many repeats were suppressed so an ongoing failure stays visible
+- The umh-core log is quieter: high-frequency per-tick status detail moved from INFO to DEBUG, and repeated lines below WARN are rate-limited — warnings and errors are never dropped
 
 ## [0.44.29]
 
--- a/umh-core/internal/fsm/baseFSM.go
+++ b/umh-core/internal/fsm/baseFSM.go
@@ -37,6 +37,10 @@ import (
 // BaseFSMInstance implements the public fsm.FSM interface.
 type BaseFSMInstance struct {
 
+	// lastSuppressionSummary is the time the most recent suppression summary was
+	// emitted, used to throttle the periodic WARN summary.
+	lastSuppressionSummary time.Time
+
 	// fsm is the finite state machine that manages instance state
 	fsm *fsm.FSM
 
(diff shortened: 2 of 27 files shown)


umh-core/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Improvements
+
+- The umh-core log is now easier to read and far less noisy: entries line up in fixed `[level] [caller] [component]` columns, repeated reconcile errors are logged once at ERROR (with a note) and then demoted to DEBUG until they change or clear, high-frequency per-tick status detail moved from INFO to DEBUG, and repeated log lines below WARN are rate-limited — warnings and errors are never dropped
```
JeremyTheocharis: i feel these are three different changes at the same time? maybe pslit it up in three:

* Log output is now easier to scan with fixed `[level] [caller] [component]` columns.
* High-frequency status and heartbeat-style messages were reduced by moving recurring detail to DEBUG and rate-limiting repeated logs below WARN.
* Repeated reconcile errors are now surfaced once at ERROR, then repeated unchanged occurrences are logged at DEBUG until they change or clear.

umh-core/internal/fsm/baseFSM.go
```diff
@@ -346,6 +355,32 @@ func (s *BaseFSMInstance) ShouldSkipReconcileBecauseOfError(tick uint64) bool {
...
+// LogErrorDedup keeps reconcile-loop errors visible without spamming the log:
+//   - first occurrence of a message: Error, logged plainly
+//   - first repeat: Error, with a suffix announcing that further repeats are
+//     demoted to Debug, so a reader following the error log knows suppression
+//     has kicked in
+//   - subsequent repeats: Debug
+//
+// A changed message, or ResetState after a successful reconcile, resets the
+// cycle so the next error is again logged plainly at Error.
+func (s *BaseFSMInstance) LogErrorDedup(format string, args ...any) {
```
JeremyTheocharis, line 384: Here’s a review comment I’d leave:

I’m a bit concerned about the repeated reconcile-error suppression behavior here. Right now we log the first occurrence at `ERROR`, the first repeat at `ERROR` with a suppression note, and then all subsequent identical occurrences get demoted to `DEBUG`. That reduces noise, but at normal log levels we lose visibility into **how often** the same failure is still happening.

A more robust pattern is to keep a suppressed counter and emit a periodic summary while suppression is active, plus a final summary when the error changes or clears. For example: first occurrence at `ERROR`, an early “repeats are being suppressed” message, then every N occurrences or every X seconds emit something like “repeated reconcile error suppressed 127 times”, and finally “error cleared after 842 suppressed repeats”.

That keeps the logs readable without hiding the scale or persistence of an ongoing failure.

Idea: if there was an error message and the last emit was over 60 secodns emit a new message as WARN that xxx messgaesi n the alst 60 seconds were swallowed

JeremyTheocharis (changes requested):

umh-core/internal/fsm/baseFSM.go
FalconTube: Yes, very good idea.

But instead of exposing the "repeated reconcile, suppredded" via `WARN`, let's even do it as `ERROR`. Then it's very easy to catch them.

Also makes it clear, that there are still errors.

Looks like this then:

```log
2026-07-16 14:14:10.916498873  [ERROR]  [fsm/baseFSM.go:408]                     [protocolconverter-schneeflocke]              repeated reconcile error suppressed 596 times in the last 1m0s (still failing): test error
```

umh-core/pkg/fsm/agent_monitor/reconcile.go
```diff
@@ -90,7 +90,7 @@ func (a *AgentInstance) Reconcile(ctx context.Context, snapshot fsm.SystemSnapsh
 
 		// Log the error but always continue reconciling - we need reconcileStateTransition to run
 		// to restore services after restart, even if we can't read their status yet
-		a.baseFSMInstance.GetLogger().Warnf("failed to update observed state (continuing reconciliation): %s", err)
+		a.baseFSMInstance.GetLogger().Debugf("failed to update observed state (continuing reconciliation): %s", err)
```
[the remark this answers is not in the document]
FalconTube: Will not do this now. Might do this in a follow up PR, but only when needed. Because then we can change that on lots of places

JeremyTheocharis (approved):

PR #2654, opened 2026-07-17
dedup to separate logger

FalconTube:
- Exposes the `LogErrorDedup` function outside of the `reconcile.go` files

### DIFF
--- a/umh-core/pkg/logger/dedup.go
+++ b/umh-core/pkg/logger/dedup.go
@@ -49,6 +49,12 @@ const DefaultErrorDedupResetInterval = 5 * time.Minute
 type DedupLogger struct {
 	*zap.SugaredLogger
 
+	// dedupSugar is SugaredLogger with one extra caller-skip frame, used for the
+	// LogErrorDedup path so the reported caller is the code that called
+	// LogErrorDedup rather than dedup.go itself. Promoted methods on the embedded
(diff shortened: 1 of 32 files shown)


umh-core/pkg/logger/dedup.go
```diff
@@ -0,0 +1,188 @@
...
+// interval restarts deduplication automatically, so callers never need to call
+// Reset by hand.
+const DefaultErrorDedupResetInterval = 5 * time.Minute
+
+// DedupLogger is a *zap.SugaredLogger that additionally deduplicates repeating
+// error logs from a hot loop (such as an FSM reconcile loop). It embeds the
+// SugaredLogger, so all standard logging methods are available directly, plus
+// LogErrorDedup for the deduplicated path.
+//
+// LogErrorDedup logs the first occurrence and first repeat at Error, demotes
```
led0nk: Can we in general here reduce the comments to basically just state -> why do we need this?

Imo the rest is overhead

led0nk (approved):
Other than that works for me pretty reliably and like it!

umh-core/pkg/logger/dedup.go
FalconTube: I agree. Will clean up the comments and then merge it.
Will ping you once it is in, so that you can continue with your branch!
Thank you for the review :)

PR #2660, opened 2026-07-20
Historian bridge: inherit connection detected from the write DFC

Aaron99B:
## What

A bridge that writes to the historian inherits its connection health check from the instance's stored historian connection instead of a user-supplied host/port. Such a bridge is recognised by its write DFC's destination protocol (`historian`); its Nmap target then resolves from the shared historian section (`{{ .historian.timescale.host }}` / `.port`) at render time, so the check follows the historian and can't be pointed at the wrong host.

No new wire field — `ProtocolConverterConnection` stays `{ ip, port }`. Pairs with the ManagementConsole PR (same branch name) that adds the "Historian (auto-connect)" template.

## How

- **Detection:** `DataflowComponentWriteConfigInput.WritesToHistorian()` is `Destination.Protocol == "historian"`. Matching bridges inherit the historian connection; every other bridge uses the standard `{{ .IP }}/{{ .PORT }}` template. Nothing is persisted to mark the choice.
- **Deploy / edit** select the connection template from the write DFC. An edit with no write DFC in the payload re-derives from the stored config, so a state- or read-only edit preserves a historian bridge's inherited connection.
- **No IP/PORT stored** for a historian bridge: deploy skips the connection variables, edit strips any left by an older deploy, and config-change detection ignores them.
- **Display:** `get-protocolconverter` and the status generator report the concrete host:port the connection FSM resolved (empty/0 until it has), never the raw template placeholder. `get`'s `variables` is emitted as `[]` rather than a nil slice.

The frontend still sends the concrete host/port, so deploy `Validate()` and the edit rollout-wait are unchanged — the write DFC drives template selection, IP/PORT persistence, and display only.

## Testing

- Unit: protocol-based historian detection (including a non-historian bridge that references `{{ .historian.* }}` and does *not* inherit), end-to-end render resolving to the configured historian, the actionable error when no historian is configured, and no IP/PORT variables stored for a historian bridge.
- `go build ./...`, `go vet`, `golangci-lint`, and the affected package tests (`communicator/actions`, `communicator/pkg/generator`, `config/dataflowcomponentserviceconfig`) pass.

ENG-5368

### DIFF
--- a/umh-core/pkg/communicator/actions/connectiontemplate_historian_test.go
+++ b/umh-core/pkg/communicator/actions/connectiontemplate_historian_test.go
@@ -1,0 +1,62 @@
+// Copyright 2025 UMH Systems GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package actions
+
+import (
+	ginkgo "github.com/onsi/ginkgo/v2"
+	gomega "github.com/onsi/gomega"
+
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/config/dataflowcomponentserviceconfig"
+)
+
+func historianWriteDFC() dataflowcomponentserviceconfig.DataflowComponentWriteConfigInput {
+	return dataflowcomponentserviceconfig.DataflowComponentWriteConfigInput{
+		Destination: dataflowcomponentserviceconfig.WriteConfigDestination{
+			Protocol: "historian",
+			Code:     "host: '{{ .historian.timescale.host }}'\ndata_contract_name: pump",
+		},
+		Source: dataflowcomponentserviceconfig.WriteConfigSource{Topics: "umh.v1.*"},
+	}
+}
+
+func standardWriteDFC() dataflowcomponentserviceconfig.DataflowComponentWriteConfigInput {
+	return dataflowcomponentserviceconfig.DataflowComponentWriteConfigInput{
+		Destination: dataflowcomponentserviceconfig.WriteConfigDestination{
+			Protocol: "kafka",
+			Code:     "topic: t",
+		},
+		Source: dataflowcomponentserviceconfig.WriteConfigSource{Topics: "umh.v1.*"},
+	}
+}
+
+var _ = ginkgo.Describe("writesToHistorian", func() {
+	ginkgo.It("detects a write DFC whose destination protocol is the historian plugin", func() {
+		gomega.Expect(writesToHistorian(historianWriteDFC())).To(gomega.BeTrue())
+	})
+
+	ginkgo.It("ignores a bridge with a different destination protocol", func() {
+		gomega.Expect(writesToHistorian(standardWriteDFC())).To(gomega.BeFalse())
+	})
+
+	ginkgo.It("ignores a non-historian bridge even if its code references the historian scope", func() {
+		dfc := dataflowcomponentserviceconfig.DataflowComponentWriteConfigInput{
+			Destination: dataflowcomponentserviceconfig.WriteConfigDestination{
+				Protocol: "kafka",
+				Code:     "host: '{{ .historian.timescale.host }}'",
+			},
+		}
+		gomega.Expect(writesToHistorian(dfc)).To(gomega.BeFalse())
+	})
+})
--- a/umh-core/pkg/communicator/actions/deploy-protocolconverter.go
+++ b/umh-core/pkg/communicator/actions/deploy-protocolconverter.go
@@ -340,13 +340,28 @@ func (a *DeployProtocolConverterAction) createProtocolConverterConfig() (config.
 // configuration from a parsed protocol converter payload.
 func buildProtocolConverterConfig(payload models.ProtocolConverter) (config.ProtocolConverterConfig, error) {
 	userVars := buildUserScope(payload.TemplateInfo)
-	userVars["IP"] = payload.Connection.IP
-	userVars["PORT"] = strconv.FormatUint(uint64(payload.Connection.Port), 10)
+
+	var writeDFC dataflowcomponentserviceconfig.DataflowComponentWriteConfigInput
+	if w := payload.WriteDFCPayload; w != nil {
+		writeDFC = w.DataflowComponentWriteConfigInput
+	}
+
+	// Resolve the connection template once from the write DFC. A bridge that writes to the
+	// historian inherits the historian connection and stores no IP/PORT variables; any other
+	// bridge uses the {{ .IP }}/{{ .PORT }} template and stores them. The same decision drives
+	// both the variable merge and the stored connection template.
+	connectionTemplate := newIPPortConnectionTemplate()
+	if writesToHistorian(writeDFC) {
+		connectionTemplate = historianConnectionTemplate()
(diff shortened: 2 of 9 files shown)


umh-core/pkg/communicator/actions/deploy-protocolconverter.go
```diff
@@ -287,13 +287,25 @@ func (a *DeployProtocolConverterAction) createProtocolConverterConfig() (config.
...
+	}
+
+	// Resolve the connection template once from the write DFC. A bridge that writes to the
+	// historian ({{ .historian.* }}) inherits the historian connection and stores no IP/PORT
+	// variables; any other bridge uses the {{ .IP }}/{{ .PORT }} template and stores them.
+	connectionTemplate := connectionTemplateForWriteDFC(writeDFC)
+	if !writesToHistorian(writeDFC) {
+		userVars["IP"] = payload.Connection.IP
+		userVars["PORT"] = strconv.FormatUint(uint64(payload.Connection.Port), 10)
+	}
```
FalconTube, line 359: This part here is very hard to follow/understand. See suggestion
```suggestion
    // Resolve the connection template once from the write DFC. A bridge that writes to the
	// historian ({{ .historian.* }}) inherits the historian connection and stores no IP/PORT
	// variables; any other bridge uses the {{ .IP }}/{{ .PORT }} template and stores them. The
	// same decision drives both the variable merge and the stored connection template.
	connectionTemplate := newIPPortConnectionTemplate()
	if writesToHistorian(writeDFC) {
		connectionTemplate = historianConnectionTemplate()
	} else {
		userVars["IP"] = payload.Connection.IP
		userVars["PORT"] = strconv.FormatUint(uint64(payload.Connection.Port), 10)
	}

```

umh-core/pkg/communicator/actions/protocolconverter_helpers.go
```diff
@@ -87,6 +87,37 @@ func newIPPortConnectionTemplate() connectionserviceconfig.ConnectionServiceConf
...
+// write DFC. A bridge that writes to the historian inherits the historian connection; any other
+// bridge uses the standard {{ .IP }}/{{ .PORT }} template fed by the connection variables.
+func connectionTemplateForWriteDFC(writeDFC dataflowcomponentserviceconfig.DataflowComponentWriteConfigInput) connectionserviceconfig.ConnectionServiceConfigTemplate {
+	if writesToHistorian(writeDFC) {
+		return historianConnectionTemplate()
+	}
+
+	return newIPPortConnectionTemplate()
+}
+
```
FalconTube: This helper function is not needed any more, if you follow my suggestion in `deploy-protocolconverter.go`

umh-core/pkg/communicator/pkg/generator/protocolconverter.go
```diff
@@ -121,6 +122,17 @@ func buildProtocolConverterAsDfc(
...
+		// {{ .historian.timescale.host }}) that the variable substitution above cannot
+		// resolve. Use the rendered connection config, which holds the concrete host:port
+		// the connection FSM is probing, so the bridges list follows the shared connection.
+		if observed.ObservedProtocolConverterSpecConfig.Config.DataflowComponentWriteServiceConfig.ReferencesHistorian() {
+			if resolved := observed.ServiceInfo.ConnectionObservedState.ObservedConnectionConfig.NmapServiceConfig; resolved.Target != "" {
+				target = resolved.Target
+				port = strconv.Itoa(int(resolved.Port))
+			}
+		}
+
```
FalconTube: This section here is also hard to follow, because you first generate `target` and `port` and then overwrite if for historian case.
Better split it up into an `if - else` like this:

```go
var target, port string
if referencesHistorian{
	// historian logic
	target =
	port =
} else {
	// "normal" logic
	target =
	port =
}
```

umh-core/pkg/config/dataflowcomponentserviceconfig/write_config.go
```diff
@@ -15,11 +15,17 @@
...
+	"regexp"
 	"strings"
 
 	"gopkg.in/yaml.v3"
 )
 
+// historianTemplateRefPattern matches a Go-template reference to the reserved `historian`
+// render scope (e.g. {{ .historian.timescale.host }}), which the runtime resolves from the
+// instance's shared historian: config section.
+var historianTemplateRefPattern = regexp.MustCompile(`\{\{-?\s*\.historian\b`)
```
FalconTube: what is the `\b` at the end for?

umh-core/pkg/config/dataflowcomponentserviceconfig/write_config.go
```diff
@@ -136,6 +142,16 @@ func (c DataflowComponentWriteConfigInput) HasOutput() bool {
...
+// ReferencesHistorian reports whether this write DFC references the reserved historian render
+// scope ({{ .historian.* }}). A bridge writing to the historian is exactly one that references
+// that scope, so it can be recognised without any explicit marker on the config.
+func (c DataflowComponentWriteConfigInput) ReferencesHistorian() bool {
+	return historianTemplateRefPattern.MatchString(c.Destination.Code) ||
+		historianTemplateRefPattern.MatchString(c.Destination.Protocol) ||
+		historianTemplateRefPattern.MatchString(c.Processing.Code) ||
+		historianTemplateRefPattern.MatchString(c.Source.Topics)
+}
+
```
FalconTube: Idea: Don't do a regex matcher here. Instead, just check if the `c.Destination.Protocol` equals `historian`. Then we know, that it is our historian plugin and this is the one that we support.
For all other stuff, people just have to fill stuff manually i.e. we only support this functionality with the IP/PORT matching for our historian plugin

FalconTube (changes requested):

umh-core/pkg/communicator/actions/deploy-protocolconverter.go
Aaron99B: included that

umh-core/pkg/communicator/actions/protocolconverter_helpers.go
Aaron99B: removed it

umh-core/pkg/communicator/pkg/generator/protocolconverter.go
Aaron99B: done

umh-core/pkg/config/dataflowcomponentserviceconfig/write_config.go
Aaron99B: \b is a word-boundary anchor. It asserts a transition between a word character ([A-Za-z0-9_]) and a non-word character. So after the historian there must be a . or empty space something like "historians" won't be matched

umh-core/pkg/config/dataflowcomponentserviceconfig/write_config.go
Aaron99B: Good idea. Changed that

umh-core/pkg/config/dataflowcomponentserviceconfig/write_config.go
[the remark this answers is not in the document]
Aaron99B: removed the regEx check

umh-core/pkg/communicator/actions/connectiontemplate_historian_test.go
```diff
@@ -0,0 +1,66 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package actions
+
+// This test lives in package actions (not actions_test) to exercise the unexported
+// writesToHistorian helper directly. The actions package declares its own Label
```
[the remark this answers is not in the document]
Aaron99B: done

FalconTube (approved):
Looks good now!
Found some issues with the historian benthos-umh plugin itself, but not blocking for this PR.

PR #2669, opened 2026-07-29
fix(bridge): roll back a renamed bridge under its new UUID (ENG-5444)

Aaron99B:
## TL;DR
- Today renaming a bridge whose deployment fails results in an endless loop on the save anyway button, because the UUID in the frontend for the rollback cannot be found aynmore
- Now we changed the backend to not use the saved UUID for the rollback, but instead derive the UUID from the name
- We also cleaned up the error output by moving the error message to a constant 

Fixes [ENG-5444](https://linear.app/united-manufacturing-hub/issue/ENG-5444/cannot-save-modbus-bridge-after-failed-rename-deployment). Frontend half: ManagementConsole#3367.

## What breaks today

Rename a bridge whose deployment then fails, and the bridge becomes unsaveable. umh-core reports the edit as rolled back, but keeps the new name. The frontend still holds the pre-rename UUID, so every later save fails with `bridge with UUID ... not found`, including Save Anyway, which re-sends the same UUID. The only way out is to leave the page.

Reported against a Modbus bridge whose first deploy failed on a degraded WAGO connection (ENG-5446), but the trigger is the rename, not the protocol.

## Root cause

A bridge's UUID is a hash of its name (`GenerateUUIDFromName`), so a rename changes its identity. `Execute` persists the rename first, then waits for the rollout, then rolls back on failure:

- `persistConfig` commits the new name; the bridge now answers to `uuid(new-name)`
- `awaitRollout` times out (connection never reaches `active`)
- `rollbackEdit` looks the bridge up under `a.atomicEditUUID`, the pre-rename UUID, and misses

The rollback returns `protocol converter with UUID ... not found`, the config keeps the new name, and the failure carries no error code, so the frontend treats it as ignorable and keeps offering Save Anyway for a request that can never succeed.

## Changes

**Roll back under the name the edit persisted.**

```go
_, err := a.configManager.AtomicEditProtocolConverter(
    ctx, dataflowcomponentserviceconfig.GenerateUUIDFromName(a.name), pcConfig)
```

`persistConfig` guarantees the bridge is stored under `a.name`, so the lookup always resolves; `pcConfig` is the pre-edit config, so the restore puts the old name and old UUID back. When the name did not change, this is the same value `atomicEditUUID` held, so behaviour is unchanged. All three abort paths (rollout timeout, render fail-fast, Benthos config error) share `rollbackEdit`, so one change covers them. `atomicEditUUID` existed only to carry that value and is now removed; `persistConfig` keeps taking the pre-rename UUID, which is correct for it.

`edit-dataflowcomponent.go` already keys its rollback off the post-rename UUID and documents why.

**Report a missing bridge with a code.** `applyMutation` wraps a sentinel error, leaving the message text unchanged, and `Execute` sends `ERR_BRIDGE_NOT_FOUND` so the frontend can stop offering a retry that cannot succeed. It does not let the frontend recover: the reply carries no payload, so there is no replacement UUID to follow. See the second known gap below.

## Resulting behaviour

- Save Anyway not pressed: the rename is reverted with the rest of the configuration, and the bridge keeps serving its previous config
- Save Anyway pressed: `ignoreErrors` skips the rollout wait, so the rename is stored and the reply carries the new UUID

## Tests

Four specs under `EditProtocolConverter > Execute > when the edited bridge was renamed`:

- rename plus failed rollout restores the pre-rename name
- the pre-rename UUID stays resolvable, so a retry reaches the rollout rather than dying on a dead UUID
- with the read flow ignoring errors, the rename is persisted without waiting for the rollout
- a stale UUID reports `ERR_BRIDGE_NOT_FOUND`

The first two fail on staging with the exact strings from the report. Full `pkg/communicator/actions` suite passes; the render fail-fast specs still assert their existing rollback behaviour unchanged.

## Known gaps, deliberately out of scope

**`ERR_BRIDGE_NOT_FOUND` tells the frontend to stop, not where to go** (ENG-5635, raised by CodeRabbit on ManagementConsole#3367). The reply passes `nil` as `payloadV2`, so the frontend cannot learn the bridge's new UUID and cannot offer to follow it; it drops the user back onto the same stale editor. Neither flow in this ticket reaches that state, since a failed deploy now reverts the rename and Save Anyway returns the post-rename UUID. It needs the bridge to move from elsewhere, such as a second tab or a direct config edit. Attaching the current UUID to this reply would be one way to close it, and is worth deciding on its own rather than here.

Save Anyway is a no-op for a bridge with **no flow configured at all**: `ignoreErrors` only arrives inside `readDFC`/`writeDFC`, so such a request cannot express "skip the health check" and waits out the full rollout. Closing that needs an action-scoped field on the wire; worth its own ticket rather than riding along here.

### DIFF
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixes
 
-- Renaming a bridge whose deployment then fails no longer leaves it unsaveable. The rename was kept even though the deployment was reported as rolled back, so every later save (including Save Anyway) failed with "bridge not found" and the only way out was to leave the page. The rename is now reverted with the rest of the configuration, so retrying works
+- Renaming a bridge whose deployment then fails no longer leaves it unsaveable. The rename was kept even though the deployment was reported as rolled back, so every later save failed with "bridge not found" and the only way out was to leave the page. A failed deployment now reverts the rename along with the rest of the configuration, and Save Anyway stores the new name as before
 
 ## [0.44.34]
 
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Renaming a bridge whose deployment then fails no longer leaves it unsaveable. The rename was kept even though the deployment was reported as rolled back, so every later save (including Save Anyway) failed with "bridge not found" and the only way out was to leave the page. The rename is now reverted with the rest of the configuration, so retrying works
+
 ## [0.44.34]
 
 ### Improvements
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixes
 
-- Renaming a bridge whose deployment then fails no longer leaves it unsaveable. The rename was kept even though the deployment was reported as rolled back, so every later save failed with "bridge not found" and the only way out was to leave the page. A failed deployment now reverts the rename along with the rest of the configuration, and Save Anyway stores the new name as before
+- Renaming a bridge whose deployment then fails no longer leaves it impossible to edit or save. A failed deployment now undoes the rename along with the rest of the configuration, while Save Anyway keeps the new name as before
 
 ## [0.44.34]
 
--- a/umh-core/pkg/communicator/actions/edit-protocolconverter.go
+++ b/umh-core/pkg/communicator/actions/edit-protocolconverter.go
@@ -57,6 +57,8 @@ import (
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/service/s6"
 )
 
+var errBridgeNotFound = errors.New("not found")
+
 // DFCType represents the type of dataflow component configuration.
 type DFCType string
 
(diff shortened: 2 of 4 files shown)


umh-core/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Fixes
 
 - A misconfigured data contract no longer stops your other data contracts from being set up. Valid contracts are now registered and validated as usual, while the misconfigured one is skipped and reported as a warning
+- Renaming a bridge whose deployment then fails no longer leaves it unsaveable. The rename was kept even though the deployment was reported as rolled back, so every later save failed with "bridge not found" and the only way out was to leave the page. A failed deployment now reverts the rename along with the rest of the configuration, and Save Anyway stores the new name as before
```
JeremyTheocharis: maybe we can explain a little bit better what the preivous user flow was?<br><br>Previously, if you renamed a bridge and the deployment failed, it could become impossible to edit or save afterwards.<br>This fix makes sure the rename is properly rolled back when deployment fails, while “Save Anyway” still keeps the new name.

JeremyTheocharis (changes requested):

led0nk (approved):
works for me, but agree on that we need to refine the changelog-entry

umh-core/CHANGELOG.md
Aaron99B: changed that

JeremyTheocharis (approved):

PR #2672, opened 2026-07-30
feat: throttling logger for services

FalconTube:
## Problem

* Logs lines like `not enough time left to reconcile instance` and `Context deadline exceeded` are spamming the Logs on CPU-starved instances in such a way, that they are unusable
* Users can't debug problems any more

## What we are shipping

* Second, custom logger to deal with flood of logs from `reconcile` loops and `context deadline exceeded`.
* Has functions for throttling logs at all levels
* Has a `EscalatingXXX()` functions, which raise the info level to `ERROR` if too many logs occur in a short amount of time

## What we are not shipping

* We don't implement this through the entire codebase, just in the two most critical sites for now. These are sites, which were reported by the Issuer Merle

All info in [ENG-5465](https://linear.app/united-manufacturing-hub/issue/ENG-5465/introduce-deduplogger-for-opc-ua-bridge-components)

Closes [ENG-5465](https://linear.app/united-manufacturing-hub/issue/ENG-5465/introduce-deduplogger-for-opc-ua-bridge-components)


JeremyTheocharis (approved):
Approving on behalf of Jeremy while Yannic is on vacation.

The changelog entry was rewritten to describe the actual problem this PR solves: on a CPU-starved host, "context deadline exceeded" / "not enough time to reconcile" were logged once per instance per tick, flooding the log. The fix keeps that surfaced but throttled — one warning per minute with the suppressed count folded in, escalating to an error when the overload persists.

This PR is ready to merge — all CI checks pass (Unit Tests, Build arm64/amd64, changelog guard). Two non-blocking follow-ups tracked for Yannic's return:

1\. The throttled/escalating messages drop per-instance location/name context, so when one does fire you can't tell which instance/reconcile is stuck. Decide whether the escalate path should carry the instance name.

2\. Add a test covering \`Reset\` with the same format but changed arguments (the \`format != lastLoggedKey\` branch under a pending Reset isn't directly covered).

umh-core/CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 - The TimescaleDB historian output no longer warns when a batch holds only other contracts' data; it logs once when it stores its first message, and once if data arrives but none matches the configured data contract
 - The tag processor no longer converts numeric-looking string values to numbers; a string read from the source stays a string in the UNS, and msg.meta.datatype = "number" remains available for explicit coercion
 - A misconfigured data contract no longer stops your other data contracts from being set up. Valid contracts are now registered and validated as usual, while the misconfigured one is skipped and reported as a warning
+- Further suppression of noisy log messages, especially on CPU-starved instances
```
JeremyTheocharis: should be adjusted to the problem in the PR

PR #2675, opened 2026-07-31
fsmv2/simple: per-instance dependencies via a NewDeps builder

JeremyTheocharis:
Closes ENG-5483. Spec is the document on the ticket.

## Problem

All instances of a `simple` worker share their dependencies with each other, which is fine for the current workers: nmap has no state, and historian has state but only ever runs as one instance. Workers like benthos_monitor require state for each individual worker that should not be shared.

## What we're shipping

- `MonitorSpec.Deps` becomes `NewDeps func(id deps.Identity, bd *deps.BaseDependencies) TDeps`, called once per worker instance at construction. Breaking: the field is removed.
- That value now lives in `WorkerBase`'s deps slot instead of a field of our own, so a simple worker holds its dependencies the same way every other worker does. `bd` is the framework's own dependencies for the instance, so a worker can take the logger from there rather than a package-level one. historian does.
- The framework guide no longer tells new workers to override `GetDependenciesAny()` to nil. Both `struct{}{}` and `nil` fail the collector's check identically, so the override never did anything.

## What we're NOT shipping

- Framework metrics for workers that declare no dependencies. The collector still decides that from the worker's bound deps, so a `TDeps` of `struct{}` gets none. Making it automatic for every worker is a framework change and follows separately.
- benthos_monitor, which stacks on top of this and is the worker that needs per-instance state.

### DIFF
--- a/umh-core/pkg/fsmv2/CLAUDE.md
+++ b/umh-core/pkg/fsmv2/CLAUDE.md
@@ -244,9 +244,9 @@ func (w *MyWorker) GetDependencies() *MyDependencies {
 | `DeriveDesiredState` | Custom children specs (application, communicator) or non-standard config parsing |
 | `GetInitialState` | Worker does NOT register via fsmv2.RegisterInitialState (e.g., push, pull) |
 
-Never override `GetDependenciesAny`. The collector asserts the bound deps against
-`deps.FrameworkAccessor` and attaches framework metrics and action history only on
-success; `struct{}{}` and `nil` both fail that assertion, so returning nil changes
+Never override `GetDependenciesAny`. Before attaching framework metrics and action
+history, the collector checks whether the bound deps can carry framework state;
+`struct{}{}` and `nil` both fail that check identically, so returning nil changes
 nothing. A worker that needs framework telemetry must bind a `TDeps` embedding
 `*deps.BaseDependencies`; leaving `TDeps` as `struct{}` means no telemetry, whatever
 the accessor returns.
--- a/umh-core/pkg/fsmv2/CLAUDE.md
+++ b/umh-core/pkg/fsmv2/CLAUDE.md
@@ -243,7 +243,13 @@ func (w *MyWorker) GetDependencies() *MyDependencies {
 | `CollectObservedState` | Always — worker-specific snapshot |
 | `DeriveDesiredState` | Custom children specs (application, communicator) or non-standard config parsing |
 | `GetInitialState` | Worker does NOT register via fsmv2.RegisterInitialState (e.g., push, pull) |
-| `GetDependenciesAny() any { return nil }` | No-deps worker. `WorkerBase[..., struct{}]` boxes `struct{}{}` into `any`, which is non-nil and silently skips framework metrics injection. The override returns a true nil to prevent that. |
+
+Never override `GetDependenciesAny`. The collector asserts the bound deps against
+`deps.FrameworkAccessor` and attaches framework metrics and action history only on
+success; `struct{}{}` and `nil` both fail that assertion, so returning nil changes
+nothing. A worker that needs framework telemetry must bind a `TDeps` embedding
+`*deps.BaseDependencies`; leaving `TDeps` as `struct{}` means no telemetry, whatever
+the accessor returns.
 
 ## Worker Registration
 
--- a/umh-core/pkg/fsmv2/CLAUDE.md
+++ b/umh-core/pkg/fsmv2/CLAUDE.md
@@ -244,12 +244,16 @@ func (w *MyWorker) GetDependencies() *MyDependencies {
 | `DeriveDesiredState` | Custom children specs (application, communicator) or non-standard config parsing |
 | `GetInitialState` | Worker does NOT register via fsmv2.RegisterInitialState (e.g., push, pull) |
 
-Never override `GetDependenciesAny`. Before attaching framework metrics and action
-history, the collector checks whether the bound deps can carry framework state;
-`struct{}{}` and `nil` both fail that check identically, so returning nil changes
-nothing. A worker that needs framework telemetry must bind a `TDeps` embedding
-`*deps.BaseDependencies`; leaving `TDeps` as `struct{}` means no telemetry, whatever
-the accessor returns.
+Never override `GetDependenciesAny`. **Framework telemetry is decided by the type you
+bind**: a `TDeps` embedding `*deps.BaseDependencies` receives framework metrics and
+action history, a `TDeps` of `struct{}` receives neither, whatever the accessor
+returns.
+
+Neither delivery path checks the bound value for nil. Both type-assert on it, and they
+assert on different method sets: the collector wants the `BaseDependencies` getters
+(`GetFrameworkState`, `GetActionHistory`, `MetricsRecorder`), while the supervisor
+wants `SetFrameworkState`. `struct{}{}` and `nil` satisfy neither, so overriding the
+accessor to return nil changes nothing.
 
 ## Worker Registration
 
--- a/umh-core/pkg/fsmv2/historian/timescale.go
+++ b/umh-core/pkg/fsmv2/historian/timescale.go
@@ -59,7 +59,6 @@ import (
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsmv2/deps"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsmv2/simple"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsmv2/workers/configworker/dynamicchildren"
-	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/logger"
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/models"
 )
 
@@ -112,9 +111,10 @@ type TimescaleStatus struct {
 	Reachable bool `json:"reachable"`
 }
 
-// Deps carries the poll dependencies. The pool holder lazily builds and caches a
-// pgx pool keyed by DSN, so Poll reuses pooled connections instead of dialing
-// every tick.
+// Deps carries the poll dependencies: the framework's BaseDependencies for this
+// worker instance (the logger Poll writes to) plus the pool holder. The holder
+// lazily builds and caches a pgx pool keyed by DSN, so Poll reuses pooled
+// connections instead of dialing every tick.
 //
 // The holder is process-wide, not per worker instance: nothing in the worker
 // framework closes what a worker holds, so a per-instance pool would leak its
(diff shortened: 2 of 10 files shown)


umh-core/pkg/fsmv2/historian/timescale.go
```diff
@@ -112,12 +111,48 @@ type TimescaleStatus struct {
 	Reachable bool `json:"reachable"`
 }
 
-// Deps carries the poll dependencies, shared across ticks. The pool holder
-// lazily builds and caches a pgx pool keyed by DSN, so Poll reuses pooled
-// connections instead of dialing every tick.
+// Deps carries the poll dependencies: this instance's BaseDependencies (the
```
JeremyTheocharis: my feeling: this is too long of a comment

umh-core/pkg/fsmv2/historian/timescale.go
```diff
@@ -112,12 +111,48 @@ type TimescaleStatus struct {
...
+// material at the sslrootcert and sslcert paths is never re-read.
 type Deps struct {
-	Logger deps.FSMLogger
-	pool   *poolHolder
+	*deps.BaseDependencies
+
+	pool *poolHolder
+}
+
+// sharedPool is the process-wide holder every worker instance polls through. It
```
JeremyTheocharis: does historian actually need it? i would challenge this

JeremyTheocharis: lets maybe not have a shared pool?

JeremyTheocharis:
@coderabbitai full review

umh-core/pkg/fsmv2/historian/timescale.go
JeremyTheocharis: Fixed in 27619d303, and the diagnosis turned out to be placement rather than length.

An independent pass over the 18 lines found nothing deletable: 4 lines define the fields, 5 state the singleton invariant, 7 disclose two defects that are both really there. What was wrong is that two of the three paragraphs were not about `Deps` at all. They described the *sharing* design, which belongs on `sharedPool`. Moving them there drops the `Deps` comment to 2 lines without deleting a word, and removes a forward reference on the way, since `Deps` named `sharedPool` 29 lines before it was declared.

Two statements did go, because they were false:

- *"only Close stops the health-check goroutine, and nothing calls it"* — `h.pool.Close()` is called on a DSN change. The test in this file has the true version, *"nothing in the **framework** closes it"*. The godoc had lost the qualifier that made it true.
- *"Only one historian worker **can** run on that holder"* — nothing enforces that, and the next two lines went on to describe two instances thrashing. What actually constrains it is the hardcoded `Ref`, one writer (`syncHistorian` upserting that single `Ref`), and a single `Historian` config pointer. It now says "is a singleton today" and states the consequence instead of claiming a guarantee.

Both remaining disclosures now cite a ticket (ENG-5375 for the missing teardown hook, ENG-5593 for cert rotation, the latter moved onto `poolHolder.get` where the DSN is actually parsed).

Worth flagging for calibration: the file is still 131 comment lines of 314. This change removed 2. If the objection was the file's overall comment volume rather than this one comment, it is not addressed and I would treat that as a separate pass.

umh-core/pkg/fsmv2/historian/timescale.go
JeremyTheocharis: Leaving this open, because it is a design question and I do not think it should be settled by editing the comment. Here is what I can establish, so you can decide.

**Does historian need a shared pool? Not for correctness today.** Historian is a singleton: one hardcoded `Ref` (`InstanceName = "timescale"`), one writer (`syncHistorian` upserting that single `Ref`), and a single `Historian` config pointer. With exactly one instance, "shared" and "per-instance" are behaviourally identical. So the process-global buys nothing today.

**What it does buy is avoiding a leak caused by a missing framework hook.** `simple.MonitorSpec` has no teardown hook, and `fsmv2.GracefulShutdowner` is declared at `api.go:443` and invoked nowhere in production (verified across the whole module: declaration, one doc mention, no type assertion, no dispatch). So a worker is never called at a point where it could close a pool it owns. With a per-instance holder, every despawn that followed a poll would orphan a pgxpool health-check goroutine with nothing able to close it. With the single holder, `get` closes the previous pool on a DSN change, which bounds the process to one pool.

Note the shared pool does not eliminate the leak, it bounds it. Removing the historian config block still despawns the child without closing the pool, so the goroutine and up to `maxConns` (2) sessions survive until the DSN changes or the process exits.

**Three ways out, as I see them:**

1. **Keep it** (status quo). Bounded leak, one process-global, zero work.
2. **Per-instance holder now.** Removes the global, and orphans a goroutine on every despawn-after-poll. Strictly worse today.
3. **Per-instance holder plus a teardown hook** (ENG-5375). The only genuinely clean end state, and it needs a framework change, not a historian change.

**My recommendation: 1 now, 3 as the real fix.** This PR is about `NewDeps` being per-instance; the pool lifetime is a separate concern and option 3 is a framework change that wants its own PR. ENG-5375 already carries the teardown hook and is the natural home.

One thing I would push back on: if we do go per-instance without a teardown hook, the comment defending the global should not simply be deleted, since the leak it describes becomes real rather than hypothetical.

umh-core/pkg/fsmv2/simple/worker_test.go
```diff
@@ -274,4 +504,67 @@ var _ = Describe("Register", func() {
...
+				built = append(built, w)
+
+				return w
+			},
+			Poll: func(_ context.Context, w *window, _ probeConfig) (probeStatus, error) {
+				w.polls++
+
+				return probeStatus{}, nil
+			},
+		})
```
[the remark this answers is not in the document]
JeremyTheocharis: Declining this one, because the hazard is not reachable in this suite.

The mechanism is real: `factory.go:53` returns `"worker type already registered"` and `register.Worker` panics on it. What is missing is a path that produces a duplicate.

- Every `Register` call in this file uses a distinct `WorkerType` string (`simpleworker_nointerval`, `simpleworker_nopoll`, `simpleworker_register_factory`, and so on), so no two specs collide.
- Each spec runs once per process, so a registration cannot repeat within a run.
- A same-process re-run is the one thing that would trigger it, and **Ginkgo refuses it outright**. `go test -count=2` on this package fails with *"Ginkgo does not support using go test -count to rerun suites. Only -count=1 is allowed."* The `ginkgo -repeat=N` alternative forks a fresh process per attempt, so the registry starts empty each time.

So the `AfterEach` would be a guard that cannot fail. The existing `fsmv2.ResetInitialStateRegistry()` is there because that registry *is* keyed such that specs collide; the factory registry is not.

For the record on the other half of the risk: `factory.ResetRegistry()` clears the process-wide registry including anything registered at `init()`, which it cannot restore. That happens to be harmless here (no worker packages are linked into this test binary, confirmed via `go list -deps`), but it is the reason this is not a free change in general.

Happy to add it if you would rather have the belt regardless of the braces, since it is two lines.

umh-core/pkg/fsmv2/CLAUDE.md
```diff
@@ -243,7 +243,13 @@ func (w *MyWorker) GetDependencies() *MyDependencies {
 | `CollectObservedState` | Always — worker-specific snapshot |
 | `DeriveDesiredState` | Custom children specs (application, communicator) or non-standard config parsing |
 | `GetInitialState` | Worker does NOT register via fsmv2.RegisterInitialState (e.g., push, pull) |
-| `GetDependenciesAny() any { return nil }` | No-deps worker. `WorkerBase[..., struct{}]` boxes `struct{}{}` into `any`, which is non-nil and silently skips framework metrics injection. The override returns a true nil to prevent that. |
+
+Never override `GetDependenciesAny`. Before attaching framework metrics and action
```
Aaron99B: I do not fully get the description here: Does that mean we always have to attach framework metrics and action history, or does it mean we just have to attach these if we think about overwriting GetDependenciesAny?

proposal for rewrite:
Never override GetDependenciesAny. The collector decides whether to deliver
framework metrics and action history by checking the bound deps for a
SetFrameworkState method — not by checking for nil. struct{}{} and nil
both lack that method, so returning nil changes nothing. Framework telemetry
is decided by the type you bind: a TDeps embedding *deps.BaseDependencies
gets it, a TDeps of struct{} does not, whatever the accessor returns.

I think this is the most important part to state: Framework telemetry
is decided by the type you bind: a TDeps embedding *deps.BaseDependencies
gets it, a TDeps of struct{} does not, whatever the accessor returns.

Aaron99B (approved):
Little comment for the Claude.md, then good to go

umh-core/pkg/fsmv2/historian/timescale.go
JeremyTheocharis: Decided: keep the shared pool here, fix it properly later. Recording the reasoning so the next reader does not have to re-derive it.

Historian runs exactly one instance today, so shared and per-instance behave identically. What the single holder buys is not sharing, it is bounding a leak: `simple.MonitorSpec` has no teardown hook and `fsmv2.GracefulShutdowner` is declared at `api.go:443` and invoked nowhere in production, so a worker is never called at a point where it could close a pool it owns. With one holder, `get` closes the previous pool on a DSN change, which bounds the process to a single pool. With a holder per instance, every despawn that followed a poll would orphan a pgxpool health-check goroutine and nothing could ever close it.

So going per-instance now would be strictly worse than today, not better. The clean end state is per-instance plus a teardown hook, and that is a framework change rather than a historian one: **ENG-5375** already carries the hook.

Two things this does not fix, stated plainly rather than left implied. The single holder bounds the leak, it does not remove it: removing the historian config block still despawns the child without closing the pool, so the goroutine and up to `maxConns` (2) sessions survive until the DSN changes or the process exits. And the singleton is a fact about today's config, not an invariant anything enforces, which is why the comment now says "is a singleton today" rather than claiming a guarantee.

Resolving on that basis. If ENG-5375 lands and the per-instance holder goes in, the comment defending the global should go with it rather than being left to rot.

umh-core/pkg/fsmv2/CLAUDE.md
JeremyTheocharis: Taking your rewrite, it is clearer than what was there. You put your finger on the actual problem: "Before attaching framework metrics and action history, the collector checks..." reads as though attaching is the default and the check is a formality, so it never answers your question. Leading with **framework telemetry is decided by the type you bind** does, and that sentence is now the first thing in the paragraph.

One correction, because this is the file that teaches the rule and it should be exactly right rather than nearly right. There is no single `SetFrameworkState` check. The two delivery paths assert on **different method sets**:

- the collector asserts against the `BaseDependencies` getters, `GetFrameworkState` / `GetActionHistory` / `MetricsRecorder` (`collector.go`, `baseDepsAccessor`)
- the supervisor asserts against `SetFrameworkState` (`api.go`)

`struct{}{}` and `nil` satisfy neither, so your conclusion holds either way; only the mechanism needed widening. Both are now named.

I also dropped "fail that check" in favour of your framing that neither path tests for nil, since that is precisely why overriding the accessor to return nil accomplishes nothing.

In `9355e5dba`.

PR #2661, opened 2026-07-20
docs: update getting-started bridge creation flow + downsampler note

robertschwarz-umh:
## TL;DR

- update screenshots of `1-getting-started` section
- add "Choose How to Create the Bridge" section
- more descriptive image alt text
- add section that explains downsampling

---

> `skip-changelog-guard` is used for this PR, because docs updates do not merit a changelog update.

## What

Updates `umh-core/docs/getting-started/1-connect-data.md` (Step 2: Connect Your First Data Source) to match the current UI, plus refreshed screenshots.

- Document the bridge-creation flow the UI now enforces: the "Create a New Bridge" choice (From Scratch / From Existing Bridge) and template selection. New users previously hit a template picker with ~83 options and got stuck.
- Guide the user to the `generate` template ("Generic via Generate") so the tutorial works without a real PLC.
- Correct the General-tab description and screenshot (tabs are now General / Read Flow / Write Flow).
- Add a note in the Topic Browser section explaining why a repeated `hello world` shows only one row: benthos-umh appends a downsampler that drops byte-for-byte repeats. Links https://docs.umh.app/benthos-umh/processing/downsampler.

## Why

Onboarding feedback from a new (IT-background) user walking the live flow: the page no longer matched the UI at two points, and the single Topic Browser row read as a bug rather than intended downsampling.

## Verification

Downsampler behavior confirmed against source: `umh-core/pkg/service/protocolconverter/runtime_config/runtime_config.go:344-348` auto-appends the downsampler for Time Series data; default dead-band config suppresses byte-for-byte repeats (`benthos-umh/docs/processing/downsampler.md`).

## Notes

- Docs-only change; no code paths touched. Hence `skip-changelog-guard`.
- `*.BACKUP.old.png` files are the pre-update screenshots kept alongside the new ones.

Ref: ENG-5396

PR #2677, opened 2026-08-02
[ENG-5469] Update bridges documentation

rpaland:
## What
The Docs page on Bridges was updated in this PR to reflect the current process for creating a bridge.


umh-core/docs/usage/data-flows/bridges.md
```diff
@@ -26,28 +29,22 @@ Go to **Data Flows** and click **Add Bridge**:
...
+### Step 3: Select a Template
 
-Click the **Read** tab and select your protocol:
+![Select a Template](./images/bridges-template-selection.png)
 
-![Protocol Selection](./images/bridges-read-no-dfc-protocol-selection.png)
+Templates carry a working protocol configuration. Search by name, or narrow the list with the **Vendor** and **Protocol** filters.
 
-We support **50+ industrial protocols** through [Benthos-UMH](https://docs.umh.app/benthos-umh/input) plus everything from [Redpanda Connect](https://docs.redpanda.com/redpanda-connect/components/inputs/about/):
+The list currently holds **84 templates**. Behind them we support **50+ industrial protocols** through [Benthos-UMH](https://docs.umh.app/benthos-umh/input) plus everything from [Redpanda Connect](https://docs.redpanda.com/redpanda-connect/components/inputs/about/):
```
robertschwarz-umh: coderabbit listed this as a nitpick, but i think we should drop exact counts from our docs. otherwise we have to update this number every time we change the count of templates.

"50+ protocols" is nice, but i'd drop "holds 84 templates".

umh-core/docs/usage/data-flows/bridges.md
```diff
@@ -56,13 +53,40 @@ Common industrial protocols:
...
+- **IP Address**: Device IP (becomes `{{ .IP }}` variable)
+- **Port**: Connection port (becomes `{{ .PORT }}` variable)
+
+The protocol itself needs no selection here. It comes from the template you picked in Step 3, and the **Read** tab opens on that protocol's settings.
+
+### Step 5: Configure Protocol Settings
 
 Each protocol has specific settings. For example, S7:
 
 ![Siemens S7 Configuration](./images/bridges-read-no-dfc-data-type-selector-for-siemens.png)
```
robertschwarz-umh, line 79: This screenshot shows the old UI. it looks quite different now.
But rather than updating the screenshot, I think we can just remove this section (unless we're planning to document the diff between relational vs time-series data any time soon

Image

robertschwarz-umh (changes requested):
Just a random question: Is ENG-5299 a typo? 🤔 I don't see a relationship between this PR's docs updates and the description in the ticket.

**Two more things:**
- the new UI has a toggle to turn Read/Write flows on or off. Is this toggle worth explaining in the docs?
- "from scratch" tile when creating a new bridge: maybe we should rename this long-term. It's not really from scratch, right? It's creating a new bridge from an existing one. 

Also, comparing my PR to yours [(see here)](https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/2661/changes), there's a ton of overlap.

My preference would be to "fuse" both pages into a single `bridges.md` page, and then just link to it in the `getting-started` section. Otherwise, I see a risk of 1 page staying up-to-date, while we forget to update the other one.

rpaland:
@robertschwarz-umh 

> "Just a random question: Is [ENG-5299](https://linear.app/united-manufacturing-hub/issue/ENG-5299) a typo? 🤔"

Yes was by accident. Updated the PR description.

> "from scratch" tile when creating a new bridge: maybe we should rename this long-term. It's not really from scratch, right? It's creating a new bridge from an existing one.

No I think you're mixing things up here. There is the "from scratch" variant, which is technically a "create from template" way, and there is the "from existing bridge", which is what you actually stated. So not really from scratch, but also not from existing bridge.

rpaland:
@robertschwarz-umh 

> "My preference would be to "fuse" both pages into a single bridges.md page, and then just link to it in the getting-started"

Yes good idea!

umh-core/docs/usage/data-flows/bridges.md
rpaland: I'll rephrase it. Good point!

umh-core/docs/usage/data-flows/bridges.md
rpaland: Yes I left the old screenshot in there, because that's the current look with experimental features disabled, and from my regard, what we should currently state. Maybe in an extra "experimental" section or so - happy to discuss though. Planned on updating / removing it, once this is set to enabled by default.

robertschwarz-umh: Makes sense, lets wait until the actual ff becomes the standard then!

robertschwarz-umh (approved):

umh-core/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Improvements
+
+- The [Bridges](https://docs.umh.app/usage/data-flows/bridges) documentation now matches the current creation flow, from choosing a template or an existing bridge as the starting point through the address mapping table, where a custom data contract turns Virtual Path and Tag Name into a single dropdown of the fields its data model defines
```
JeremyTheocharis: we dont document in the changelog docs changes

JeremyTheocharis (changes requested):

JeremyTheocharis (approved):

PR #2687, opened 2026-08-06
benthos_monitor on fsmv2 (ENG-5511)

JeremyTheocharis:
> Stacked on united-manufacturing-hub/united-manufacturing-hub#2698. That PR makes the FSMv2 runtime unconditional and deletes the legacy HTTP communicator, so review it first — this branch is based on it, not on staging.

## Problem

Every benthos instance runs a second S6 service that exists only to scrape that instance's metrics. Its run-script forks curl, gzip and xxd about fourteen times a second and writes through a compressed log, which umh-core then re-reads and re-parses ten times a second — while the data behind it changes once a second. On a ten-bridge container that scrape is about 40% of all CPU the container burns, over half of it kernel time spent forking.

## What we're shipping

* **A** `benthos_monitor` **worker on fsmv2's** `simple` **package** scrapes the four endpoints in-process once a second.
* **One feature flag,** `USE_FSMV2_BENTHOS_MONITOR`**, default off.**
* **Two FSMv1 throughput bugs corrected on the new path**: a loaded box no longer overstates throughput, and a benthos restart no longer publishes a cumulative counter as a rate.
* Long comments on the adapter.go, as there is legit a lot of mental complexity around it that is difficult to put into code. 

## What we're NOT shipping

* **The flag stays off by default.** Two follow-ups planned: enable it by default, then delete the FSMv1 path.
* No change to the field shapes or units any FSMv1 reader consumes.
* No seamless flag flip — the flag is read at startup, so switching means a restart.
* Removal-pending detection stays as strong as nmap's, weaker than FSMv1's. Blocked on [ENG-5107](https://linear.app/united-manufacturing-hub/issue/ENG-5107/cse-tombstone-despawned-workers-deleted-at-so-reads-stop-reporting).
* **No tests added to** `fsm/protocolconverter` **or** `fsm/streamprocessor` — both have none, which is why the degraded-bridge defect was invisible to the suite.

## Confidence

Tested by running an experiment:

```
  Hypothesis
    The per-bridge metrics scrape runs as a separate helper process per bridge
    that forks several short-lived processes a second. Moving it in-process
    should remove a fixed CPU cost per bridge, visible mainly as kernel time.

  Design
    2x3 factorial: flag {off, on} x bridges {0, 5, 10}, six cells.
    Both arms ran the SAME image (sha-984ae5e), so no build difference can
    explain the result. One fresh container per arm, --cpus 2 --memory 2g,
    identical config and identical dummy backend credentials in both arms, so
    USE_FSMV2_BENTHOS_MONITOR is the only difference. Bridge counts were reached
    by rewriting the whole config file, never by patching it.

  Measurement
    Three consecutive 60s windows per cell after a settle period; each cell
    reports the median of its three windows. Read from the container cgroup:
    total/user/system CPU, fork count, process count, memory, PSI.
    Slope endpoints were pinned at N=0 and N=10 BEFORE the run, so no endpoint
    was chosen after seeing the numbers.

  Validity controls
    - Liveness: every cell asserts all N bridges reached Active, so a CPU
      saving cannot come from work silently failing to happen. Held at every N.
    - Monitor count: flag off must show N monitor processes, flag on must show
      0. Held at every N.
    - Fork rate: flag off must fork 12-16 times/second/bridge, the documented
      behaviour of the old scrape script. Measured 13.9.
    - Ambient state: host load average and the count and names of any foreign
      containers recorded at start and end. Zero foreign containers both times.

  Result
    Per-bridge CPU saving 0.0202 cores, stable across both sub-intervals
    (0.0200 for N=0-5, 0.0204 for N=5-10) even though total CPU per bridge is
    strongly sub-linear in both arms. That stability is what licenses quoting a
    per-bridge figure instead of a percentage of total.
    At 10 bridges: 0.445 -> 0.242 cores, 45.5% lower.
    Mechanism confirmed where it should appear: marginal kernel time per bridge
    fell 91.4%, and forks fell from 8,380 to 62 per minute.

  Deliberately NOT established
    - No claim about stall pressure. PSI read 0.00 in all 18 windows in both
      arms; at this scale on this host nothing was ever pressured.
    - Memory is unresolved. It read ~9% higher with the flag on, but was still
      climbing in all six cells and was 19% higher at N=0 where the flag creates
      no monitors. Not a usable number; it needs its own experiment.
    - Bridges carried almost no message traffic. The absolute saving should hold
      under load because the scrape is a fixed 1 Hz timer, but that was not
      tested; the percentage of total CPU will be lower on a busy instance.
    - Measured on aarch64 under Docker Desktop, not an x86 server.
    - The image predates this branch's last three commits. The only shipped
      behaviour among them is the IsRunning fix, which cannot change a fork
      count.
```


JeremyTheocharis:
@coderabbitai full review

umh-core/pkg/fsmv2/benthos_monitor/manager.go
```diff
@@ -0,0 +1,285 @@
...
+	if err != nil {
+		return nil, 0, err
+	}
+	// Defensive: http.Client.Do may return a nil resp alongside a non-nil error
+	// on some paths; the error check above normally covers it, but guard the
+	// dereferences that follow so a nil resp can never slip through.
+	if resp == nil {
+		return nil, 0, fmt.Errorf("GET %s: nil response", url)
+	}
+	defer resp.Body.Close()
```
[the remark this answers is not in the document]
JeremyTheocharis: Fixed in 2ae602145 — `defer func() { _ = resp.Body.Close() }()`. No behaviour change: the body is fully drained by `io.ReadAll` above, so a Close error has nothing left to report.

umh-core/pkg/fsmv2/benthos_monitor/throughput_red_test.go
```diff
@@ -0,0 +1,266 @@
...
+		switch r.URL.Path {
+		case "/ping":
+			_, _ = w.Write([]byte("pong"))
+		case "/ready":
+			_, _ = w.Write([]byte(`{"error":""}`))
+		case "/version":
+			_, _ = w.Write([]byte(`{"version":"1.2.3"}`))
+		case "/metrics":
+			fmt.Fprintf(w, "input_received{label=\"\",path=\"root.input\"} %d\n", inputCounter)
+			fmt.Fprintf(w, "output_sent{label=\"\",path=\"root.output\"} %d\n", outputCounter)
```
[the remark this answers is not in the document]
JeremyTheocharis: Fixed in 2ae602145, all four call sites (both handlers). The consistency point was the useful part of this one: the same switch already writes `_, _ = w.Write(...)` for /ping, /ready and /version and then dropped the convention for /metrics two lines later, so the file disagreed with itself about whether these returns matter.

umh-core/pkg/service/benthos/benthos_backend_test.go
```diff
@@ -0,0 +1,171 @@
...
+		// test file's own import — is what must have registered the worker. If a
+		// refactor removes the production constructor call (or the import), FF-on
+		// ships nothing while the CPU rig counts a fake win; this gate goes red.
+		registered := false
+		for _, wt := range factory.ListRegisteredTypes() {
+			if wt == fsmv2benthosmonitor.WorkerType {
+				registered = true
+				break
+			}
+		}
```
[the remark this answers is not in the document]
JeremyTheocharis: Fixed in 2ae602145.

One correction to the stated reason, since it applies to all three of these: `nlreturn` and `errcheck` are enabled in `umh-core/.golangci.yaml`, but **golangci-lint runs in no workflow in this repo** — `grep -rl golangci .github/workflows/` returns nothing, while `nilaway` appears in two — and the local hooks are deliberately unplugged. So "golangci-lint reports this as an error" is true of the config and false of CI; nothing would have caught these. Taken anyway because they are cheap and in files this PR owns.

umh-core/pkg/fsmv2/benthos_monitor/manager.go
```diff
@@ -0,0 +1,293 @@
...
+	Ready     bool
+	Version   string
+}
+
+// Poll scrapes the configured benthos monitor's /ping, /ready, /version, and
+// /metrics endpoints once. The client comes from d (a nil deps falls back to the
+// default client). All requests carry ctx, so cancellation surfaces as a request
+// error.
+func Poll(ctx context.Context, d *benthosMonitorDeps, cfg config.BenthosMonitorConfig) (BenthosMonitorStatus, error) {
+	client := http.DefaultClient
```
JeremyTheocharis: unsure, shouldnt this fail hard instead? looks like a programming error if we have this?

umh-core/pkg/fsmv2/benthos_monitor/manager.go
```diff
@@ -0,0 +1,293 @@
...
+	if err := json.Unmarshal(versionBody, &v); err != nil {
+		return status, fmt.Errorf("decode /version: %w", err)
+	}
+	status.Version = v.Version
+
+	metricsBody, _, err := get(ctx, client, base+"/metrics")
+	if err != nil {
+		return status, fmt.Errorf("scrape /metrics: %w", err)
+	}
+	// The fsmv1 parser is the only parser: one implementation cannot disagree
```
JeremyTheocharis: the comments here is too long, just say "re-use the fsmv1 parser for full compatbility"

umh-core/pkg/fsmv2/benthos_monitor/manager.go
```diff
@@ -0,0 +1,293 @@
...
+	client := http.DefaultClient
+	if d != nil && d.client != nil {
+		client = d.client
+	}
+
+	base := fmt.Sprintf("http://localhost:%d", cfg.MetricsPort)
+	status := BenthosMonitorStatus{ScrapedAt: time.Now()}
+
+	_, _, err := get(ctx, client, base+"/ping")
+	if err == nil {
```
JeremyTheocharis, line 129: if this fails, do we even need to continue here? i dont think so

umh-core/pkg/fsmv2/benthos_monitor/manager.go
```diff
@@ -0,0 +1,293 @@
...
+	status.BenthosMetrics = m
+
+	// The window lives in d and survives across polls, which is what makes a rate
+	// computable from per-poll counter snapshots; a nil d has no window, so Input,
+	// Output and IsActive stay zero.
+	//
+	// Add is passed the scrape port because it keys the window on that port: an
+	// in-place config update can re-point this child at a new MetricsPort with no
+	// worker restart, and throughputWindow.port explains the wipe that follows.
+	// Its counter arguments are the totals summed across every leaf path, which is
```
JeremyTheocharis:
```
Its counter arguments are the totals summed across every leaf path, which is
	// what the counters mean for throughput: a switch or broker emits one series
	// per leaf with no top-level aggregate, so the sum is the pipeline's traffic.
```

→ feels overhwleming

umh-core/pkg/fsmv2/benthos_monitor/throughput_window.go
```diff
@@ -0,0 +1,181 @@
...
+	newest := w.samples[0]
+	for _, s := range w.samples[1:] {
+		if s.at.After(newest.at) {
+			newest = s
+		}
+	}
+	return newest
+}
+
+// inputRate returns the input rate in messages per second over the in-window
```
JeremyTheocharis, line 105: these comment blocks seem to large, maybe it could make sense to move some of this inline into the function?

umh-core/pkg/fsmv2/benthos_monitor/throughput_window.go
```diff
@@ -0,0 +1,181 @@
...
+	kept := w.samples[:0]
+	for _, s := range w.samples {
+		if !s.at.Before(cutoff) {
+			kept = append(kept, s)
+		}
+	}
+	w.samples = kept
+}
+
+// wipeOnRestart reports whether the process restarted between prev and the new
```
JeremyTheocharis: maybe it makes sense to move some of this inline? we coiuld split up `at.After(prev.at)` into a variable and add a comment explaining it and then the same also for `input < prev.input` and then combine it using the && and  explain it there as well. might make it easier to follow?

umh-core/pkg/service/benthos/benthos.go
```diff
@@ -66,6 +68,13 @@ const (
 	OPCDebugEnvVar = "OPC_DEBUG"
 )
 
+// envUseFsmv2BenthosMonitor names the environment variable that selects the
```
JeremyTheocharis: feels overkill the comment, just make it one or max 2 lines

umh-core/pkg/service/benthos/benthos.go
```diff
@@ -277,18 +317,47 @@ func WithS6Manager(s6Manager *s6fsm.S6Manager) BenthosServiceOption {
...
+			log.Infof("%s is on: this benthos instance is monitored by the fsmv2 benthos_monitor worker (in-process scrape, no S6 monitor service)", envUseFsmv2BenthosMonitor)
+			service.benthosMonitorManager = fsmv2benthosmonitor.NewFsmv2BenthosMonitorManager(
+				// Match the fsmv1 backend's monitor log prefix so an operator
+				// grepping logs sees the same labels across a flag flip.
+				fmt.Sprintf("%s_%s", logger.ComponentBenthosMonitorManager, benthosName),
+			)
+		} else {
+			// The missing counterpart to the Infof above is deliberate, not an
+			// oversight: this is the branch every container with
+			// USE_FSMV2_BENTHOS_MONITOR unset takes, and the flag-off default
```
JeremyTheocharis: do we need to restate again this FF? seems to be rather intuitive because of useFsmv2?

umh-core/pkg/fsmv2/benthos_monitor/manager.go
JeremyTheocharis: Fixed in `6840124bc`. You were right, and it turned out to be reachable rather than theoretical: `simpleWorker.pollDeps` discards the comma-ok on its type assertion, so deps that were never bound arrive here as a nil pointer. The old fallback would then scrape *successfully* while dropping the cross-poll window — publishing zero `Input`/`Output`/`IsActive` on a monitor that otherwise looked healthy.

`Poll` now returns an error before any request goes out, and the framework turns that into a degraded worker carrying the reason.

Four accommodations that existed only for the nil case went with it: the `http.DefaultClient` fallback, the `if d != nil` guard around the throughput block, and two comments describing the nil path.

Pinned by `TestPollRefusesToScrapeWithUnboundDeps`. Worth recording that its first version was vacuous: pointing `Poll` at a closed port made the "returns an error" assertion pass for the wrong reason, since `http.DefaultClient` fails on a closed port too. It now scrapes a live healthy server, which is the only configuration that discriminates, plus a control subtest proving the guard does not reject bound deps.

umh-core/pkg/fsmv2/benthos_monitor/manager.go
JeremyTheocharis: Fixed in `d38400a99` — now exactly `// Re-use the fsmv1 parser for full compatibility.`

The rest argued against writing a second parser, which nobody proposed. The framework's degraded-on-parse-error behaviour is its own documented contract, so that isn't restated at the call site either.

umh-core/pkg/fsmv2/benthos_monitor/manager.go
JeremyTheocharis: Answering rather than changing this one, and leaving the thread **open** in case you disagree.

I'd keep polling after a `/ping` failure. `PingAlive=false` alongside a working `/metrics` is a real state, not an error: `mapObserved` maps `PingAlive`→`IsLive` and `Ready`→`IsReady` as separate fields and the FSM reads them separately. An early return collapses "process up, pipeline broken" into "process down", which is exactly the distinction bridge health uses to tell a broken pipeline from a dead container. fsmv1 reported these endpoints independently for the same reason.

Cost of continuing is three more HTTP calls to localhost against a dead port, which fail fast.

If you'd rather have the early return, say so and I'll do it as its own commit — it changes behaviour, so it shouldn't ride along with comment work.

umh-core/pkg/fsmv2/benthos_monitor/manager.go
JeremyTheocharis: Fixed in `d38400a99`.

The summed-leaf-paths sentence described what `InputReceivedTotal` and `OutputSentTotal` sum, which is a property of those methods in `benthosmonitorservice` rather than of this call site — so it's cut here rather than reworded into something shorter and vaguer. The port-keying rationale stays, tightened, because that one really is about this call.

The nil-window sentence in the same block disappeared separately in `6840124bc`, since that case can no longer happen.

umh-core/pkg/fsmv2/benthos_monitor/throughput_window.go
JeremyTheocharis: Fixed in `a7b4c6257` — each explanation now sits on the guard it explains instead of in the doc ten lines above, and `outputRate` points at `inputRate` by name rather than repeating the prose.

One honest note on the outcome: the file's comment count went 68 → 69, so this is **relocation, not reduction**. What improved is that you no longer have to hold the guards in your head while reading about them.

umh-core/pkg/fsmv2/benthos_monitor/throughput_window.go
JeremyTheocharis: Done in `a7b4c6257`, as you suggested — `isNewer` and `counterDropped`, each with the short reason it exists, combined at the return with why both are required. The ten-line doc shrank to one sentence saying what the function decides.

Both load-bearing facts moved into the named conditions rather than disappearing: that `input_received` resets to 0 on a restart (hence a drop is the signal), and that the output counter is deliberately *not* consulted, because a restart can leave output at ~0 and a both-counters detector would never wipe there.

Verified behaviour-preserving: no test file touched, suite unchanged, and neutering `counterDropped` reddens `TestThroughputWindowWipesOnCounterReset`, `TestThroughputWindowWipesOnOneCounterZeroRestart` and `TestThroughputWindowRecoveredAboveBaseline` — so the named condition is load-bearing, not decorative.

umh-core/pkg/service/benthos/benthos.go
JeremyTheocharis: Fixed in `d38400a99` — two lines, saying what the variable selects.

The list of accepted boolean spellings (`on/off`, `yes/no`, `1/0`, …) belongs to `env.GetAsBool`, so it's dropped here rather than duplicated.

umh-core/pkg/service/benthos/benthos.go
JeremyTheocharis: Fixed in `d38400a99` — the restatement is gone.

One fact was worth keeping and is now two lines: the missing `Infof` counterpart is deliberate, because this is the branch a container with the flag unset takes, the flag-off default must keep its pre-flag log output, and no test pins that silence.

umh-core/pkg/fsmv2/benthos_monitor/throughput_window.go
```diff
@@ -0,0 +1,186 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fsmv2benthosmonitor
+
+import "time"
+
+// windowSpan is the width of the throughput window in seconds. Throughput is
```
JeremyTheocharis: this comment feels too long

umh-core/pkg/fsmv2/benthos_monitor/throughput_window.go
```diff
@@ -0,0 +1,186 @@
...
+	at     time.Time
+	input  int
+	output int
+}
+
+// throughputWindow holds a by-time window of counter samples. Its zero value is
+// a valid empty window: every method is safe on an empty window, so there is no
+// constructor; declare it by value.
+type throughputWindow struct {
+	// port is the scrape port this window's samples belong to. Add wipes the
```
JeremyTheocharis, line 37: i feel like

```
 so it
	// never subtracts one poll's counter from a different endpoint's. Add
	// compares this field only once the window holds a sample (its n > 0 guard),
	// so the zero value it starts at is never read as a port. 
```

can be cut?

umh-core/pkg/fsmv2/benthos_monitor/throughput_window.go
```diff
@@ -0,0 +1,186 @@
...
+	// compares this field only once the window holds a sample (its n > 0 guard),
+	// so the zero value it starts at is never read as a port.
+	port    int
+	samples []throughputSample
+}
+
+// Add records a sample at the given observed time with the given port and input
+// and output counter values, then drops any sample older than windowSpan of the
+// newest. It wipes the window first when the port changed or when wipeOnRestart
+// reports true. In production the only caller is Poll (manager.go), one call per
```
JeremyTheocharis, line 47:
```
In production the only caller is Poll (manager.go), one call per
// poll passing that poll's status.ScrapedAt, so samples normally arrive in time
// order and the aged prefix is all that is removed. Skew or a re-stamped scrape
// breaks that order, so the scan also drops an out-of-order aged sample, keeping
// the window bounded to the span regardless of arrival order.
```

i think this can be cut?

umh-core/pkg/fsmv2/benthos_monitor/manager.go
```diff
@@ -0,0 +1,280 @@
...
+	MessagesPerSecond float64
+	LastCount         int
+}
+
+// BenthosMonitorStatus is the result of one scrape of the benthos monitor: the
+// time it happened, the /metrics counters, /ping liveness, /ready readiness, the
+// /version string, and the per-direction throughput computed over windowSpan.
+//
+// BenthosMetrics is the per-path struct the fsmv1 parser produces, carried
+// whole. Narrowing it to scalars breaks the four connection-counter reads that
```
JeremyTheocharis:
```
arrowing it to scalars breaks the four connection-counter reads that
// decide whether a bridge is connected; those reads, their access path and their
// two consumers are named in
// TestMapObservedDeliversParsedConnectionCountersToBridgeHealth.
```

can be cut?

umh-core/pkg/fsmv2/benthos_monitor/manager.go
```diff
@@ -0,0 +1,280 @@
...
+// endpoints and the throughput window that accumulates counter samples across
+// polls. Poll receives a pointer to the deps, so the same *http.Client is
+// shared across polls.
+type benthosMonitorDeps struct {
+	client *http.Client
+	window throughputWindow
+}
+
+// ComponentThroughput carries the rate for one direction of the benthos
+// pipeline. MessagesPerSecond is the by-time rate the worker computes (never the
```
JeremyTheocharis: explanations can be moved to the fields itself?

umh-core/pkg/fsmv2/benthos_monitor/manager.go
```diff
@@ -0,0 +1,280 @@
...
+	monitorClientTimeout = 400 * time.Millisecond
+
+	// maxScrapeBody caps the response bodies Poll reads, so a misbehaving
+	// monitor cannot be buffered without bound on each poll.
+	maxScrapeBody = 4 << 20 // 4 MiB
+)
+
+// benthosMonitorDeps holds the HTTP client used to scrape the benthos monitor's
+// endpoints and the throughput window that accumulates counter samples across
+// polls. Poll receives a pointer to the deps, so the same *http.Client is
```
JeremyTheocharis:
```
Poll receives a pointer to the deps, so the same *http.Client is
// shared across polls.
```

can be cut?

umh-core/pkg/fsmv2/benthos_monitor/manager.go
```diff
@@ -0,0 +1,280 @@
...
+	// computed the same rule from a tick delta (s.IsActive =
+	// s.Input.MessagesPerTick > 0).
+	IsActive  bool
+	PingAlive bool
+	Ready     bool
+	Version   string
+}
+
+// Poll scrapes the configured benthos monitor's /ping, /ready, /version, and
+// /metrics endpoints once, using the HTTP client d carries. All requests carry
```
JeremyTheocharis, line 116:
```
All requests carry
// ctx, so cancellation surfaces as a request error. Deps that were never bound
// are rejected before any request goes out: substituting a client would report a
// wiring fault as ordinary idle traffic.
```

can be cut

umh-core/pkg/fsmv2/benthos_monitor/manager.go
```diff
@@ -0,0 +1,280 @@
...
+		WorkerType: WorkerType,
+		Log:        deps.NewFSMLogger(logger.For(managerName)),
+		ExtractConfigs: func(s publicfsm.SystemSnapshot) []config.BenthosMonitorConfig {
+			return s.CurrentConfig.Internal.BenthosMonitor
+		},
+		NameOf:      func(c config.BenthosMonitorConfig) string { return c.Name },
+		CfgFor:      cfgFor,
+		MapFresh:    mapFresh,
+		MapObserved: mapObserved,
+		// DesiredRunning is the state reported when a config leaves desiredState
```
JeremyTheocharis, line 272: i dont get what this is for? and we even need this comment? isnt this jsut expalantion of statevocabulary?

PR #2688, opened 2026-08-06
docs: fix GraphQL default state and mark it early access

robertschwarz-umh:
## Problem

Two HTTP API reference pages said the GraphQL API was enabled by default. It's off by default, so a customer who followed the docs found no config to reach it.

## What we are shipping

- Corrected the default to "disabled; opt in via `graphql.enabled`" on both reference pages, so all three GraphQL pages now agree.
- Added an "Early Access" note to the three GraphQL pages so the feature reads as an opt-in preview.

Fixes ENG-5527

### DIFF
--- a/umh-core/docs/reference/http-api/README.md
+++ b/umh-core/docs/reference/http-api/README.md
@@ -1,7 +1,7 @@
 # HTTP API Reference
 
 {% hint style="info" %}
-**Early Access.** We built the GraphQL API for developers who want to query the Unified Namespace programmatically. It's experimental, so it's off by default. Switch it on with `graphql.enabled: true` and give it a try. Tell us what you think.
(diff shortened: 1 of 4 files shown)


umh-core/docs/reference/http-api/README.md
```diff
@@ -1,7 +1,11 @@
 # HTTP API Reference
 
+{% hint style="info" %}
+**Early Access.** We built the GraphQL API for developers who want to query the Unified Namespace programmatically. It's experimental, so it's off by default. Switch it on with `agent.graphql.enabled: true` and give it a try. Tell us what you think.
```
Aaron99B: Where do the users have to set this?

JeremyTheocharis (approved):

umh-core/docs/reference/http-api/README.md
robertschwarz-umh: ahh, nice catch, thank you!!! I just went ahead and copy-pasted the same info box into all 3 pages without thinking about it too much.

All hint-boxes about this being early-access should link to `./topic-browser-graphql#schema`. Will update on Monday!

RuneRoven (approved):

PR #2692, opened 2026-08-10
[ENG-5550] skip the YAML round-trip

rpaland:
## Problem

Every time umh-core checks whether a bridge's running configuration still matches the desired one, it re-serializes the whole config to YAML text and parses it back. On both sides, several times per bridge, every tick. On instances with large bridge configurations that is where most of the CPU goes, and the control loop runs out of time to reconcile.

## What we are shipping

- The comparison now works directly on the values instead of going through text. Bridges on large configurations get reconciled again, and the control loop's cycle time roughly halves. 
- Nothing about the outcome changes. The direct path reproduces what the old one produced, and where it cannot do that exactly it declines and lets the old code run, so it can never make two matching configurations look different.
- If it ever does decline, that is now visible in the log instead of being silently slow.

## What we are NOT shipping

- A fix for the old text-based comparison itself, which turns out to mangle deeply indented blocks a little more on every pass. That predates this change and needs its own PR, filed separately with a reproduction.
- Proof that this resolves the customer report it came from. Their instance was not actually short on reconcile time, so this buys them headroom rather than fixing what they saw.


umh-core/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Improvements
+
+- Instances with large bridge configurations no longer starve their reconcile loop. Checking whether a bridge's running configuration still matches the desired one re-serialized it to YAML text and parsed it back, on both sides of every comparison and several times per bridge per tick. This could under the wrong circumstances push the control loop past its tick budget, so bridges stopped being reconciled and could show up as inactive. The comparison now works directly on the values and falls back to the old path only for the rare shapes it cannot reproduce exactly, which roughly halves the control loop's cycle time in every configuration measured.
```
JeremyTheocharis: add the benchmark results to the changelog and say how much average cpu consumption on how big config.yaml can be saved, and how much it reduces "CPU pressure" also known as "starvations"

JeremyTheocharis:
@m4jus

FalconTube (approved):
Looks good, did nice pair review!

PR #2693, opened 2026-08-10
fix(ENG-3670): S6 directory health message varies only by trigger

JeremyTheocharis:
## Problem

Sentry alerts are too noisy. S6 directory health has >100 distinct Sentry issues because the service name is interpolated into the error message, and Sentry fingerprints on the message — so every new bridge name mints a new permanent issue.

## What we're shipping

The message now carries only `trigger`, which has five bounded literals across the five `logS6DirectoryState` call sites. Service name and S6 state continue to travel in `diagnosticContext`, which `pkg/sentry` promotes to searchable event tags — so nothing is lost on the Sentry side.

Plus a regression test that was observed failing before believing it: re-adding `service=%s` to the format string turns it red.

## What we're NOT shipping

* Not touching `pkg/sentry/sentry.go` / `{{ default }}` — that fingerprint is shared by \~146 FSMv1 call sites, and removing the default there would collapse unrelated errors into one issue per level.
* Not deleting any call site. The empty-state call on the `IsBenthosS6Stopped` success branch is separate work.
* No behaviour change: severity, control flow, retry and the 2h debounce are untouched.
* No retroactive merge. Existing issues keep their fingerprints and go quiet; bulk-resolving them is a separate step and must be resolve-in-release, or the next event regresses them.
* The WARN log line loses the service name. Accepted: that line is already debounce-gated, and the Debug diagnostics retain full detail.

### DIFF
--- a/umh-core/pkg/fsm/benthos/actions.go
+++ b/umh-core/pkg/fsm/benthos/actions.go
@@ -308,10 +308,8 @@ func (b *BenthosInstance) logS6DirectoryStateAsync(
 	)
 }
 
-// s6DirectoryHealthError builds the error Sentry groups the S6 directory health
-// diagnostics by. Sentry fingerprints on the message, so it carries only the
-// trigger, which has five bounded literals; the service name and S6 state travel
-// as tags in the diagnostic context instead.
+// Sentry fingerprints on the message, so this must not interpolate per-service or
+// per-state values — they travel as tags instead. Guarded by s6_diagnostic_message_test.go.
 func s6DirectoryHealthError(trigger string) error {
 	return fmt.Errorf("S6 directory health issue: trigger=%s", trigger)
 }
--- a/umh-core/pkg/fsm/benthos/actions.go
+++ b/umh-core/pkg/fsm/benthos/actions.go
@@ -301,12 +301,19 @@ func (b *BenthosInstance) logS6DirectoryStateAsync(
 	}
 
 	sentry.ReportIssueWithContext(
-		fmt.Errorf("S6 directory health issue: service=%s, state='%s', trigger=%s",
-			serviceName, s6State, trigger),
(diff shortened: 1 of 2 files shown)


umh-core/pkg/fsm/benthos/actions.go
```diff
@@ -301,14 +301,21 @@ func (b *BenthosInstance) logS6DirectoryStateAsync(
...
-		fmt.Errorf("S6 directory health issue: service=%s, state='%s', trigger=%s",
-			serviceName, s6State, trigger),
+		s6DirectoryHealthError(trigger),
 		sentry.IssueTypeWarning,
 		logger,
 		diagnosticContext,
 	)
 }
 
+// s6DirectoryHealthError builds the error Sentry groups the S6 directory health
```
Aaron99B: nit: I would remove this code comment here. The function is self-explanatory

Aaron99B (approved):

umh-core/pkg/fsm/benthos/actions.go
JeremyTheocharis: Half-taken, and thanks — the comment was too long.

Dropped the part that restated the code (it "carries only the trigger" is visible from the one-line body). Kept one line for the bit the signature genuinely cannot express: that interpolating a per-service value here mints a Sentry issue per bridge. That is not a hypothetical — it is what ENG-3670 shipped in Oct 2025, and it stayed live until now (679 distinct issue groups over 90 days, production-only). The function looks like a plain error constructor precisely because the constraint lives outside its type.

Also added a pointer to `s6_diagnostic_message_test.go`, so someone hitting a red test finds the reason rather than assuming the test is wrong.

Happy to drop it entirely if you still think the test alone carries it — it is your call and I will not re-litigate.

PR #2694, opened 2026-08-10
fix(fsmv2): Sentry stack traces name the caller, not the logger

JeremyTheocharis:
## Problem

Every FSMv2 Sentry issue shows its culprit as `(*zapLogger).SentryError`.

## What we're shipping

Two terms added to the existing `frame.Function` check, beside `captureToSentry` / `SentryHook.Write` / `zapcore.`:

```go
strings.Contains(frame.Function, "SentryError") ||
strings.Contains(frame.Function, "SentryWarn") ||
```

## What we're NOT shipping

* No change to what is captured, grouped, or reported. Stack-trace contents only.


Aaron99B:
I don't fully understand the problem from the problem description

JeremyTheocharis:
added a screenshot :D

Aaron99B (approved):

PR #2695, opened 2026-08-10
Memoize the rendered runtime config per bridge

rpaland:
## Problem

Based on PR #2692 

Two things in the reconcile loop redo their full work on every tick, whether or not
anything they depend on has changed. Each bridge's configuration is rebuilt from its
template, and each nmap check re-scans everything still sitting in that service's log
buffer to work out the result of the newest scan. On instances with many bridges the
two together are most of what the loop does.

## What we are shipping

- A bridge's built configuration is kept and reused until one of the inputs that
  produced it actually differs. Instances with many bridges get noticeably more
  reconcile time back, and CPU drops with it.
- Reuse is never assumed. Every input is stored alongside the result and compared on
  each call, so a stale answer cannot survive a config change even if nothing notifies
  us about it. What is kept is bounded, so bridges created and deleted over a long
  time cannot accumulate.
- Reading an nmap result no longer depends on how much log history a service has kept.
  It looks for the newest scan from the end and stops there instead of walking
  everything, and the patterns it matches with are prepared once rather than rebuilt on
  every check.

## What we are NOT shipping

- Anything that reduces how often the loop compares a bridge's desired and running
  configuration. That comparison is the point of reconciling and has to happen every
  tick. Only the work in front of it is now skipped.
- The same treatment for benthos, which reads its service logs the same way and was
  not touched here.
- A change to how service logs are handed to the code that reads them. Each read still
  copies the whole retained buffer even when almost none of it is new. That is the next
  thing worth looking at and needs its own change.

Fixes ENG-5561

### DIFF
--- a/umh-core/pkg/service/nmap/nmap.go
+++ b/umh-core/pkg/service/nmap/nmap.go
@@ -22,6 +22,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/config"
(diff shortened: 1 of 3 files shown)


umh-core/pkg/service/nmap/nmap.go
```diff
@@ -38,6 +39,33 @@ import (
 	"go.uber.org/zap"
 )
 
+// Compiled once. These used to be built inside parseScanLogs, which runs for every
```
JeremyTheocharis, line 42: unsure that we even need this. we have with nmap in fsmv2 already a good fix for this that is way more performant. this here could introduce new bugs, so lets keep the current unoptimized thign as a fallback for the new fsmv2 nmap worker

umh-core/pkg/service/protocolconverter/runtime_config/runtime_config.go
```diff
@@ -77,6 +79,72 @@ const BridgedByPlaceholder = "unimplemented"
...
+	agentLocation map[string]string
+	globalVars    map[string]any
+	nodeName      string
+	spec          protocolconverterserviceconfig.ProtocolConverterServiceConfigSpec
+	result        protocolconverterserviceconfig.ProtocolConverterServiceConfigRuntime
+}
+
+// renderMemo caches the last render per bridge, keyed by pcName. BuildRuntimeConfig
+// runs for every bridge on every tick and is a pure function of its arguments, which
+// only change when the config file does. A profile attributed 12.9% of all CPU to the
```
JeremyTheocharis, line 96: Can we benchmark this properly and attach the benchmark to the PR as a comment or something? you could run with umh-experiment and could measure for the same config how much CPU (and pressure!) it was using before, and then how much was later. Could be a way better chagnelog entry to say "reduced average CPU by 13% in umh-core instances with over x thousand lines of code" or so

JeremyTheocharis (changes requested):
additionally could you make problem, what we are shipping, and what we are not shipping more clearer? it reads currently still inteh LLM informational overload mode.<br><br>also what we are not shipping makes sense to say what was done inteh PR before, so one knows why this is stacked upon.

JeremyTheocharis:
@m4jus

PR #2696, opened 2026-08-10
feat(fsmv2): CPU monitor worker behind USE_FSMV2_CPU (default off)

JeremyTheocharis:
Ticket: ENG-5128

## Problem

CPU health flickers and is only dependent on average CPU at the moment. This stacked PR introduces a new calculation method.

## What we are shipping

- An fsmv2 monitor worker that runs the CPU calculation from the previous PRs: polls the signals each tick and reports a verdict with a plain message
- Never invents a reading: a signal the box cannot measure is reported as unmeasured, and a tick that cannot measure at all reports that rather than a healthy zero
- Raises a single Sentry warning, once per worker rather than once per tick, when a signal the box *could* answer has still produced no reading 10 seconds after start — so a box that cannot fully see its own CPU is surfaced instead of silently reporting healthy
- Said worker is registered and de-registered by the existing config-worker behind `USE_FSMV2_CPU` (default off), the same way the historian monitor already is

## What we are NOT shipping

- Connecting the verdict to the Management Console — the communicator still reports the old CPU record; that is the next PR
- `pkg/diagnosis` and `pkg/cpuhealth` themselves — those are the previous PRs
- The scenarios that drive this worker over a fake machine — they are their own PR on top of this one, so that the worker and the test harness can be reviewed as the separate things they are
- The flag default on, and a changelog entry — nothing user-visible changes yet
- Blocking new bridges on the CPU verdict — `IsResourceLimited` still reads the fsmv1 container monitor, and wiring the fsmv2 verdict into it is a later PR

PR #2697, opened 2026-08-10
feat(cpu): seam sends fsmv2 CPU verdict to the Management Console behind USE_FSMV2_CPU (default off)

JeremyTheocharis:
Ticket: ENG-5128

## Problem

CPU health flickers and is only dependent on average CPU at the moment. This stacked PR introduces a new calculation method.

## What we are shipping

- **When `USE_FSMV2_CPU` is on, the behaviour changes from the rollout ticket go live on a deployed instance** — the measured signals (pressure, starvation, throttle) replace the fixed 70% usage mark, the verdict no longer flickers, and the CPU record reaches the Management Console through the new calculation, not the old one.
- Uses the CPU health from the new calculation when the flag is enabled and sends it to the Management Console
- `totalUsageMCpu` and `coreCount` are omitted on the wire when a tick genuinely cannot measure (a real 0 is still 0)
- `featureUsage` reports whether this instance actually runs the new path (`fsmv2CpuEnabled`), so the Console can tell a box that supports it from one that doesn't
- When the new calculation is on but cannot measure (stale, never-observed, or unreadable), CPU health degrades with a plain reason — it is never reported healthy on missing data

## What we are NOT shipping

- Flipping `USE_FSMV2_CPU` on by default — that is its own later PR
- Counting how often the CPU verdict blocks new bridges — a later PR in this stack
- Removing the old CPU path — clean-up is the final PR in the stack
- Showing the new metrics in the Console also needs the frontend PR merged (Management Console) — this PR ships the wire; the UI reads it only after that PR lands


JeremyTheocharis:
## What the Management Console needs on this wire

Written while preparing the MC side (ManagementConsole#3308, stacked under #3480). Posting here because p4 owns the seam, and the MC's extended CPU panel currently has no producer.

### The situation

The MC panel reads seven things that are on no branch:

`state`, `attribution`, `causes[]`, `avgMCpu`, `p95MCpu`, `p99MCpu`, and `verdictBasis` (with `headroom`, `hostBusy`, and a `{value, threshold, fired, applies}` triple per signal).

Where they came from: **#2613, closed unmerged on 2026-08-17.** `verdictBasis` appears 159 times in that PR's diff and **zero** times in p3 (#2696) and p4 (#2697). `pkg/cpuhealth` did land on staging as an internal package, but its output does not reach the wire. `models.CPU` on staging is still six fields, and p4 changes two of them to pointers without adding any.

So the MC is not reading a wire that needs simplifying. It is reading a wire we dropped. p4's replacement sends `CPUStatus.Message`, a rendered sentence from `ComposeMessage`. A sentence cannot fill a metric panel: the MC would have to parse prose to get a number back out.

That is the decision to make here, and it is a design choice rather than a bug. **Either the panel dies and we ship the sentence, or the wire carries the numbers the panel needs.** Both are defensible. What does not work is the current split, where umh-core plans to send prose and the MC has built a table.

### If we keep the panel, this is the payload that needs no re-derivation

Per signal that has an opinion (throttling, pressure, steal, saturation, host contention):

```
{ kind, value, threshold, fired, applies }
```

`Cause` in `attribute.go` already carries `Kind`, `Value` and `Unit`. Adding `threshold` and `applies` makes it the whole panel. `applies` matters because "not measurable here" and "measured and fine" must render differently, and a zero cannot distinguish them.

Plus, once:

```
ceiling: "host" | "limit"      // which ceiling the verdict used
capacity                        // that ceiling's size
avg, p95, p99                   // measurements, not decisions
hostBusy: { mean, available }
primaryCause: <kind>            // which signal to mark, decided once
limitedVisibility: bool         // already exists, keep it
```

**Please do not send these.** `headroom.limitSaturationFired`, `hostFullFired`, `noHostStatsSaturationFired`, `noLimitHostFired`, or the per-signal `fired` latch as a separate channel from `causes[]`. Those are `decide.go` internals, and publishing them means the MC reconstructs a decision umh-core already made. That reconstruction is what Aaron flagged on the MC PR, and it is the anti-pattern our own abstractions standard names: kubelet publishes three condition types from nine internal signals, not the nine.

One mapping subtlety for whoever does it: `noHostStatsSaturationFired` maps to `limitedVisibility` **and** a saturation cause. Never `limitedVisibility` alone.

### What should stay in the frontend

Worth saying, because "move business logic out of the frontend", taken literally, lands in the wrong place. Units, labels, row order and colour belong in the MC. If a row label ships from Go, changing a word needs a umh-core release. Send semantics; the MC renders them.

### Two things in p4 that break the MC today

**`TotalUsageMCpu *float64` and `CoreCount *int` with `omitempty`.** The MC's TS mirror declares both required. `cpuUsageSample.ts` reads `cpu.totalUsageMCpu` straight into a telemetry event, so an unmeasured tick would emit `usageMCpu: undefined`. The display paths survive on `Number.isFinite` guards, which is luck rather than design. The MC needs to make both optional, and I am flagging it so both sides move in the same release.

**p4 declares five cause kinds; the MC knows four.** `CauseKindHostContention` is missing from the MC's `RECOGNIZED_CAUSE_KINDS`, which is an exhaustive `Record<CauseKind, true>`. An unrecognised kind is dropped, and the MC's own comment records the cost: a throttled-and-degraded CPU whose only cause is the unknown kind loses the red on its throttled row. If `causes[]` ever reaches the wire, please treat the set as a pinned wire contract. The strings were last renamed on 2026-08-25 and the MC cannot track that silently.

### On deleting `containerCpuDisplay.ts`

It will not go away, and I would not aim for that. It is 975 lines, and roughly where they go:

| what | lines | after a good wire |
|---|---|---|
| shape gate (`VERDICT_BASIS_FIELD_CHECKS`, `isValidVerdictBasis`, fingerprint) | 139 | ~20, because most of it exists only while the payload is wide, optional and unvalidated |
| legacy path for instances with no basis | 155 | 0 once the supported floor rises; not umh-core's problem |
| `computeVerdict` latch combining | 31 | 0, if conclusions arrive instead of latches |
| mode and ceiling re-derivation in `deriveCpuDisplay` | 143 | ~40, if the wire says which ceiling it used |
| row construction | 219 | ~120, still ours: which rows, in what order |
| number formatting | 122 | 122, stays |
| feature-flag gate, types | 82 | ~60, stays |

So roughly 975 down to 360. The reduction is worth doing. The file is not the problem; the duplicated decision-making is.

PR #2698, opened 2026-08-10
fix(fsmv2): run the FSMv2 runtime without Management Console credentials

JeremyTheocharis:
> Stacked on united-manufacturing-hub/united-manufacturing-hub#2683 — based on that branch, not on staging.

### Problem

With either `API_URL` or `AUTH_TOKEN` unset, FSMv2 never started. That was fine while the communicator was the only FSMv2 component, but nmap and benthos_monitor are being added and they have to work with no Management Console connected.

### What we're shipping

FSMv2 is decoupled from the Management Console credentials, and the feature flags and legacy communicator code the migration left behind are cleared out.

* **nmap and benthos_monitor no longer need a Management Console connection.** The runtime always starts; credentials now decide only whether the connection to the console is wired up.
* **The legacy Management Console client is deleted** — 2,669 lines across 17 files. It was already unreachable once the FSMv2 transport took over, so nothing changes behaviour.
* `USE_FSMV2_TRANSPORT` **and** `USE_FSMV2_PROTOCOL_CONVERTER` **are gone as environment variables.**
* **The Dockerfile builds the** `./cmd` **package** instead of `main.go` by filename, so a second file in that package no longer breaks the image.

### What we're NOT shipping

* `ALLOW_INSECURE_TLS` **is not restored.** It has been dead on the default path since v0.44.9 — the FSMv2 transport never had an insecure option — and deleting the legacy client removes the last code that honoured it. Its documentation is removed here and the changelog says so. Restoring it end to end, together with a `Reset()` that discards the setting and the TLS 1.0 floor the old client applied, is [ENG-5587](https://linear.app/united-manufacturing-hub/issue/ENG-5587/allow-insecure-tls-is-silently-dead-restore-it-end-to-end-fix-reset), with groundwork in united-manufacturing-hub/united-manufacturing-hub#2706.
* **The retired config keys are not deleted** — that is [ENG-4900](https://linear.app/united-manufacturing-hub/issue/ENG-4900/rollout-config-architecture-refuse-invalid-config-fix-the-root-cause).
* **No automated regression guard on the invariant.** It is a property of `main()`'s control flow, so no test can call it. A no-credentials container run is the evidence that it holds today; a guard is tracked separately.
* **Three pre-existing gaps are left in place**, none introduced here: the adapter's warning reaches stdout but not Sentry, the 1 Hz notifier goroutine has no stop path, and `Watchdog.Start()` is never called in production.

### Confidence

Had Klaus test it usign prod frontend and prod backend against [`ghcr.io/united-manufacturing-hub/umh-core:sha-936c196:`](<http://ghcr.io/united-manufacturing-hub/umh-core:sha-936c196>)<br>

```
I ran a container check by hand against this PR's own image — ghcr.io/united-manufacturing-hub/umh-core:sha-936c196, which is the merge of 4dba10afd (this branch's head) into its base. The instance registered and went from offline to active in the console, a bridge created from the MC UI deployed and reached active, and it was still active more than five minutes later with no degraded states. Throughput held at 1.00 msg/sec, measured from Redpanda's own offsets rather than from the console's own number, and the console agreed. RestartCount stayed 0.
```


JeremyTheocharis:
@coderabbitai full review

JeremyTheocharis:
@coderabbitai review

umh-core/cmd/renderer.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	if cfg.APIURL != "" && cfg.AuthToken != "" {
+		specBytes, err := yaml.Marshal(communicatorSpec{
+			RelayURL:     cfg.APIURL,
+			InstanceUUID: instanceUUID,
+			AuthToken:    cfg.AuthToken,
+			Timeout:      "10s",
+			State:        "running",
+		})
```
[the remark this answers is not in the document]
JeremyTheocharis: Fixed in 9cd97156f, and the finding understated it: there were **two** swallowing paths in this file, `:62` and `:75`, not one.

Worth naming why it mattered more than an unchecked error. Returning `""` does not just drop the communicator child — it yields a document with **zero** children, persistence included, so the supervisor comes up with nothing to supervise and no error anywhere. That is the exact silent bring-up stall this PR exists to remove, reintroduced in the PR's own new file. It now returns `(string, error)` and `buildFSMv2Supervisor` propagates it, matching the existing early-return-after-`Start(ctx)` precedent at `main.go:671`.

Honest limit: the unreachable half stays unguarded. Both structs hold only strings, so `yaml.Marshal` has no reachable failure mode and a re-swallowed error would be behaviourally identical to correct code — no test can catch that. What is guarded is the consequence: a mutation returning an empty document with no error fails the persistence assertion, verified red on the mutant and green after restoring the file byte-identical.

umh-core/cmd/renderer_test.go
```diff
@@ -0,0 +1,131 @@
...
+		Expect(yaml.Unmarshal([]byte(comm.UserSpec.Config), &userCfg)).To(Succeed())
+		Expect(userCfg.RelayURL).To(Equal(relayURL))
+		Expect(userCfg.AuthToken).To(Equal(authToken))
+		Expect(userCfg.InstanceUUID).To(Equal(placeholderUUID))
+
+		// Partial credentials (only API URL) are treated as no credentials
+		// (E2): the existing && contract must be preserved in the renderer.
+		partial := config.AgentConfig{CommunicatorConfig: config.CommunicatorConfig{APIURL: relayURL}}
+		Expect(childNames(render(partial, "uuid-1"))).
+			NotTo(ContainElement("communicator"))
```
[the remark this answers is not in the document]
JeremyTheocharis: Fixed. Worth recording that the stated reason was wrong even though the fix was right: `nlreturn` **is** enabled (`umh-core/.golangci.yaml:42`), but golangci-lint runs in **no** workflow in this repo — `grep -rl golangci .github/workflows/` returns nothing while `nilaway` appears in two — and the local hooks are deliberately unplugged. So nothing was going to report this as an error. Taken anyway because it is a one-line consistency fix in a file this PR adds.

umh-core/cmd/renderer_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/config"
+)
+
+var _ = Describe("renderSupervisorChildrenYAML", func() {
+	// renderSupervisorChildrenYAML is the pure core of buildFSMv2Supervisor's
+	// child selection. The design decision that matters is that choosing the
+	// children moves out of the effectful builder, so the central invariant --
+	// persistence always, communicator iff credentials -- is provable without a
+	// supervisor, a store or a goroutine.
```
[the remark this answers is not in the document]
JeremyTheocharis: Declining, same as the equivalent suggestion on #2683 — **leaving this thread open** rather than resolving it, since it is a judgment call and Jeremy may disagree.

The comments in this spec record *why* the assertions exist: which escape (E2, E14/E15) each one pins, and why a hostile auth token disqualifies `fmt.Sprintf` and forces `yaml.Marshal`. That reasoning is not recoverable from the test name, and it is the part a future reader needs before changing the assertion.

Where comments in this stack were genuinely wrong they were deleted, not defended: five `USE_FSMV2_MEMORY_CLEANUP` sites on this PR, two of which instructed operators to set a now-dead flag to "revert to the legacy behavior" this branch removes. Three of those five were not reported here and were found by grepping for the pattern behind the two that were.

umh-core/CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+### Breaking Changes
+
+- The legacy backend communicator and its `USE_FSMV2_TRANSPORT` environment variable are removed, so the documented rollback to the legacy communicator no longer exists. If you set `USE_FSMV2_TRANSPORT=false` to stay on the legacy path, that variable is now ignored and umh-core runs the FSMv2 communicator regardless. The `useFSMv2Transport` and `useFSMv2MemoryCleanup` settings in `config.yaml` likewise have no effect; they are still accepted, so you can leave them in place rather than editing your config
```
JeremyTheocharis: these are not breaking changes. this is just another improvement

umh-core/pkg/fsmv2/workers/transport/http/http.go
```diff
@@ -209,48 +209,55 @@ type HTTPTransport struct {
 // If timeout is 0, defaults to LongPollingDuration (30 seconds).
 // Transport settings match legacy communicator to avoid Cloudflare issues.
 // Connection pooling replaces the legacy DisableKeepAlives setting.
-func NewHTTPTransport(relayURL string, timeout time.Duration) *HTTPTransport {
+// If insecure is true, TLS certificate verification is skipped.
+func NewHTTPTransport(relayURL string, timeout time.Duration, insecure bool) *HTTPTransport {
```
JeremyTheocharis: this feels like the TLS ticket leaking into this, we should maybe take that out?

umh-core/CHANGELOG.md
JeremyTheocharis: Done in `369bc62d0`. You were right and my entry was wrong. The bar this changelog uses is visible in its only two real Breaking Changes entries: a UID change requiring a `chown`, and an environment variable you must add. Both mean the operator must act or they are broken. Losing the `USE_FSMV2_TRANSPORT=false` rollback requires no action and breaks nothing, so it belongs under Improvements. The heading is gone from `## Unreleased` and the entry is shorter.

Worth knowing for next time: `### Breaking Changes` **is** a legal category in the validator (`VALID_CATEGORIES` in `changelog.umh.app`), and omitting a category is allowed. So this was a judgement about the bar, not a schema fix.

umh-core/pkg/fsmv2/workers/transport/http/http.go
JeremyTheocharis: Done in `f73b071d9`, and the evidence was worse than the comment suggested. `NewHTTPTransport` took an `insecure bool` and the single production call site passed a hardcoded `false`. Nothing on this branch reads `config.Agent.AllowInsecureTLS`, since the legacy communicator that read it is deleted here, so the parameter could only ever receive `false`: a knob wired to a constant, which reads as implemented and is worse than its absence.

`NewHTTPTransport` and `buildHTTPClient` are back to staging's two-argument form. `http.go` now diffs **empty** against `origin/staging`, so this reverts rather than reshapes.

Two follow-ons, in `856aeafc7`. Ten documentation sites still promised the flag disabled certificate verification, so they are removed and the changelog says the option has no effect and points at the corporate CA certificate route. The config key stays parsed and the struct field stays declared, because deleting it would stop instances that still carry the key from loading their config at all.

The insecure-override spec went with the parameter. The secure-rejection spec stayed, on evidence rather than taste: setting `InsecureSkipVerify` in `buildHTTPClient` turns it red, so it is a live guard rather than a test of unchanged behaviour.

umh-core/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
...
+- The `USE_FSMV2_PROTOCOL_CONVERTER` environment variable is removed. It never gated any behavior, so setting it did nothing. Its `useFSMv2ProtocolConverter` config key is still accepted and still has no effect, along with `enableFSMv2` and `useFSMv2MemoryCleanup`, so retired flags can stay in your `config.yaml` without changing anything
+- `ALLOW_INSECURE_TLS` and `agent.communicator.allowInsecureTLS` no longer have any effect. The new FSMv2 communicator never implemented this option, and nothing indicated it was in use. For corporate networks that intercept TLS, add your corporate CA certificate instead; see the corporate firewalls guide. The config key is still accepted, so you do not need to edit your `config.yaml`
+- umh-core now starts and runs without `AUTH_TOKEN` or `API_URL`. Previously an instance with no Management Console credentials brought up no runtime at all: its components reported `starting` indefinitely, with no error and no exit, so a missing or mistyped token looked like a hung container. Such an instance now runs fully — it simply cannot reach the Management Console
```
FalconTube, line 10: I feel like these lines are quite detailed and not super interesting for a user.<br>What they need to know, is that the ENV variables can be removed, because they are now the default.<br><br>But no strong opinion on it.

umh-core/cmd/renderer.go
```diff
@@ -0,0 +1,88 @@
...
+// strings, so yaml.Marshal has no reachable failure mode today -- but an empty
+// document is not a degraded result, it is a supervisor with no children at
+// all, persistence included. That is the silent bring-up stall this package
+// exists to prevent, so the error travels up to the caller instead.
+func renderSupervisorChildrenYAML(cfg config.AgentConfig, instanceUUID string) (string, error) {
+	children := []childNode{
+		{Name: "persistence", WorkerType: "persistence"},
+	}
+
+	if cfg.APIURL != "" && cfg.AuthToken != "" {
```
FalconTube, line 61: To avoid duplication and confusion, we could re-use the new `communicatorEnabled()` function from the `main.go` file here.<br>Would just need an adjustment in the `communicatorEnabled()` function to take the `Agent` config, instead of the `Full` config.

umh-core/pkg/config/config.go
```diff
@@ -101,17 +101,25 @@ type AgentConfig struct {
...
+	// These are accepted and ignored. Nothing reads them: FSMv2 and its
+	// persistence worker run unconditionally. They stay declared because config
+	// parsing rejects unknown keys (ParseConfig passes allowUnknownFields=false,
+	// so KnownFields is on), which means deleting a field stops an instance from
+	// loading a config.yaml that still carries it, at startup and on every
+	// refresh. Removing one is a breaking change, not a cleanup. retired_flags_test.go
+	// pins every key here.
+	EnableFSMv2               bool `yaml:"enableFSMv2,omitempty"`
+	UseFSMv2ProtocolConverter bool `yaml:"useFSMv2ProtocolConverter,omitempty"`
+	UseFSMv2MemoryCleanup     bool `yaml:"useFSMv2MemoryCleanup,omitempty"`
```
FalconTube, line 115: Had a thought:<br>What about implementing a "catchall" field, that catches all unknown fields and gathers them under something like "unusedOptions".<br>That way we <br>a) can add and remove any past/future keys as we wish<br>b) don't break the entire config.yaml, if something in the Agent config is set incorrectly<br><br>Only downside it, that it would silently catch any typos, that are made to any keys here in the agent config

FalconTube (changes requested):
Some minor findings, then good to go.<br>Might not need any fixes at all, if not deemed worthy.<br>In that case, just reply to the comments and request my re-review 🙂

PR #2708, opened 2026-08-13
nmap: verify the connection before reporting a bridge edit rolled out (ENG-5579, ENG-5586)

JeremyTheocharis:
Bottom of the stack: **#2708** → #2716 → #2751. Umbrella: ENG-5853.

## Problem

Editing a bridge's connection to a target that does not answer reports a successful deploy. No config is corrupted, and the overview page shows the connection degraded a moment later. What gets lost is the rollback.

Three independent causes, each enough on its own:

1. nmap reported the config it was *asked* to scan as the config it observed. `awaitRollout` compares observed against requested, so it compared 445 with 445 and passed on the first tick, before anything had dialled the new port. FSMv2 only.
2. Only the port number was compared, never whether the port answered. `Poll` records the port it dialled even when the dial is refused, so a scanned-and-closed port passed too.
3. Only the port was compared, never the host. Repointing a bridge from one PLC to another on 502 was accepted by the old host's scan, with nothing having dialled the new host (ENG-5586).

Causes 2 and 3 are on the default backend as well, so those halves change deploy behaviour for every instance.

## What we are shipping

The check for connection-only edits, the branch a bridge with no flows takes.

- An edit to a target that does not answer now fails and rolls back. The check waits for the requested endpoint to be scanned and found **open**, and `NmapStatus` carries the target and port actually dialled.
- Both halves of the endpoint are compared, so a move to a different host on the same port no longer passes on the old host's scan.
- A port reporting filtered rather than open fails too. Affects fsmv1 as well.
- A scan taken before the edit is no longer evidence about the new endpoint. `NmapStatus` records when a scan started, on the open, refused and cancelled paths alike, and the check requires a scan started after the persist.
- A templated connection is compared against its resolved endpoint, re-resolved every tick. A historian bridge scans `{{ .historian.timescale.host }}`, which the old comparison could never match, so every such edit waited out the 30-second timeout and rolled back.
- `PortState` and its constants move from `pkg/fsm/nmap` to `pkg/service/nmap`, where the scan result already lives. No behaviour change.

The `IsRunning` divergence between backends has a test behind it rather than only a comment: a scanner reporting running with a closed port must fail the rollout. Mutating the check to trust `IsRunning` reddens that spec and nothing else.

## What we are NOT shipping

- The check still runs only for a bridge with no flows. A bridge with a read or write flow is accepted on its own pre-edit state; #2716 covers it.
- For a flowless bridge whose target is down, a location-only or state-only edit now fails where it used to succeed. ENG-5858 tracks whether the exemption #2716 adds can ever fire for a console edit.
- Deploying a new bridge still checks nothing. ENG-5599.
- Global variables still reach `BuildRuntimeConfig` as `nil` at three call sites. ENG-5855.

Fixes ENG-5579
Fixes ENG-5586

### DIFF
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -4,8 +4,7 @@
 
 ### Fixes
 
-- Editing a bridge's connection to a port that does not answer now fails and restores the previous configuration, instead of reporting a successful deploy. The deploy waits for a scan that finds the new port open, so a port that is closed or filtered is no longer accepted
-- Repointing a bridge to a different host on the same port is now checked too. The deploy waits for a scan taken after the change, so the previous host's scan can no longer satisfy it
+- Repointing a bridge to a different host on the same port now waits for a check of the new host, so the previous host can no longer make the deploy report success
 
 ## [0.44.38]
 
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixes
 
-- Bridge connection edits to an unreachable port no longer report the deployment as successful and skip the rollback, on instances running the FSMv2 connection probe (`NMAP_BACKEND=fsmv2`, off by default). The rollout now keeps waiting until a scan finds the new port open, not merely until the requested port number has been probed
+- A connection edit no longer completes on a stale reading of the old connection. The rollout waits for a scan of the new port that finds it open
 
 ## [0.44.38]
 
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixes
 
 - Editing a bridge's connection to a port that does not answer now fails and restores the previous configuration, instead of reporting a successful deploy. The deploy waits for a scan that finds the new port open, so a port that is closed or filtered is no longer accepted
+- Repointing a bridge to a different host on the same port is now checked too. The deploy waits for a scan taken after the change, so the previous host's scan can no longer satisfy it
 
 ## [0.44.38]
 
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixes
 
-- Bridge connection edits to an unreachable port no longer report the deployment as successful and skip the rollback, on instances running the FSMv2 connection probe (`NMAP_BACKEND=fsmv2`, off by default). The check compared the requested port against itself instead of against the port that was actually probed, so it passed on the first attempt
+- Bridge connection edits to an unreachable port no longer report the deployment as successful and skip the rollback, on instances running the FSMv2 connection probe (`NMAP_BACKEND=fsmv2`, off by default). The rollout now keeps waiting until a scan finds the new port open, not merely until the requested port number has been probed
 
 ## [0.44.38]
 
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixes
 
-- A connection edit to a port that does not answer no longer completes. The rollout waits for a scan of the new port that finds it open
+- Editing a bridge's connection to a port that does not answer now fails and restores the previous configuration, instead of reporting a successful deploy. The deploy waits for a scan that finds the new port open, so a port that is closed or filtered is no longer accepted
 
 ## [0.44.38]
 
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixes
 
-- A connection edit no longer completes on a stale reading of the old connection. The rollout waits for a scan of the new port that finds it open
+- A connection edit to a port that does not answer no longer completes. The rollout waits for a scan of the new port that finds it open
 
 ## [0.44.38]
 
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Bridge connection edits to an unreachable port no longer report the deployment as successful and skip the rollback, on instances running the FSMv2 connection probe (`NMAP_BACKEND=fsmv2`, off by default). The check compared the requested port against itself instead of against the port that was actually probed, so it passed on the first attempt
+
 ## [0.44.38]
 
 ### Improvements
--- a/umh-core/pkg/communicator/actions/edit-protocolconverter.go
+++ b/umh-core/pkg/communicator/actions/edit-protocolconverter.go
@@ -583,6 +583,11 @@ func (a *EditProtocolConverterAction) awaitRollout(previousConfig config.Protoco
 		// consecutive identical failures instead of burning the full timeout.
 		prevRenderErrMsg     string
 		identicalRenderFails int
+
+		// currentStateReason holds what the most recent tick was waiting for. The
+		// timeout branch reads it, so a rollback message names the half that never
+		// arrived instead of only saying the bridge did not become active.
+		currentStateReason string
 	)
 
 	const maxIdenticalRenderFails = 3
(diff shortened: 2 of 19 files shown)


umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -1137,6 +1142,91 @@ func (a *EditProtocolConverterAction) mergeUserVariables(base map[string]any, in
...
+	if renderErr != nil {
+		a.lastRenderErr = renderErr
+
+		return "waiting to resolve the bridge's connection endpoint"
+	}
+
+	wantEndpoint := formatEndpoint(resolved)
+
+	nmapObs := pcSnapshot.ServiceInfo.ConnectionObservedState.ServiceInfo.NmapObservedState
+	scannedEndpoint := formatEndpoint(nmapObs.ObservedNmapServiceConfig)
```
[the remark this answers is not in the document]
FalconTube: I agree with CR here. I am quite sure, that at this point here, we can't know for certain if the pcSnapshot is the one that we expect.

umh-core/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Fixes
+
+- Editing a bridge's connection to a port that does not answer now fails and restores the previous configuration, instead of reporting a successful deploy. The deploy waits for a scan that finds the new port open, so a port that is closed or filtered is no longer accepted
```
FalconTube: This first line is only for fsmv2 implementation. It is not really exposed to the user. Let's not mention it here at all

umh-core/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
...
+- Repointing a bridge to a different host on the same port is now checked too. The deploy waits for a scan taken after the change, so the previous host's scan can no longer satisfy it
```
FalconTube: This is user facing, so we can show it. But could be a bit shorter

umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -507,7 +511,7 @@ func (a *EditProtocolConverterAction) persistConfig(atomicEditUUID uuid.UUID, ne
 // The function returns the error code and the error message via an error object.
 // The error code is a string that is sent to the frontend to allow it to determine if the action can be retried or not.
 // The error message is sent to the frontend to allow the user to see the error message.
-func (a *EditProtocolConverterAction) awaitRollout(pcConfig config.ProtocolConverterConfig, desiredPCState string) (string, error) {
+func (a *EditProtocolConverterAction) awaitRollout(previousConfig config.ProtocolConverterConfig, newConfig config.ProtocolConverterConfig, desiredPCState string) (string, error) {
```
FalconTube: here `pc` is not `previousConfig` if you thought that this was the acronym for it. 
It is the `currently` deployed one.<br><br>If we decide to change the name off of `pcConfig`, then we should rather call it `currentConfig` or `currentPCConfig` (current protocol converter config)

umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -1137,6 +1142,91 @@ func (a *EditProtocolConverterAction) mergeUserVariables(base map[string]any, in
...
+
+// connectionCheckWait reports what the connection check is still waiting for,
+// or "" when the scan agrees with the endpoint this edit persisted.
+//
+// The endpoint is rendered from the persisted spec, never taken from the action
+// payload: get-protocolconverter hands back the raw connection template when
+// the spec carries no IP/PORT variables, so a payload can name an unresolved
+// target on port 0 that no scan will ever report.
+func (a *EditProtocolConverterAction) connectionCheckWait(
+	newConfig config.ProtocolConverterConfig,
```
FalconTube: this should just be named `config` here, since this function does not know about `old` and `new`

FalconTube (changes requested):
1. As discussed in meeting, this part of the stack DOES NOT solve the "connection only change is misbehaving part". But since part 2 of the stack mostly does it (I found a bug there, so needs fixing on there), this part of the stack here should be fine
2. Noted a few things, but nothing major. Biggest one is the one that CR also found, needs detailed inspection. Known states here in the fsm are very hard to follow

umh-core/pkg/communicator/actions/edit-protocolconverter.go
```diff
@@ -1137,6 +1142,91 @@ func (a *EditProtocolConverterAction) mergeUserVariables(base map[string]any, in
...
+	lastScan := nmapObs.ServiceInfo.NmapStatus.LastScan
+	if lastScan == nil || !lastScan.Timestamp.After(a.persistedAt) {
+		return "waiting for a scan of " + wantEndpoint + " taken after the edit"
+	}
+
+	// Do not use IsRunning instead: it means the port answered on fsmv2 but only
+	// "the scanner process is up" on fsmv1, where a refused port reads true.
+	if lastScan.PortResult.State != string(nmapservice.PortStateOpen) {
+		return "waiting for nmap to report " + wantEndpoint + " open (last scan: " + lastScan.PortResult.State + ")"
+	}
```
FalconTube: Also, let's adjust the messages here.
They should not state `nmap` any more + I find them a bit too verbose.
This is exposed to the user a lot, so might make sense to have this very user centric :)

umh-core/pkg/communicator/actions/edit-protocolconverter.go
Aaron99B: renamed it to rollbackConfig

umh-core/pkg/communicator/actions/edit-protocolconverter.go
Aaron99B: changed it to pcConfig. config would shadow the imported package

umh-core/pkg/communicator/actions/edit-protocolconverter.go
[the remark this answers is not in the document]
Aaron99B: adapted this

umh-core/pkg/communicator/actions/edit-protocolconverter.go
Aaron99B: rewrote it

umh-core/pkg/fsm/nmap/models.go
```diff
@@ -35,7 +35,7 @@ const (
 	OperationalStateDegraded = "degraded"
 
 	// Running phase states
-	// See nmap-states here: https://nmap.org/book/man-port-scanning-basics.html
+	// See nmap-states here: https://nmapservice.org/book/man-port-scanning-basics.html
```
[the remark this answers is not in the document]
Aaron99B: restored it

umh-core/pkg/service/nmap/types.go
```diff
@@ -82,6 +82,17 @@ type PortResult struct {
 	Port uint16 `json:"port"`
 }
 
+type PortState string
```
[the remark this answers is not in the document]
Aaron99B: documented it

umh-core/CHANGELOG.md
Aaron99B: excluded it

umh-core/CHANGELOG.md
Aaron99B: shortened it

umh-core/pkg/communicator/actions/edit-protocolconverter_awaitrollout_test.go
```diff
@@ -0,0 +1,377 @@
...
+							},
+							ServiceInfo: nmapservice.ServiceInfo{
+								NmapStatus: nmapservice.NmapServiceInfo{
+									IsRunning: true,
+									LastScan: &nmapservice.NmapScanResult{
+										PortResult: nmapservice.PortResult{
+											State: string(nmapservice.PortStateClosed),
+											Port:  port,
+										},
+									},
```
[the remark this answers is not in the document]
Aaron99B: addressed it

FalconTube (approved):
Nice, looks good now!

PR #2710, opened 2026-08-18
fix(topicbrowser): stop sending topic metadata

JeremyTheocharis:
Ticket: [ENG-5679](https://linear.app/united-manufacturing-hub/issue/ENG-5679/umh-core-stop-sending-topicinfometadata).

## Problem

umh-core sends a map of Kafka headers on every topic, in every bundle, to every subscribed browser tab. On instances with many topics that map is most of the payload. The console does not read it: ManagementConsole#3456 removed the read and has merged.

A bundle reaches the console two ways, and both carried the map. See `GetSubscriberData` and `SubscriberData.UnsBundles`:

* A tab that has just connected gets bundle 0, rebuilt from umh-core's internal topic cache. Once per connection.
* Every poll after that gets bundles 1+, the payloads queued as they arrive from the ring buffer. This is almost all of the traffic.

Clearing the map at ingest cleans bundle 0, because the cache is built from the parsed entries. It does nothing for bundles 1+, which were forwarded as the original bytes. That is why there are two commits here.

Shrinking the payload is part of getting the Topic Browser to update once a second rather than once every ten seconds ([ENG-5525](https://linear.app/united-manufacturing-hub/issue/ENG-5525/umh-core-re-sends-the-full-topic-browser-cache-every-10s-and-never), console side [ENG-5652](https://linear.app/united-manufacturing-hub/issue/ENG-5652/topic-browser-stop-storing-topicinfometadata-on-every-topic-row)).

## What we're shipping

* Metadata is cleared at ingest, before an entry enters the internal cache. That covers bundle 0.
* The queued payload is re-encoded from the cleared entries. That covers bundles 1+. `updateInternalCache` returns that value now, and is renamed `ingestBuffer` because it no longer only updates the cache.
* The queued value has its own type, `strippedBundle`. A ring-buffer item cannot be queued by mistake: it does not compile there. Queueing one is the defect being fixed.
* `getUnsentBundles` gives the queued path a name. `GetSubscriberData` now shows both sources side by side, so a change to one is visibly not a change to the other.
* Re-encoding costs one marshal per buffer at ingest, so the cost does not grow with the number of subscribers.
* The queue no longer returns items to the ring buffer's pool. It no longer holds them, and returning a snapshot item was never safe, because the ring buffer keeps its own reference. This was the last production caller of `PutBufferItems`.

## What we're NOT shipping

* **Not touching the GraphQL path.** The cache behind GraphQL's `Topic.metadata` keeps the map, so this change does not affect that surface.
* **Not changing the ring buffer or its pool.** Only the communicator's queue stops holding pool-issued items.
* **Fixing the source.** benthos-umh still puts the map into every emission; umh-core now drops it on arrival. Stopping it there needs a plugin release and an upgrade on every instance ([ENG-5653](https://linear.app/united-manufacturing-hub/issue/ENG-5653/topic-browser-re-sends-every-topic-it-has-ever-seen-in-every-emitted)).

## Confidence

* Analysed against actual customer instances: around 60% reduction in size.
* The changes cause not much overhead. In a test, around 0.012% of a core more. GC gets cheaper rather than dearer, because less is retained.

### DIFF
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Improvements
+
+- Topic browser payloads sent to the Management Console no longer carry the per-topic metadata map, which duplicated the Kafka headers already attached to every event
+
 ## [0.44.35]
 
 ### Improvements
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Improvements
 
-- Topic browser payloads sent to the Management Console no longer carry the per-topic metadata map, which duplicated the Kafka headers already attached to every event
+- Topic Browser updates are around 60% smaller on instances with many topics, because per-topic metadata is no longer resent with every refresh. This helps most on slow or metered connections
 
 ## [0.44.35]
 
--- a/umh-core/pkg/communicator/topicbrowser/cache.go
+++ b/umh-core/pkg/communicator/topicbrowser/cache.go
@@ -247,8 +247,9 @@ func (c *Cache) processBuffer(buf *topicbrowserservice.BufferItem, log *zap.Suga
 
 	// upsert the uns map
 	for _, entry := range ub.GetUnsMap().GetEntries() {
-		// generate a hash from the entry by calling HashUNSTableEntry
 		hash := HashUNSTableEntry(entry)
+		// Metadata stays populated: Topic.metadata in graphql/schema.graphqls
+		// resolves from this map.
 		c.unsMap.Entries[hash] = entry
 	}
 
--- a/umh-core/pkg/communicator/topicbrowser/communicator.go
+++ b/umh-core/pkg/communicator/topicbrowser/communicator.go
@@ -337,9 +337,13 @@ func (tbc *TopicBrowserCommunicator) updateInternalCache(buf *topicbrowserservic
 		}
 	}
 
-	// Update topic map
+	// Update topic map, stripping metadata before storing to both the internal
+	// cache and (via getCacheBundle) the bootstrap wire.
 	for _, entry := range ub.GetUnsMap().GetEntries() {
+		// entry aliases freshly unmarshaled ub; mutating in place is safe and
+		// avoids a deep copy per buffer on the hot ingest loop.
 		hash := HashUNSTableEntry(entry)
+		entry.Metadata = nil
 		tbc.unsMap.Entries[hash] = entry
 	}
 
--- a/umh-core/pkg/communicator/topicbrowser/communicator.go
+++ b/umh-core/pkg/communicator/topicbrowser/communicator.go
@@ -350,15 +350,12 @@ func (tbc *TopicBrowserCommunicator) ingestBuffer(buf *topicbrowserservice.Buffe
 		}
 	}
 
-	// Metadata is cleared before the entry is stored, so nothing built from these
-	// entries carries it: not the internal cache, and not the bundle re-encoded
-	// below.
+	// Nothing built from this bundle carries metadata afterwards: not the
+	// internal cache below, and not the payload encoded from it.
+	removeMetadata(&ub)
+
 	for _, entry := range ub.GetUnsMap().GetEntries() {
-		// entry points into the ub unmarshaled just above, so nothing else holds
-		// it yet and clearing Metadata in place is safe.
-		hash := HashUNSTableEntry(entry)
-		entry.Metadata = nil
-		tbc.unsMap.Entries[hash] = entry
+		tbc.unsMap.Entries[HashUNSTableEntry(entry)] = entry
 	}
 
 	// Encoding on ingest costs one marshal per buffer, so the cost is
@@ -461,6 +458,18 @@ func (tbc *TopicBrowserCommunicator) cleanupOldPendingBuffers() {
 	tbc.pendingToSend = filtered
 
 	tbc.logger.Debugf("Cleaned up %d old pending buffers", dropped)
+}
+
+// removeMetadata clears the per-topic metadata map on every topic in the bundle.
+// The console reads none of it, and it is the bulk of what a topic costs to
+// send.
+//
+// The entries belong to a bundle the caller has just unmarshaled, so nothing
+// else holds them yet and clearing in place is safe.
+func removeMetadata(ub *tbproto.UnsBundle) {
+	for _, entry := range ub.GetUnsMap().GetEntries() {
+		entry.Metadata = nil
+	}
 }
 
 // unsentBundles returns the queued bundles a subscriber has not received yet, in
--- a/umh-core/pkg/communicator/topicbrowser/communicator.go
+++ b/umh-core/pkg/communicator/topicbrowser/communicator.go
@@ -69,15 +69,18 @@ type SubscriberData struct {
 	TopicCount      int            // Current number of topics in cache
 }
 
-// strippedBundle is one UnsBundle re-encoded by this package with the per-topic
-// metadata map removed.
+// queuedBundle is a bundle this package has encoded and is holding for
+// delivery; pendingToSend is a slice of these.
 //
-// It is deliberately not a topicbrowserservice.BufferItem. A ring-buffer item's
-// payload still carries the metadata, and its memory belongs to the ring buffer,
-// which keeps its own reference after handing out a snapshot. See the
-// Data & Ownership Flow section of the pkg/service/topicbrowser package doc.
-// Keeping the types apart means a ring-buffer item cannot reach pendingToSend.
-type strippedBundle struct {
+// It is deliberately not a topicbrowserservice.BufferItem. This package owns a
+// queuedBundle's bytes, while a ring-buffer item's belong to the ring buffer,
+// which keeps its own reference after handing out a snapshot. Keeping the types
+// apart means a ring-buffer item cannot reach pendingToSend. See the Data &
+// Ownership Flow section of the pkg/service/topicbrowser package doc.
+//
+// The per-topic metadata map is dropped on the way in, so the payload differs
+// from the ring-buffer item it came from.
+type queuedBundle struct {
 	Timestamp time.Time // ingest timestamp, copied from the ring-buffer item
 	Payload   []byte    // re-encoded UnsBundle, allocated here
 }
@@ -96,8 +99,8 @@ type TopicBrowserCommunicator struct {
 	logger    *zap.SugaredLogger // Component-specific logging
 
 	// 📡 COMMUNICATION STATE: Subscriber and delivery management
-	pendingToSend         []strippedBundle // Bundles not yet sent to subscribers
-	lastProcessedSequence uint64           // Last processed buffer sequence number
+	pendingToSend         []queuedBundle // Bundles not yet sent to subscribers
+	lastProcessedSequence uint64         // Last processed buffer sequence number
 
 	// 🔧 CONFIGURATION
 	maxPendingBuffers int // Cleanup threshold for old pending buffers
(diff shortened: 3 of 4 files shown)


umh-core/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Improvements
+
+- Topic browser payloads sent to the Management Console no longer carry the per-topic metadata map, which duplicated the Kafka headers already attached to every event
```
JeremyTheocharis: focus more on the user impact there, e.g., reduced required bandwith by x % or so by removing unnecessaery data from the protcol.

JeremyTheocharis: → better user experiences in bandwith constrained environments

umh-core/pkg/communicator/topicbrowser/communicator.go
```diff
@@ -337,13 +354,33 @@ func (tbc *TopicBrowserCommunicator) updateInternalCache(buf *topicbrowserservic
...
 	for _, entry := range ub.GetUnsMap().GetEntries() {
+		// entry aliases freshly unmarshaled ub; mutating in place is safe and
+		// avoids a deep copy per buffer on the hot ingest loop.
 		hash := HashUNSTableEntry(entry)
+		entry.Metadata = nil
 		tbc.unsMap.Entries[hash] = entry
 	}
 
-	return nil
+	// Re-encode here rather than in GetSubscriberData, which runs once per
```
JeremyTheocharis: weird comment, better to explain more the architectural decision and not say "instead of somewhere else" as somewher else is time dpedentent

umh-core/pkg/communicator/topicbrowser/cache.go
```diff
@@ -248,6 +248,10 @@ func (c *Cache) processBuffer(buf *topicbrowserservice.BufferItem, log *zap.Suga
 	// upsert the uns map
 	for _, entry := range ub.GetUnsMap().GetEntries() {
 		// generate a hash from the entry by calling HashUNSTableEntry
+		// Metadata is deliberately kept here, unlike the communicator's copy of
```
JeremyTheocharis: this feels weird here as a comment? feels out of place / could be solved somehow with adjusting code?

umh-core/pkg/communicator/topicbrowser/communicator.go
```diff
@@ -408,32 +439,40 @@ func (tbc *TopicBrowserCommunicator) MarkDataAsSent(timestamp time.Time) {
...
 		}
 	}
 
+	dropped := len(tbc.pendingToSend) - len(filtered)
+
 	tbc.pendingToSend = filtered
 
-	// Return old buffers to pool
-	if len(buffersToReturn) > 0 {
-		topicbrowserservice.PutBufferItems(buffersToReturn)
```
Aaron99B, line 434: We can now remove this function in umh-core/pkg/service/topicbrowser/buffer.go:126 as it is no longer used

umh-core/pkg/communicator/topicbrowser/communicator.go
```diff
@@ -69,6 +69,19 @@ type SubscriberData struct {
...
 
+// strippedBundle is one UnsBundle re-encoded by this package with the per-topic
+// metadata map removed.
+//
+// It is deliberately not a topicbrowserservice.BufferItem. A ring-buffer item's
+// payload still carries the metadata, and its memory belongs to the ring buffer,
+// which keeps its own reference after handing out a snapshot. See the
+// Data & Ownership Flow section of the pkg/service/topicbrowser package doc.
+// Keeping the types apart means a ring-buffer item cannot reach pendingToSend.
+type strippedBundle struct {
```
Aaron99B: Regarding the name: Will it still be explanatory in the future when people do not know anymore that once the meta data was stripped from it?

umh-core/pkg/communicator/topicbrowser/communicator.go
```diff
@@ -337,13 +347,36 @@ func (tbc *TopicBrowserCommunicator) updateInternalCache(buf *topicbrowserservic
 		}
 	}
 
-	// Update topic map
+	// Metadata is cleared before the entry is stored, so nothing built from these
+	// entries carries it: not the internal cache, and not the bundle re-encoded
+	// below.
 	for _, entry := range ub.GetUnsMap().GetEntries() {
```
Aaron99B, line 361: Can we move this part in a removeMetadata function so it is obvious that the meta data is removed?

Aaron99B (changes requested):
We can now also remove the bufferItemPool variable in the buffer.go file, but I would propose to handle this in a follow-up ticket, but link this ticket here

umh-core/pkg/communicator/topicbrowser/communicator.go
JeremyTheocharis: Done, in #2712 rather than here, since that PR already edits `buffer.go`. The function is gone along with the package-doc clause telling every `GetSnapshot` consumer to call it, and the two specs that existed only to exercise it.

One correction on the follow-up you suggested: `bufferItemPool` itself cannot be removed. `parseBlock` still takes an item from it (`parse.go:262`) and returns one on the hex-decode error path (`parse.go:275`), so dropping the variable would not compile. What ended is recycling *after* the ring buffer has an item. If a follow-up ticket is still worth writing, the accurate scope is that the pool provides almost no reuse, because only the error path ever returns an item.

umh-core/pkg/communicator/topicbrowser/communicator.go
JeremyTheocharis: Renamed to `queuedBundle`. Your question was the right one: the old name described something that happened to the bundle once, rather than what the bundle is, and stripping stops being interesting as soon as nobody remembers it used to be there.

The type doc now leads with its job and with the ownership rule that is the reason the type exists at all, and mentions the metadata last.

umh-core/pkg/communicator/topicbrowser/communicator.go
JeremyTheocharis: Extracted as `removeMetadata`, called as a single statement on the bundle, so the loop below it does nothing but fill the cache.

Worth stating the cost: it walks the topic entries twice per buffer instead of once. The second walk only nils a pointer field, next to a `proto.Marshal` of the same bundle, so it does not move the ingest cost this PR measures.

umh-core/pkg/communicator/topicbrowser/communicator.go
```diff
@@ -371,22 +406,18 @@ func (tbc *TopicBrowserCommunicator) GetSubscriberData(isBootstrapped bool) (*Su
...
+	unsent := tbc.unsentBundles()
 
-	for _, buf := range tbc.pendingToSend {
-		if buf.Timestamp.After(tbc.lastSentTimestamp) {
-			data.UnsBundles[index] = buf.Payload
-			index++
-			incrementalCount++
+	for _, bundle := range unsent {
+		data.UnsBundles[index] = bundle.Payload
+		index++
```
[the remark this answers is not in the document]
JeremyTheocharis: Not taking this one. The hazard is real in principle, but the only consumer, `generator/topicbrowser.go:83`, passes the map straight to the encoder and writes nothing.

Copying per subscriber would reintroduce the cost this path exists to avoid: the bundle is encoded once on ingest precisely so delivery does not scale with subscriber count. The read-only contract is now stated on the `UnsBundles` field instead, naming why positions 1 and up are shared with `pendingToSend`.

umh-core/pkg/communicator/topicbrowser/communicator_metadata_strip_test.go
```diff
@@ -0,0 +1,193 @@
...
+		// the console cannot rebuild the topic map.
+		preserved := false
+		for _, entry := range wiredEntries {
+			if entry.GetName() == "temperature" && entry.GetLevel0() == "corpA" && entry.GetDataContract() == "_historian" {
+				preserved = true
+
+				break
+			}
+		}
+		Expect(preserved).To(BeTrue(), "non-metadata TopicInfo fields must survive the strip")
```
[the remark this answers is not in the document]
JeremyTheocharis: I could not reproduce this. `golangci-lint run` with the repo config, at 2.5.0 with `wsl_v5` enabled and no config warnings, reports 0 issues on this package.

I checked that the file is genuinely analysed rather than skipped, by injecting a violation: `intrange` flagged it at line 67 of this same file. So the clean result is real. You cite 2.12.2, so this looks version-specific.

The stated consequence is also not right: no CI workflow in this repo runs golangci-lint, so this is not failing CI. Leaving the whitespace as is.

umh-core/pkg/communicator/topicbrowser/communicator.go
```diff
@@ -63,10 +63,34 @@ func validateBufferSizeFromSnapshot(snapshot topicbrowserservice.RingBufferSnaps
...
-	TopicCount      int            // Current number of topics in cache
+	LatestTimestamp time.Time // Latest timestamp in this subscriber batch
+
+	// UnsBundles holds ready-to-send payloads indexed by position: 0 is the
+	// whole cache, 1 and up are queued bundles.
+	//
+	// A caller must not write to these bytes. Positions 1 and up are the same
+	// slices pendingToSend holds, so a write here would reach every later
+	// subscriber too.
+	UnsBundles map[int][]byte
```
Aaron99B, line 70: for me these things were more readable before when there were less comments in the SubscriberData struct

umh-core/pkg/communicator/topicbrowser/communicator.go
```diff
@@ -408,34 +447,54 @@ func (tbc *TopicBrowserCommunicator) MarkDataAsSent(timestamp time.Time) {
...
+func removeMetadata(ub *tbproto.UnsBundle) {
+	for _, entry := range ub.GetUnsMap().GetEntries() {
+		entry.Metadata = nil
 	}
 }
 
+// unsentBundles returns the queued bundles a subscriber has not received yet, in
+// ingest order. It reads pendingToSend and lastSentTimestamp without locking;
+// GetSubscriberData holds the read lock, as it does for getCacheBundle.
+func (tbc *TopicBrowserCommunicator) unsentBundles() []queuedBundle {
```
Aaron99B: I would still call this after an action e.g. getUnsentBundles

Aaron99B (approved):

umh-core/pkg/communicator/topicbrowser/communicator.go
JeremyTheocharis: also see getCacheBundle also does it, so lets do it here as well

umh-core/pkg/communicator/topicbrowser/communicator.go
JeremyTheocharis: lets cut then

```
// A caller must not write to these bytes. Positions 1 and up are the same
	// slices pendingToSend holds, so a write here would reach every later
	// subscriber too.
```

JeremyTheocharis: Cut in `8af7464ce`. `UnsBundles` keeps only the line saying what the map is: 0 is the whole cache, 1 and up are queued bundles.

Flagging what goes with it, since that text was written to answer CodeRabbit's aliasing finding on this same field. The aliasing is unchanged — positions 1 and up are still the slices `pendingToSend` holds — only the written warning is gone. What is left is on `queuedBundle`, which says this package owns the payload bytes; that explains where the sharing comes from but does not tell a caller not to write. The single caller, `generator/topicbrowser.go:83`, hands the map to the encoder and writes nothing.

umh-core/pkg/communicator/topicbrowser/communicator.go
JeremyTheocharis: Renamed to `getUnsentBundles` in `8af7464ce`. The two selectors `GetSubscriberData` reads from — `getCacheBundle` and `getUnsentBundles` — now name themselves the same way.

The sentence above the call said "whatever `unsentBundles` has queued", which read wrong once the name started with a verb, since the function selects rather than queues. It now says it returns the bundles queued since the subscriber's last send.

JeremyTheocharis:
Filed as [ENG-5764](<https://linear.app/united-manufacturing-hub/issue/ENG-5764/topic-browser-bufferitempool-now-recycles-almost-nothing>), under ENG-5525, `Ideas`, Low.

Scoped to what is actually true rather than "remove the variable". `bufferItemPool` cannot be deleted as it stands: `parseBlock` takes an item from it at `parse.go:262`, so dropping it would not compile. What ended with #2712 is recycling *after* the ring buffer has an item. The only remaining `Put` is `parse.go:275`, on the hex-decode error path, so on the normal path nothing is ever returned and every block allocates a fresh shell.

The ticket names two ways out: swap the `Get` for a plain `&BufferItem{}` and delete the pool, or give the pool a return site that is safe under the ring buffer's ownership rule — there is no such site today, which is why #2712 removed the old one.

PR #2711, opened 2026-08-18
fix(topicbrowser): stop recycling buffer items the ring still owns

JeremyTheocharis:
Part of the **ENG-5525** stack — see that ticket for the whole arc, and for what the other three PRs do.

## Problem

The communicator returns buffer items to a shared pool it never took them from, and pooling zeroes each
struct in place. Until the previous PR in this stack those items were the ring buffer's own, so past 100
pending entries the ring lost payloads it was still serving.

## What we are shipping

- The communicator no longer pools buffer items.
- The comments claiming the consumer owns and recycles these items now say the opposite.

## What we are NOT shipping

- Retiring `PutBufferItems` or `bufferItemPool`. The function and the package-doc clause telling every
  consumer to call it both stay, and the pool now only receives items from the hex-decode error path.
  Safe and useless beats fast and corrupting. Ticketed.
- Deleting the deep-copy override. The call breaks the ownership rule, and the override also covers the
  simulator path, which never copies.
- Any change to what subscribers receive.

### DIFF
--- a/umh-core/pkg/communicator/topicbrowser/pool_ownership_test.go
+++ b/umh-core/pkg/communicator/topicbrowser/pool_ownership_test.go
@@ -1,0 +1,89 @@
+// Copyright 2025 UMH Systems GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package topicbrowser_test
+
+import (
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"go.uber.org/zap"
+	"google.golang.org/protobuf/proto"
+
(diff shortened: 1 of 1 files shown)


JeremyTheocharis:
Closing without merging. The change this PR described has already landed, and what was left behind does not carry its weight.

**The production fix is in #2710.** `afe894e71` deletes the `PutBufferItems` call from `cleanupOldPendingBuffers`. That was forced rather than chosen: #2710 changes `pendingToSend` from `[]*topicbrowserservice.BufferItem` to `[]strippedBundle`, and once the slice stops holding `*BufferItem` the pooling call no longer compiles. The type change and the deletion had to land in one commit.

**What remained here was one test file, and it has no reachable failure mode on the current base.** Two independent reasons:

1. After #2710 nothing holds a `*BufferItem` past the ingest call, so the field the test watches cannot be reassigned. The invariant is true by construction rather than by assertion.
2. The fixture compares the payload against itself. It does `original[i] = payload` and then hands that same slice to `rb.Add(&BufferItem{Payload: payload})`, so the two share one backing array and an in-place overwrite is undetectable.

A commit message on `eng-3591-delivered-watermark` claimed this spec failed again once the pooling bug was restored. That was verified on 2026-08-18, two days before the type change was authored, so it was true when written and is not evidence about this base.

**Coverage after this closes.** The ownership rule is now expressed by the type in #2710. Delivery behaviour is covered by `delivered_watermark_test.go` in #2713, which asserts both that a buffer arriving mid-send is still delivered and that it is not delivered twice.

**One gap, recorded rather than dropped.** This was the only test that drove `cleanupOldPendingBuffers` past its 100-entry threshold, although it asserted nothing about that branch. A test that asserts what cleanup keeps and what it drops belongs with the pending-buffer bound in ENG-5651.

#2712 and #2713 are restacked onto #2710 and remain green.

PR #2716, opened 2026-08-18
fix(edit-pc): run the deploy connection check for every bridge, not only flowless ones

JeremyTheocharis:
Middle of the stack: #2708 → **#2716** → #2751. Umbrella: ENG-5853.

## Problem

The connection check #2708 adds sits inside the empty-DFC branch. A bridge that already has a read or write flow never reaches it, so its edit is accepted on the bridge's own pre-edit state without the requested endpoint ever being dialled. That is the more common bridge.

## What we are shipping

- The check runs for every bridge, not only a flowless one. An edit to a bridge with a read flow now waits for the requested endpoint to be scanned and found open.
- It runs before the DFC config comparison, so a bridge whose target was never dialled is reported as a connection problem rather than a dataflow component one.
- An edit that carries no connection block is exempt, so a bridge whose target is unreachable stays editable, including the edit that stops its flows.

Five specs: a positive control that a healthy read-DFC edit is still accepted on the first tick, the read-path acceptance reproduced in-process (the deterministic form of a race the container suite reproduces intermittently), a placement guard that fails if the check moves back below the DFC comparison, the exemption, and the filtered-port spec moved along with the check.

## What we are NOT shipping

- Whether the exemption fires at all for a console edit. `get-protocolconverter` fills `Connection.IP` from the deployed spec and the wizard round-trips it, so a console payload normally carries a connection block and the exemption may be unreachable. ENG-5858.
- Any framework work. Nothing here touches `fsmv2client`, the adapter or the CSE store.

## Ticket

This branch is named for ENG-5580, but ENG-5580 is the FSMv2 config-identity/freshness ticket and lists "the consumer-side gate holes" under what it is *not* shipping. This PR is one of those gate holes, so it needs a ticket of its own, and merging it will not close ENG-5580.

### DIFF
--- a/umh-core/pkg/communicator/actions/edit-protocolconverter.go
+++ b/umh-core/pkg/communicator/actions/edit-protocolconverter.go
@@ -643,27 +643,32 @@ func (a *EditProtocolConverterAction) awaitRollout(
 				found = true
 				currentStateReason = "current state: " + instance.CurrentState
 
-				if a.dfcType == DFCTypeEmpty {
-					// Unreachable while applyMutation forces the bridge active.
-					// Kept because a stopped bridge stops its nmap service and
-					// would never report the new port.
-					if desiredPCState != protocolconverter.OperationalStateStopped {
-						if waitingFor := a.connectionCheckWait(rolloutConfig, pcSnapshot); waitingFor != "" {
-							currentStateReason = waitingFor
-							SendActionReply(
-								a.instanceUUID,
-								a.userEmail,
-								a.actionUUID,
-								models.ActionExecuting,
-								RemainingPrefixSec(remainingSeconds)+currentStateReason,
-								a.outboundChannel,
-								models.EditProtocolConverter,
-							)
+				// Runs before the DFC config comparison below, so a bridge whose target
+				// was never dialed is reported as a connection problem rather than a
+				// dataflow component one.
+				//
+				// An edit that carries no connection is exempt, so a bridge whose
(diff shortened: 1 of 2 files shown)


umh-core/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Editing a bridge's connection to a port that does not answer now fails and restores the previous configuration, instead of reporting a successful deploy. The deploy waits for a scan that finds the new port open, so a port that is closed or filtered is no longer accepted
 - Repointing a bridge to a different host on the same port is now checked too. The deploy waits for a scan taken after the change, so the previous host's scan can no longer satisfy it
+- Editing a bridge that already has a read or write flow now verifies the new connection as well, not only a bridge without flows. Such an edit previously reported success as soon as the flow was applied, even when the new address never answered
```
FalconTube: fsmv2 only, don't mention it here since we will later just switch to it under the hood

FalconTube (commented):
I can now consistently hit the issue, that a "known good" connection will not report as such i.e. the connection does not come online properly.<br><br>So now the "other way around" does not work reliably 😀 <br><br>However, it might not be an issue of this PR here, but could also be related to the "other stuff is red as well" bug.
I still need to investigate

FalconTube (approved):
Ok, was a problem on my side. Fix for nmap fsmv2 works fine with this PR here.
Just remove the changelog entry, then good to go

FalconTube (approved):
Tested it live and works perfectly!

PR #2718, opened 2026-08-19
pkg/diagnosis: fix the vocabulary, and let a signal narrow its own answer

JeremyTheocharis:
Ticket: [ENG-5128](https://linear.app/united-manufacturing-hub/issue/ENG-5128/expose-p95p99-cpu-metrics-and-persist-degraded-state)

## Problem

While reviewing the CPU health PR I could not follow a code comment. I went digging and found a bad abstraction under it.

`Track` is a weird word. Nothing says how a track differs from an instrument. Tracks exist only for attribution in the cpuhealth implementation.

But attribution belongs in the signal instead, and it should be a tree. If overall CPU is fine, everything is good. If it fires, we are degraded and do not yet know who to blame. Only then is it worth looking one level down, where a firing sub-signal turns "CPU degraded" into "CPU degraded, because the container is taking a lot of CPU".

## What we are shipping

* An "instrument" is now a "measurement" plus the "marks" that judge it. A number with no marks is just a "measurement".
* "track" is removed. If you want to just record data without alerting on it, then just use a "measurement"
* A signal declares its own "attribution" instead of being something done at the end
* A signal can declare sub-signals, called "refinements". If the parent signal fires, the engine will check all the refinements for additional information of where this is coming from.
* If the "measurement" that made a signal fire can no longer be read, the signal now switches to one it can still read and judges on that one's marks. Without this it stays degraded until the original measurement comes back, which can be forever.

## What we are NOT shipping

* **No CPU vocabulary.**
* **No consumer changes.**

### DIFF
--- a/umh-core/pkg/diagnosis/declare.go
+++ b/umh-core/pkg/diagnosis/declare.go
@@ -89,6 +89,10 @@ func (m Measurement[S]) Read(s S) (value, against Reading) {
 type Signal[S any] struct {
 	Name        string
 	Instruments []Instrument[S]
+	// Refinements are signals asked only when this one fires, narrowing its
+	// answer: each is structurally a signal with the same instruments and
+	// marks, but it is never asked unless the parent fires.
+	Refinements []Signal[S]
 	// DemoteSpan is how long a signal may go unread before its window empties:
 	// stale detection, so an old number cannot stand forever.
 	DemoteSpan time.Duration
--- a/umh-core/pkg/diagnosis/declare.go
+++ b/umh-core/pkg/diagnosis/declare.go
@@ -45,20 +45,18 @@ func (e Environment) Has(c Capability) bool {
 	return e.caps[c]
 }
 
-// Instrument is one way of measuring a signal: what to read, over how long a
-// window, under which reduction, against which thresholds.
-type Instrument[S any] struct {
+// Measurement is one way of measuring a signal: what to read, over how long a
+// window, under which reduction.
+type Measurement[S any] struct {
+	Name string
 	// Extract reads the value from a snapshot, the numerator under DeltaRatio.
 	Extract func(S) Reading
 	// Against reads the DENOMINATOR of a ratio: DeltaRatio divides the delta of
 	// Extract's counter by the delta of this one. Nil for a single series.
 	Against   func(S) Reading
-	Name      string
-	Requires  []Capability
+	Span      time.Duration
 	Reduction Reduction
-	// Marks are the thresholds this instrument's number is judged against.
-	Marks Marks
-	Span  time.Duration
+	Requires  []Capability
 	// Boolean says the series is zero or one and nothing between.
 	Boolean bool
 	// Counter says both series are monotone counters, so a backwards step means
@@ -67,14 +65,22 @@ type Instrument[S any] struct {
 	Counter bool
 }
 
+// Instrument is one way of measuring a signal: a measurement plus the marks its
+// number is judged against.
+type Instrument[S any] struct {
+	Measurement[S]
+	// Marks are the thresholds this instrument's number is judged against.
+	Marks Marks
+}
+
 // Read applies both extractors to one snapshot, so both counters of a delta
 // ratio come off the same instant. With no Against it hands back an absence.
-func (i Instrument[S]) Read(s S) (value, against Reading) {
-	if i.Against == nil {
-		return i.Extract(s), Unknown()
+func (m Measurement[S]) Read(s S) (value, against Reading) {
+	if m.Against == nil {
+		return m.Extract(s), Unknown()
 	}
 
-	return i.Extract(s), i.Against(s)
+	return m.Extract(s), m.Against(s)
 }
 
 // Signal is a QUESTION about the resource, such as whether the CPU is saturated.
--- a/umh-core/pkg/diagnosis/declare.go
+++ b/umh-core/pkg/diagnosis/declare.go
@@ -45,8 +45,10 @@ func (e Environment) Has(c Capability) bool {
 	return e.caps[c]
 }
 
-// Measurement is one way of measuring a signal: what to read, over how long a
-// window, under which reduction.
+// Measurement is a number sampled over time: what to read, over how long a
+// window, under which reduction. It carries no thresholds, so nothing about a
+// measurement can answer a signal. Its number is judged only where an
+// Instrument pairs it with marks.
 type Measurement[S any] struct {
 	Name string
 	// Extract reads the value from a snapshot, the numerator under DeltaRatio.
@@ -65,7 +67,7 @@ type Measurement[S any] struct {
 	Counter bool
 }
 
-// Instrument is one way of measuring a signal: a measurement plus the marks its
+// Instrument is one way of answering a signal: a measurement plus the marks its
 // number is judged against.
 type Instrument[S any] struct {
 	Measurement[S]
@@ -89,12 +91,10 @@ type Signal[S any] struct {
 	// Instruments each answer the signal differently. The order is significant:
 	// the first instrument that can supply a number is the one used.
 	Instruments []Instrument[S]
-	// Refinements are signals that narrow this one's answer, with their own
-	// instruments and marks. A refinement is sampled AND judged every tick,
-	// whether or not this signal fired, so its window fills and its verdict is
-	// reached independently of when this signal fires. Only the report waits on
-	// this signal: a refinement appears in Fired.Refinements under a signal that
-	// fired, and never as a verdict of its own.
+	// Refinements are signals declared under this one, narrowing its answer with
+	// their own instruments and marks. A refinement appears in Fired.Refinements
+	// under a parent that fired, and never as a verdict of its own. The
+	// Refinements section of the package doc works through an example.
 	Refinements []Signal[S]
 	// DemoteSpan is how long a signal may go unread before its window empties:
 	// stale detection, so an old number cannot stand forever.
--- a/umh-core/pkg/diagnosis/declare.go
+++ b/umh-core/pkg/diagnosis/declare.go
@@ -84,14 +84,15 @@ func (m Measurement[S]) Read(s S) (value, against Reading) {
 }
 
 // Signal is a QUESTION about the resource, such as whether the CPU is saturated.
-// Each of its Instruments answers it differently. The order is significant: the
-// first instrument that can supply a number is the one used.
 type Signal[S any] struct {
-	Name        string
+	Name string
+	// Instruments each answer the signal differently. The order is significant:
+	// the first instrument that can supply a number is the one used.
 	Instruments []Instrument[S]
-	// Refinements are signals asked only when this one fires, narrowing its
-	// answer: each is structurally a signal with the same instruments and
-	// marks, but it is never asked unless the parent fires.
+	// Refinements are signals that narrow this one's answer, with their own
+	// instruments and marks. A refinement is sampled every tick, whether or not
+	// this signal fired, so its window fills independently of when this signal
+	// fires.
 	Refinements []Signal[S]
 	// DemoteSpan is how long a signal may go unread before its window empties:
 	// stale detection, so an old number cannot stand forever.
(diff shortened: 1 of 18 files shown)


JeremyTheocharis:
@coderabbitai review

umh-core/pkg/diagnosis/judge.go
```diff
@@ -81,6 +90,16 @@ type Fired struct {
 	// Instrument is the instrument that fired, stamped on the fire transition and
 	// never refreshed; a later live winner does not overwrite it.
 	Instrument string
+	// Refinements are the signals hanging under this one that have fired
+	// themselves, each judged against its own marks. A refinement is judged
+	// every tick, whether or not this signal fired, but it is reported only
+	// here, under a signal that did fire.
+	//
+	// They come lowest Tier first, ties going to the one the table declared
```
JeremyTheocharis:
```
	// They come lowest Tier first, ties going to the one the table declared
	// first, at this and every deeper level: the entry at index 0 is the most
	// urgent narrowing of this signal, with no sorting left to the caller. Rank
	// does not reach in here; it orders the top-level set only.
```

→ feels too long of a comment

umh-core/pkg/diagnosis/judge.go
```diff
@@ -81,6 +90,16 @@ type Fired struct {
 	// Instrument is the instrument that fired, stamped on the fire transition and
 	// never refreshed; a later live winner does not overwrite it.
 	Instrument string
+	// Refinements are the signals hanging under this one that have fired
```
JeremyTheocharis: how about using more of my original wording, e.g., fro mthe PR description including an example?

umh-core/pkg/diagnosis/engine.go
```diff
@@ -67,115 +69,181 @@ const (
...
 type signalState[S any] struct {
 	signal Signal[S]
-	latch  Latch
+	// path is what this signal's windows are keyed under: its bare name at the
+	// top level, its parent's path plus its own name for a refinement.
+	path string
+	// latch is the single latch that judges this signal. Every node in the tree
+	// has its own, at every depth.
+	latch Latch
+	// refinements is the same pairing one level down, one entry per refinement
```
JeremyTheocharis: this comment feels too long

JeremyTheocharis:
PR description was a little bit cursed, fixed it now by reverting to a previous version

umh-core/pkg/diagnosis/engine.go
```diff
@@ -159,17 +157,39 @@ func NewEngine[S any](t Table[S]) (*Engine[S], error) {
...
+	}
+
+	for _, r := range s.Refinements {
+		if err := buildWindows(e, path+"/"+r.Name, r); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
JeremyTheocharis: Already fixed in `60a9e945b`, which landed after this scan. `validateSignal` rejects the separator as its first rule, and it runs at both scopes: `NewEngine` calls it once per top-level signal, and it recurses into refinements. Two specs in `engine_newengine_test.go` pin both scopes.

The shared-constant half of the suggestion was still open, and is done in `983b8dab1`: the two path joins, the guard, and its error message all read `pathSeparator`, so the join and the refusal cannot drift apart.

umh-core/pkg/diagnosis/doc.go
```diff
@@ -30,4 +30,23 @@
...
+// any signal in the level below it gives more information. For example, a
+// caller could measure the container's CPU. A refinement checks whether the
+// container CPU is over x% of the CPU currently used; if it fires, the
+// caller's message goes from "CPU degraded" to "CPU degraded, because the
+// container is taking a lot of CPU".
+//
+// There can be another refinement beside it that checks whether the host is
+// over x%, instead of the container's workload, and then it is "CPU degraded,
+// but nothing in our container and it is other software running on the host".
+// If the container metric is not available, both refinements are absent.
```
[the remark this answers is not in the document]
JeremyTheocharis: Not taking this one, though it found a real defect.

The claim was underspecified rather than false. The second refinement reads the CPU in use minus the container's share, so it does need the container metric, and "both refinements are absent" was correct. What the section never said is what that refinement reads — which is exactly why your reading follows from the text. `6b4c04ea5` states the shared dependency, and replaces the two `x%` placeholders with real numbers.

Your mechanism note is right and is why the fix goes this way: `Engine.resolve` reads only the windows keyed by that signal's own path, so availability is per-signal. Two refinements are independent unless they share an input, and here the doc had simply not said that these two do.

umh-core/pkg/diagnosis/engine_episode_test.go
```diff
@@ -0,0 +1,300 @@
...
+	// names lists one level of a fired set by signal name, so a phase can state
+	// the shape of that level in one line.
+	names := func(fired []Fired) []string {
+		out := make([]string, len(fired))
+		for i, f := range fired {
+			out[i] = f.Identity.Signal
+		}
+
+		return out
+	}
```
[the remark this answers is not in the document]
JeremyTheocharis: Split verdict.

**QF1008 — taken**, in `e2ab5f812`. It fires in two files, not one: `engine_refinements_test.go:51` carries the same selector.

**wsl_v5 — not taken.** It does not reproduce at the version this repo pins (`GOLANGCI_LINT_VERSION = v2.5.0` in `umh-core/Makefile`); 2.12.2 is not what runs here. Positive control: at 2.5.0 `wsl_v5` does report findings against other packages, so the zero on `pkg/diagnosis` is a real zero rather than a check that never ran. Separately, seven files in this package carry that blank line after `Describe(`, two of them untouched by this branch, so removing it here would leave this file the only outlier.

One thing the review missed while in this file: at the pinned version golangci-lint also reports two `intrange` findings that this branch introduces. Fixed in `2c7da2cda`.

umh-core/pkg/diagnosis/engine.go
```diff
@@ -197,117 +268,169 @@ func validate[S any](t Table[S]) error {
...
-			}
+// validateSignal applies every rule a Signal is held to, then recurses into its
+// refinements under the refinement's own path, so an error names where in the
+// tree the row lives rather than the bare refinement. Instrument and refinement names are
+// unique among their siblings only.
+func validateSignal[S any](path string, s Signal[S], interval time.Duration) error {
+	// A name holding the separator could compose a path equal to another
+	// signal's, e.g. a top-level signal named "A/X" against the refinement X of
+	// a signal A, and the two would silently share one window. Refusing it in
+	// every segment makes that collision unreachable.
```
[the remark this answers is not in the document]
JeremyTheocharis: Not taking this one. These comments each record a constraint the code cannot show — why the guard exists, and what a reader would otherwise have to rediscover — which `standards/comments.md` treats as the case that earns a comment. The `CLAUDE.md` line you are citing is real, and the two documents genuinely disagree here; that is a standards question rather than something to settle in this diff.

Worth noting the direction of travel: this branch has been reducing comment volume, not adding it. The block at the top of `validateSignal` is shorter than the one it replaces.

umh-core/pkg/diagnosis/judge.go
```diff
@@ -141,30 +156,54 @@ func crossedClear(v float64, m Marks) bool {
...
 
+	// The instrument that fired has stopped answering, so judge on the marks of
+	// one that still does. since is not re-stamped: the condition never stopped,
+	// only what can measure it changed. latch_measurement_switch_test.go works
+	// the three cases through.
+	//
+	// A one-tick gap reaches this block too, because a window goes untrusted on
+	// the first reading it misses rather than when it drains. That is harmless
+	// because a live number reading neither past its clear nor past its fire
+	// leaves the episode exactly as it was, marks and blame included.
```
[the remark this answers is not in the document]
JeremyTheocharis: Not taking this one, for the same reason as the `engine.go` thread.

On the numbers: the inline block in `Update` was 27 lines earlier on this branch and is 9 now. What was cut was the narration and a worked CPU example that `latch_measurement_switch_test.go` already drives as three named specs. What is left states two things the code cannot: that `since` is deliberately not re-stamped, and that a window goes untrusted on the first reading it misses rather than when it drains, which is what makes a one-tick gap harmless.

umh-core/pkg/diagnosis/declare.go
```diff
@@ -67,33 +67,49 @@ type Instrument[S any] struct {
...
+	// Instruments each answer the signal differently. The order is significant:
+	// the first instrument that can supply a number is the one used.
 	Instruments []Instrument[S]
+	// Refinements are signals declared under this one, narrowing its answer with
+	// their own instruments and marks. Each is sampled and judged on every tick,
+	// whether or not this signal fired, so its window is warm when this one fires
+	// and its Since is the tick IT crossed its mark. A refinement appears in
+	// Fired.Refinements under a parent that fired, and never as a verdict of its
+	// own. The Refinements section of the package doc works through an example.
+	Refinements []Signal[S]
```
Aaron99B, line 102: nit: I would now call it ChildSignals or something as this highlights better that we can not create trees of signals and refinements are no other types

umh-core/pkg/diagnosis/engine.go
```diff
@@ -67,115 +70,183 @@ const (
...
+	releaseOnAbsent bool
 }
 
-// trackState is one track and the single window it reduces through, paired for
-// the same reason. Track no longer walks one slice while indexing another.
-type trackState[S any] struct {
-	track  Track[S]
-	window SlidingWindow
+// capable applies Signal.Capable to the instruments this state holds.
+func (st *signalState[S]) capable(env Environment) []Instrument[S] {
```
Aaron99B: the name capable sounds to me like it would return a boolean

umh-core/pkg/diagnosis/engine.go
```diff
@@ -67,115 +70,183 @@ const (
...
+		e.signals = append(e.signals, buildSignalState(s.Name, s, i))
+	}
+
+	return e, nil
+}
+
+// pathSeparator joins a refinement's name onto its parent's path.
+// validateSignal refuses it inside any name, so no two signals can compose
+// the same path and share a window.
+const pathSeparator = "/"
```
Aaron99B, line 183: would be nicer to create a function for creating the paths instead

Aaron99B: Like this `func childPath(parent, name string) string {
      return parent + pathSeparator + name
}`

Aaron99B (approved):
A few smaller things. Conceptually I think it is fine now and with respect to the timeline I already approve the PR

umh-core/pkg/diagnosis/engine.go
JeremyTheocharis: Done in 08efd85c8, with the signature you wrote.

Three walks composed the path by hand — buildSignalState, buildWindows and validateSignal — and all three now call childPath, so there are no hand-written copies left.

umh-core/pkg/diagnosis/engine.go
JeremyTheocharis: Right, it returns \[\]Instrument. Renamed to capableInstruments in 08efd85c8.

I renamed the exported Signal.Capable to CapableInstruments in the same commit, even though it predates this PR. This wrapper just delegates to it, so renaming only the wrapper would have left two names for one operation. Nothing outside pkg/diagnosis imports the package, so no caller elsewhere changes.

umh-core/pkg/diagnosis/declare.go
JeremyTheocharis: I kept the name and filled the doc gap you were pointing at, in 08efd85c8.

On "refinements are no other types": agreed, and the doc now says so — "A refinement is itself a Signal, so it may declare refinements of its own, to any depth."

On trees: signals do nest to arbitrary depth. buildSignalState, buildWindows, validateSignal, observeWindows, judge and firedTree all recurse with no depth bound, and signalState's own doc says every node has its own latch "at every depth". So the type is not restricted to one level.

On the name itself: Refinements says what a child does, which is narrow the parent's answer. ChildSignals says only where it sits. The Refinements section of doc.go teaches the word that way — "the question is WHAT degraded it" — so renaming the field would leave that section explaining a word the API no longer uses. That is why I would rather keep it, but say if you still prefer the structural name and I will do the rename.

JeremyTheocharis: Sticking with `Refinements`, for the reason above — the doc section that teaches the word is named after it, and `Refinements []Signal[S]` already says a refinement is an ordinary signal.

PR #2723, opened 2026-08-20
test(fsmv2): pin three outbound-path defects and add a load-fitting scenario

JeremyTheocharis:
## Problem

An incident on 2026-08-19 left a customer's Console unresponsive, with bridge edits returning
*"Internal Error: No response from the UMH Instance"*. Four separate investigations read the agent log
and converged on the same evidence — the `outbound channel full, dropping message` lines. A
dose-response run in this PR shows those lines are a **later and more severe stage** than the one that
breaks the user's edit: at 2× oversubscription the reply is already lost with nothing refused and no log
line anywhere, and only at 5× does the queue saturate and start writing them.

Three defects on that path had no failing test, and there was no way to reproduce the behaviour outside
production or to show anyone a reproduction.

## Shipping

**Four failing specs across three defects.** No new module dependencies.

1. **A reply parks its goroutine on a full queue.** `pkg/communicator/actions/action_reply_backpressure_test.go`.
   `sendActionReplyInternal` ends in a bare `outboundChannel <- &umhMessage`: no select, no default, no
   context. Status messages take the other branch — `subscribers.go` uses select/default and logs a
   drop — so a full queue drops status and pins replies. `HandleActionMessage` sends four replies per
   action, the first *before* the action runs, so an action can park before doing any work. There are
   two such bare sends; the spec covers one.
2. **A transport reset discards an undelivered message.** `push/reply_survives_reset_test.go`.
   `CheckAndClearOnReset` sets `pendingMessages = nil`. That is right for a status snapshot regenerated
   every second and wrong for a reply nothing regenerates. Uses the real `PushDependencies`, because a
   mock's version clears nothing and would pass while production loses the reply.
3. **The drain is skipped exactly when it is needed.** `push/drain_after_deadline_test.go`.
   `if ctx.Err() == nil { a.drainChannelToPending(...) }` admits the drain only while time is left, and
   a tick that exhausted its budget never has any. Returning also skips Phase 2, the only other drain.
   **Note:** `push/action/push_test.go` contains `"should skip channel drain when context is canceled
   during retry"`, which asserts today's behaviour as the requirement. A fix has to delete that spec,
   not satisfy it. It is the only conflicting spec — the whole `push/...` tree passes with the guard
   removed.

**One parameterised scenario for fitting load against observed behaviour.**

```
go run ./pkg/fsmv2/cmd/runner --scenario push-load --duration 180s \
  --subscribers 2 --payload-bytes 8900 --per-message-cost 1s \
  --stall-every-nth 30 --stall-burst 4 --stall-for 33s
```

Dials: `--bandwidth --subscribers --payload-bytes --topic-count --queue-capacity --http-timeout
--stall-every-nth --stall-burst --stall-for --per-message-cost`. The closing `push_load_summary` line
reports the same observables an agent log yields — degrade cycles, budget expiries and resets per
minute, max pending dropped, queue-full duty — beside an echo of the settings, so comparing a run
against production is subtraction rather than interpretation.

Three things this needed that are not obvious, and each is commented where it lives:

- **A bandwidth is not a latency.** A fixed per-request delay makes message size unobservable, so the
  fault costs `ContentLength/rate` instead.
- **Burst length is a separate dial from duty cycle.** An isolated slow request leaves no trace: the
  failure is absorbed with no `action_failed`, and a child degrades only after three *consecutive*
  errors. Scattered stalls produce no observable event at all, which one run confirmed by reporting
  zeroes everywhere.
- **A per-message cost is not a bytes cost.** Both scale with batch size, and the two cannot be told
  apart by the field measurement available, which was taken where every request carries one message.
  Without that arm a fit gets attributed to the wrong dial.

**Harness capabilities**, all additive with zero-value defaults: sticky per-path faults on the mock
relay server (the existing injectors are one-shot and shared across every endpoint, so neither can hold
push slow while pull stays healthy), a bandwidth fault, a per-message cost charged after decode, an
injectable channel provider, and a configurable per-request timeout defaulting to production's 10s.

**Deleted**: two scenarios totalling 510s of runtime, superseded by the flags above. One asserted a
cause it could not identify; its final version records why, and the retraction is in the commit.

## NOT shipping

- **No production code change.** The diff is tests and harness only.
- **No fix for any of the three defects**, and no choice between the two candidate approaches — shed
  load above half-full (ENG-4743) versus drain into a bigger buffer (ENG-5669). Nothing has chosen.
- **The four specs are RED by design, so CI cannot pass and this must not merge as-is.** Either the fix
  lands here, or the specs are skipped with a pointer until it does. Flagging rather than hiding it.
- **No reproduction of the customer's full log shape.** The best fit reaches four of six observables
  within ~10% (degrade cycles 0.99 vs 1.06, resets 0.33 vs 0.29, budget expiries 0.66 vs 0.74, reply
  lost) and misses both depth measures. The residual is payload size, which is ENG-5651 and the ENG-3591
  family, not something this PR addresses.
- `--topic-count` converts through a measured floor plus slope. The slope is 190 B/topic from a
  120–8,000-topic sweep; it gives an order of magnitude, **not a threshold**, and a topic count must not
  be quoted as a limit.


JeremyTheocharis:
Superseded. A design review established that this drives the transport worker but re-implements the producer and the accounting around it: 773 lines of scenario where the existing harness already provides the run, the seeding and the logging.

Rebuilding on a fresh branch as one parameterised scenario driving the transport worker on **its own default config** — no timeout or queue-capacity dials — emitting the worker's own log lines rather than a derived summary, so a scenario log and an agent log are grep-compatible.

The four unit tests here pinned three real defects and are parked outside the repo rather than lost. Branch kept for that history.

Handoff with the commit ladder: `artifacts/2026-08-20_sonnlaender-latency/HANDOFF-transport-load-scenario.md`

PR #2728, opened 2026-08-24
feat: topic validation enforcement for write flows

FalconTube:
- Input topics of a write flow must now be located BELOW the bridge that uses it

## Problem

See ticket ENG-5694

## What we are shipping

See ticket ENG-5694

## What we are NOT shipping

- No data contract validation in this ticket. Will be done in ENG-5696

Fixes ENG-5694

### DIFF
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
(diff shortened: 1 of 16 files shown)


JeremyTheocharis:
Moving it to draft until we are ready to review this properly and have started the rollout ticket.

PR #2729, opened 2026-08-25
fix: disable fossa container scan for now

led0nk:
# Description

From yesterdays slack conversation we decided to temporarily disable the container-scan test until the issue with the dependencies are resolved by @JeremyTheocharis .


FalconTube (approved):

JeremyTheocharis (changes requested):
maybe not needed (klaus ran through the night), let me test

led0nk:
seems to be resolved

PR #2730, opened 2026-08-25
Fix/ci improvements

FalconTube:
## Problem

CI grew slow and messy: `betteralign` added noise without value, and `vet-nilaway-lint` duplicated work in pre-test since we don't act on lint fixes there anyway.

## What we are shipping

- Removed `betteralign` entirely
- Removed redundant `vet-nilaway-lint` from pre-test steps
- Faster CI with better caching overall using filter and sparse checkouts

## What we are NOT shipping

Nothing deferred here — this is cleanup only.

Fixes ENG-5750

### DIFF
--- a/.github/workflows/build-umh-core.yml
+++ b/.github/workflows/build-umh-core.yml
@@ -50,6 +50,11 @@ jobs:
       digest: ${{ steps.digest.outputs.digest }}
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          filter: blob:none
+          sparse-checkout: |
+            umh-core
+            .github
 
       - name: Compute short SHA
         id: short-sha
--- a/.github/workflows/go-analysis.yml
+++ b/.github/workflows/go-analysis.yml
@@ -28,6 +28,11 @@ jobs:
 
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          filter: blob:none
+          sparse-checkout: |
+            umh-core
+            .github
 
       - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7
         with:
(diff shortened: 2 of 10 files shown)


.github/workflows/build-umh-core.yml
```diff
@@ -50,6 +50,11 @@ jobs:
       digest: ${{ steps.digest.outputs.digest }}
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
```
JeremyTheocharis, line 53: for what do we have this? maybe add this to what we are shipping?

.github/workflows/go-analysis.yml
```diff
@@ -28,30 +28,42 @@ jobs:
...
           cache: true
+          cache-dependency-path: umh-core/go.sum
 
       - name: Run go vet
         run: make vet
 
   nilaway:
     name: Nilaway Analysis
-    runs-on: depot-ubuntu-24.04
+    runs-on: depot-ubuntu-24.04-4
```
JeremyTheocharis, line 48: why do we need to increase it?

JeremyTheocharis (approved):
pls adjsut PR description 🙂

.github/workflows/build-umh-core.yml
JeremyTheocharis: is it the caching? then add to the pr descirpion faster caching using filter and sparse checkout

.github/workflows/pull-request.yml
```diff
@@ -224,6 +240,10 @@ jobs:
       # the changelog guard for tag lookup and merge-base computation)
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
+          filter: blob:none
+          sparse-checkout: |
+            umh-core/CHANGELOG.md
+            .github
```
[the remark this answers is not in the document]
FalconTube: Show me evidence for this claim. I can not find it in the release notes of the action

.github/workflows/build-umh-core.yml
FalconTube: exactly, added it to description

.github/workflows/go-analysis.yml
FalconTube: nilaway was CPU starved. With this now it takes half the time.
We "only" pay double for this runner, so we get a net zero cost change for half the runtime :)

.github/workflows/pull-request.yml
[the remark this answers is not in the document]
FalconTube: I added a fixup now

PR #2734, opened 2026-08-27
fix(cpuhealth): read every CPU threshold from the rule that owns it

JeremyTheocharis:
## Problem

The healthy state gave out detailed Technical Details, but the degraded states did not. And in a
degraded state you want to know a bit about when it triggers and when it does not.

## What we are shipping

- **Every state carries Technical Details.**
- **Each rule states its own threshold.**
- **A rule that has tripped shows what it takes to recover, not what it took to trip.**
- **Rules that cannot apply here say so, and say why** — "not possible" on a machine that cannot
  report the signal, "measuring" while the window is still filling. Previously they were absent.
- **Thresholds are read from the rule that enforces them, not copied into the message.**

## What we are NOT shipping

- **No wire change.**
- **No changelog entry.**
- **A failed cgroup read keeps its bare line.** When the read failed we do not know which rules
  apply, so a table would state more than we know.
- **A failed `/proc/stat` read still says "measuring" rather than naming the read failure.** It needs
  a third wording and a new field, and it only happens on a box that is already broken.

### DIFF
--- a/umh-core/pkg/cpuhealth/message.go
+++ b/umh-core/pkg/cpuhealth/message.go
@@ -15,15 +15,22 @@
 // The customer-facing text: every sentence this package can print, and the
 // rules that pick which one.
 //
-// Every customer-visible string in this file is written down exactly once —
-// the constants below, plus the host-headroom sentence composeHealthy formats
-// inline, and one deliberate duplicate the const block names. A second copy of
-// a sentence is where drift starts.
+// Every customer-visible string in this file is written down exactly once. The
+// const block below holds the whole sentences; the Technical Details lines are
+// assembled from formats at the sites that build them, and the const block
+// names the one deliberate duplicate. A second copy of a sentence is where
+// drift starts.
 //
-// No threshold is written down here at all. Every number the Technical Details
+// No signal threshold is written down here. Every number the Technical Details
 // table states is read from the mark pair its own signal_*.go declares, which
 // is the pair the engine judges against, so the rule the message states and the
-// rule the code applies cannot come apart.
+// rule the code applies cannot come apart. That is why each pair is one var:
+// the signal_*.go files declare them and say no more about it.
+//
+// The one number this file decides for itself is the 0.05 cores at which the
+// healthy headline starts saying a box is close to being marked degraded. It is
+// a display band, not a rule the engine judges, and it is written relative to
+// hostHeadroomMarks.Fire without reading it.
 
 package cpuhealth
 
@@ -81,10 +88,10 @@ func composeHealthy(details Details) string {
 	// is withheld the message has no headline sentence left, so render the
 	// single "CPU: starting up." line over the table, where a line whose own
 	// window has not reduced yet says so rather than stating a figure nothing
-	// measured. It lasts two ticks after each start and
-	// respawn: the sampler derives no rate from its first read, and the mean
-	// over the rates after that needs two of them. It is not the zero-capacity
-	// case (a standing state) - they share a rendering path and nothing else.
+	// measured. It lasts two ticks after each start and respawn: the sampler
+	// derives no rate from its first read, and the mean over the rates after
+	// that needs two of them. It is not the zero-capacity case (a standing
+	// state): that one returns bare, this one carries the table.
 	if !usageMeasured(details) {
 		return cpuStartingUp + technicalDetails(nil, details)
 	}
@@ -133,8 +140,8 @@ func composeHealthy(details Details) string {
 			fmtCoresTotal(details.LogicalCpus), fmtCoresTotal(details.HostCpus))
 	}
 
-	// A healthy tick has no fired cause, so every rule shows the mark that
-	// would fire it.
+	// nil causes: a healthy tick has fired none, so no line states a clear
+	// mark.
 	return msg + technicalDetails(nil, details)
 }
 
@@ -262,9 +269,6 @@ type cpuRule struct {
 // contradiction: a rule still latched while its reading has fallen back under
 // its own fire mark would otherwise be shown a threshold it is already inside.
 func (r cpuRule) render() string {
-	// A fired instrument is proof the rule ran on this box, whatever the
-	// capability flag says, so a fired rule is never reported as one the box
-	// cannot run.
 	switch {
 	case !r.applies && !r.firedHere:
 		return r.label + " not available (not possible)."
@@ -313,8 +317,8 @@ func markValue(m diagnosis.Mark, unit string) string {
 }
 
 // technicalDetails renders the labelled table: the label on its own line, then
-// one line per rule the engine is judging, in a fixed order. causes is this
-// tick's fired list, and is empty on a healthy tick.
+// one line per rule in cpuRules' fixed order, whether or not this box can run
+// the rule. causes is this tick's fired list, and is empty on a healthy tick.
 //
 // A failed cgroup read (CapacityCores == 0) gets no table at all. On that tick
 // we do not know which rules apply, so a table would assert knowledge we do not
@@ -768,9 +772,10 @@ func toPercent(fraction float64) int {
 }
 
 const (
-	// The one line rendered on a tick whose usage figure is withheld (window
-	// below its floor, or the host reading absent): no headline, no advisory,
-	// no technical details.
+	// The line rendered on a tick whose usage figure is withheld (window below
+	// its floor, or the host reading absent): no headline and no advisory. The
+	// Technical Details table follows it, so an operator still sees which rules
+	// will be judged and which windows are still filling.
 	//
 	// Not cpuMonitoringUnavailable ("cgroup read failed"): on this tick the
 	// read succeeded and there is simply one sample, so claiming a read failure
--- a/umh-core/pkg/cpuhealth/message.go
+++ b/umh-core/pkg/cpuhealth/message.go
@@ -19,6 +19,11 @@
 // the constants below, plus the host-headroom sentence composeHealthy formats
 // inline, and one deliberate duplicate the const block names. A second copy of
 // a sentence is where drift starts.
+//
+// No threshold is written down here at all. Every number the Technical Details
+// table states is read from the mark pair its own signal_*.go declares, which
+// is the pair the engine judges against, so the rule the message states and the
+// rule the code applies cannot come apart.
 
 package cpuhealth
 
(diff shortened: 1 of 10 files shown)


umh-core/pkg/cpuhealth/table_cpu.go
```diff
@@ -53,16 +53,36 @@ func cpuTable(cores, quota float64) diagnosis.Table[Sample] {
...
 	}
-	// Only when a positive quota exists; with no limit there is nothing to saturate.
-	if quota > 0 {
+	if containerLimitFullDeclared(quota) {
 		t.Signals = append(t.Signals, containerLimitFullSignal(quota))
 	}
 	return t
 }
+
+// hostCpuFullDeclared reports whether the table holds the machine-full signal.
```
JeremyTheocharis, line 65: todo: make shorter

umh-core/pkg/cpuhealth/signal_saturation.go
```diff
@@ -34,6 +34,32 @@ import (
 	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/diagnosis"
 )
 
+// Marks are the two thresholds that turn a number into a yes or no: the value
```
JeremyTheocharis, line 37: todo: make comment shorter

umh-core/pkg/diagnosis/measure.go
```diff
@@ -41,6 +42,43 @@ func Unknown() Reading { return Reading{} }
 // Get returns the value and whether it is present; an absent Reading returns 0.
 func (r Reading) Get() (float64, bool) { return r.v, r.ok }
 
+// MarshalJSON writes a value as a JSON number and an absence as null. Both
+// fields of a Reading are unexported, so without this the encoder would see no
+// fields and write {}, dropping the number silently. A known zero is written as
+// 0, not null, because a zero is a value and only an absence is null.
+//
+// A NaN or an infinite value is written as null too. JSON has no way to spell
```
JeremyTheocharis: todo: make shorter

umh-core/pkg/diagnosis/measure.go
```diff
@@ -41,6 +42,43 @@ func Unknown() Reading { return Reading{} }
 // Get returns the value and whether it is present; an absent Reading returns 0.
 func (r Reading) Get() (float64, bool) { return r.v, r.ok }
 
+// MarshalJSON writes a value as a JSON number and an absence as null. Both
```
JeremyTheocharis: todo: make comment better

JeremyTheocharis:
```
It is the one home for both numbers: the
// engine judges against it and the Technical Details table states it, so the
// threshold the customer reads is the threshold that fires.
```

→ cut this and also from all the other files

umh-core/pkg/cpuhealth/message.go
```diff
@@ -19,6 +19,11 @@
 // the constants below, plus the host-headroom sentence composeHealthy formats
 // inline, and one deliberate duplicate the const block names. A second copy of
 // a sentence is where drift starts.
+//
+// No threshold is written down here at all. Every number the Technical Details
```
JeremyTheocharis: cut

```
// No threshold is written down here at all. Every number the Technical Details
// table states is read from the mark pair its own signal_*.go declares, which
// is the pair the engine judges against, so the rule the message states and the
// rule the code applies cannot come apart.
```

umh-core/pkg/cpuhealth/message.go
```diff
@@ -148,25 +133,367 @@ func composeHealthy(details Details) string {
...
+// Latchedness is read per SIGNAL and the capability override per INSTRUMENT;
+// cpuRule's two fields say why.
+func cpuRules(causes []Cause, details Details) []cpuRule {
+	machineFull := hasKind(causes, CauseKindHostCpuFull)
+	limitFull := hasKind(causes, CauseKindContainerLimitFull)
+	_, hostHeadroomFired := firedCause(causes, CauseKindHostCpuFull, instrumentHostHeadroom)
+	_, usageFired := firedCause(causes, CauseKindHostCpuFull, instrumentUsageFraction)
+	_, limitHeadroomFired := firedCause(causes, CauseKindContainerLimitFull, instrumentLimitHeadroom)
+
+	// Headroom is one subtraction against a ceiling, and a box can be judged
```
JeremyTheocharis: please cross check all these comments here  that were aded here and reduce it at least by 60% (see also comments.md)

JeremyTheocharis:
also make it way shorter, from my feeling: max 2 lines

umh-core/pkg/cpuhealth/table_cpu.go
JeremyTheocharis: Done in `a237eb29d`, but the honest answer is that it got shorter only after getting worse first.

The old second paragraph claimed sharing this predicate with `message.go` prevents printing "a line for a rule nothing is judging". That was false. Both sides call one function, neither passes it the same input: the table is built from the startup snapshot in `NewDeps`, the message runs on the current tick through `tableCeilings`. A cpuset that fails to read at startup drops the machine-full signal for the life of the instance (ENG-5752), and a later tick that reads fine then prints a headroom line for a ceiling nothing will ever fire on.

Replacing a false guarantee with the real case took the block from 10 lines to 12. It is back to 10 now, same two facts in four lines instead of six.

umh-core/pkg/cpuhealth/signal_saturation.go
JeremyTheocharis: Done — `3a115feb1`. This header is 8 lines down to 4.

The sentence that went was "Each pair below is the one home for its two numbers: the engine judges against it and the Technical Details table states it." It was written five times across four signal files, plus a sixth statement in `message.go`s header. The fact being repeated was that it is written down once.

It now lives in `message.go`s header alone, next to the code that depends on it. `message_test.go` computes 16 assertions from the declared pairs rather than from literals, so the relationship is pinned by a running spec rather than by the prose that just went.

umh-core/pkg/cpuhealth/message.go
JeremyTheocharis: Done — `3a115feb1`. Cut, and it was worth cutting for a reason beyond length: it was false.

It said no threshold is written down in this file. `composeHealthy` decides between telling a customer they are "close to being marked degraded" and telling them how much more they can use, on a bare `0.05` — a threshold on headroom, the same quantity `hostHeadroomMarks` governs, read from nothing. The header now names it as a display band rather than denying it exists.

umh-core/pkg/cpuhealth/message.go
JeremyTheocharis: Cross-checked, and the reduction came out at **9.9%, not 60%**. Reporting the number rather than closing the gap by deleting.

```
message.go   314 -> 283 comment lines   ratio 0.664 -> 0.598
             code 473 unchanged
```
Counted comment lines in both revisions minus the licence header, code as total − comments − blanks.

**One of the comments you flagged was false, which is the best thing this pass found.** `cpuRules` said the two headroom lines are "read off the figures the table was built from". They are not — `tableCeilings` returns the current tick. It also restated `hostCpuFullDeclared`, so it went rather than got corrected. That is an eighth false claim, on top of the seven already fixed on this branch.

**What produced the 31 lines.** Rationale clauses that `message_test.go` states verbatim in its own spec descriptions are now left to the spec — `cpuRule.latched`, the absence note, `render`s latched-mark reason. `tableCeilings`, `headroomRule` and `cpuRules` stopped restating the capacity predicates and point at them instead. `causeDetails` and `BlockReason` each lost a level of nesting.

**Where it stopped, and what the remaining 283 carry.** 55 lines document the customer-facing copy in the const block. The other 228 are mostly domain rules the code cannot show:

- the two capacity causes share one paragraph because their remedies contradict — a customer given both is told to buy CPU for a machine and to raise a limit that cannot help on it
- "reduce other software running on it" is wrong advice when the load filling the machine is our own
- `BlockReason` and the details must speak with the same cause so two surfaces cannot hand one customer contradictory remedies
- the ENG-5752 divergence, which has no other written home at the render site

Reaching 126 lines means deleting those.

**The lever I did not pull.** "Can the code carry it instead" is where the remaining volume is, and in this file every instance is a rename or an extraction — `render`s switch arms, `markSide`s polarity logic, `headroomRule`s construction. That is a behaviour-preserving refactor, and putting one inside a comment cleanup makes the diff unreviewable. Worth its own change if you want the number.

umh-core/pkg/cpuhealth/message.go
```diff
@@ -148,25 +138,342 @@ func composeHealthy(details Details) string {
...
+		used:    round1(used),
+		reserve: round1(reserve),
+	}
+	b.headroom = round1(b.total - b.used - b.reserve) // clears float residue, not an independent rounding
+
+	return b
+}
+
+// machineBudget is the machine's cores against the machine's busy time, the
+// pair hostCpuFullSignal's host-headroom arm subtracts. Its reserve is the
```
Aaron99B: I think we can reduce or even remove the comments here entirely. The function names are descriptive enough

umh-core/pkg/cpuhealth/message.go
```diff
@@ -148,25 +138,342 @@ func composeHealthy(details Details) string {
...
+// Details rather than recomputed - buildDetails filled it from the verdict's
+// own reserve, never from the constant.
+func instanceBudget(details Details) budgetCores {
+	return newBudget(details.CapacityCores, details.AvgUsageCores, details.ReserveCores)
+}
+
+// coresBudget is the budget the healthy headline speaks in, which is the one
+// the mode makes the customer's ceiling: their own limit where one applies, the
+// machine otherwise.
+func coresBudget(details Details) budgetCores {
```
Aaron99B: Let's rename this to effectiveBudget. with cores budget is sounds like it is computing the unit

Aaron99B (approved):
I think we should remove the amount of code comments and their size to make things more comprehensible

umh-core/pkg/cpuhealth/message.go
JeremyTheocharis: Renamed to effectiveBudget. You were right that coresBudget read as a unit rather than a choice — the function picks which of two budgets applies, it does not convert anything.

umh-core/pkg/cpuhealth/message.go
JeremyTheocharis: Cut. The docs on the budget and measured helpers now keep only what the name cannot carry: why newBudget derives headroom from the already-rounded three, why machineBudget uses the fixed reserve instead of Details.ReserveCores, and that applies and ready must never be read as each other. instanceMeasured lost its doc entirely. cpu.go's two longest field docs came along, since this PR owns their final text.

PR #2735, opened 2026-08-27
fix(fsmv2): drop type arguments from generic state names

JeremyTheocharis:
## Problem

The state transition logs print the full generic type name for `from_state` and `to_state`. We only
need `degraded` and `stopped`.

For reference:

    "from_state": "degradedState[github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/fsmv2/cpu.CPUConfig,github.com/.../cpu.CPUStatus]"

## What we are shipping

- **State names drop their type arguments**, so `from_state` and `to_state` read `degraded` and
  `stopped`.

## What we are NOT shipping

- **No change to non-generic states.** Around sixty exist and their names are byte-identical before
  and after.

⚠️ `state_transitions_total` is labelled by `from_state` and `to_state`, so this changes metric
label values as well as log text. Dashboards keyed on the old values will need updating.

### DIFF
--- a/umh-core/pkg/fsmv2/internal/helpers/base_state.go
+++ b/umh-core/pkg/fsmv2/internal/helpers/base_state.go
@@ -71,17 +71,21 @@ func DeriveStateName(state interface{}) string {
 		t = t.Elem()
 	}
 
-	name := t.Name()
-
-	// A generic type's reflect name carries its type arguments in brackets,
-	// each spelled with its full import path. That makes the name unusable as a
-	// log field or metric label, and the "State" suffix trim below cannot match
-	// while the brackets are still there.
-	if i := strings.IndexByte(name, '['); i >= 0 {
-		name = name[:i]
-	}
+	name := stripTypeArguments(t.Name())
 
 	name = strings.TrimSuffix(name, "State")
+
+	return name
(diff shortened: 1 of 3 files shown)


umh-core/pkg/fsmv2/internal/helpers/base_state.go
```diff
@@ -71,6 +72,15 @@ func DeriveStateName(state interface{}) string {
 	}
 
 	name := t.Name()
+
+	// A generic type's reflect name carries its type arguments in brackets,
```
Aaron99B: here it would be nice to instead of commenting it wrap the logic in an explanatory function whose names reflects what we are cutting from the name here

Aaron99B (approved):
small item then good to go

umh-core/pkg/fsmv2/internal/helpers/base_state.go
JeremyTheocharis: Done — the bracket-strip is now a named stripTypeArguments function and the explaining comment is gone. The doc that remains says only why the raw reflect name is unusable as a log field or metric label, which the name alone cannot express.

PR #2738, opened 2026-08-31
fix(diagnosis): measure window coverage from the first reading

JeremyTheocharis:
## Problem

Throttling never goes away once it fires — the verdict stays `degraded` long after throttling has decayed to zero, and only a restart clears it (seen with the fsmv2 CLI runner under `stress-ng`). Reason: the release check only passes when the poll interval divides the 60-second window exactly, and real intervals never do. Firing doesn't use that check, so the signal fires normally and then has no way out.

## What we are shipping

Instead of a release check that can only pass when the timing lands exactly right, the window now reports how long it has been collecting. The check passes once the window has been collecting for at least 60 seconds, not only when a reading lands exactly 60 seconds back.

## What we are NOT shipping

- **Firing still doesn't check for a full window.** That's deliberate — a signal should warn quickly and give the all-clear only on a full window.
- **Two successful reads 60 seconds apart now satisfy the release check** — thinner than before, when it needed readings spread across the whole window. How many readings a verdict actually needs is set by each signal: two for an average, twenty for a 95th percentile.
- **A test that the verdicts come out the same at any poll rate.** Stronger than the tests here, and easy to write so that it passes without testing anything, so it gets its own change.
- **Two other ways a signal can stay stuck**, both untouched: ENG-5639, where a different measurement blocks the release, and ENG-5641, where a late reading withholds one that was due.

Fixes ENG-5780

### DIFF
--- a/umh-core/pkg/diagnosis/latch_foreign_release_test.go
+++ b/umh-core/pkg/diagnosis/latch_foreign_release_test.go
@@ -43,7 +43,7 @@ import (
 var _ = Describe("Latch release is judged on the pair the episode fired under", func() {
 	const fireSpan = 60 * time.Second
 
-	full := Coverage{span: fireSpan, covered: fireSpan}
+	full := Coverage{span: fireSpan, collectedFor: fireSpan}
 
 	// Two answers to one question, in different units and opposite directions:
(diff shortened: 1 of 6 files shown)


FalconTube (approved):
Can confirm that this resolved the "stays in degraded" issue

PR #2741, opened 2026-09-02
chore: update openssl for container manually

led0nk:
# Description

This should fix the aikido openssl issue as a manual workaround. Usually we should keep on track with alpine image updates, but seemingly this wasn't fixed yet.


FalconTube (commented):
Just did a double check on this one here.
You are right, that latest alpine does not ship `.8` version of openssl.
But I saw, that one can just call `apk add --no-cache --upgrade libcrypto3` without specifying any specific version.
That way, it will always upgrade to the latest version of `openssl` and I don't see why we would ever want to pin the version specifically.

So I would say just always grab the latest version and remove the extra ENV variable for it.
What do you think?

FalconTube (approved):

PR #2746, opened 2026-09-04
feat(fsmv2/cpu): expose CPU health evidence on /metrics

JeremyTheocharis:
## Problem

fsmv2 has a standard way of exposing metrics and already uses it, but not for the CPU ones.

## What we're shipping

- Using the framework's existing worker-metrics path, the CPU worker now publishes its evidence via the `/metrics` endpoint, documented in `docs/production/metrics.md`.

## Not shipping

`StealP95` and `P95UsageCores` (their readiness flags cannot qualify a p95 window), `LogicalCpus` (already `umh_container_cpu_core_count`), `Verdict.Attribution` (needs an encoding home), and 5 booleans the console reads from the status payload.

Closes ENG-5835.


umh-core/pkg/fsmv2/cpu/cpu.go
```diff
@@ -126,13 +126,67 @@ func Poll(ctx context.Context, d *CPUDeps, _ CPUConfig) (CPUStatus, error) {
...
 		Verdict: verdict,
 		Message: cpuhealth.ComposeMessage(verdict, details),
 		Details: details,
 	}, nil
 }
 
+// recordGauges publishes the measured evidence for the framework's worker-metrics
+// exporter, which turns each name into umh_fsmv2_worker_<name>.
+//
+// This does not run on a tick that could not measure, because Poll returns
```
JeremyTheocharis: cut

```
// This does not run on a tick that could not measure, because Poll returns
// first. That does NOT mean the gauges go absent on a failed read: the
// collector still runs (simple.CollectObservedState reports a poll failure as
// a degraded observation, not as an error), so it reloads the previous gauge
// map from CSE and re-publishes every value. A persistent read failure
// therefore freezes this whole family at its last reading, and the worker's own
// state metric is what reveals it — not anything here.
```

umh-core/pkg/fsmv2/cpu/cpu.go
```diff
@@ -126,13 +126,67 @@ func Poll(ctx context.Context, d *CPUDeps, _ CPUConfig) (CPUStatus, error) {
...
+// first. That does NOT mean the gauges go absent on a failed read: the
+// collector still runs (simple.CollectObservedState reports a poll failure as
+// a degraded observation, not as an error), so it reloads the previous gauge
+// map from CSE and re-publishes every value. A persistent read failure
+// therefore freezes this whole family at its last reading, and the worker's own
+// state metric is what reveals it — not anything here.
+func recordGauges(m *deps.MetricsRecorder, det cpuhealth.Details) {
+	// Every gauge is set on every measured tick, including one where a signal
```
JeremyTheocharis:
```
// Every gauge is set on every measured tick, including one where a signal
	// was not ready, and the flags below are what say which reading to trust.
	// Skipping the set would be worse than publishing a 0: the exporter creates
	// gauges lazily and never deletes one (ExportWorkerMetrics only ever writes
	// the drained map), so an omitted gauge keeps being scraped at its previous
	// value, with nothing to mark it stale.
```

cut

umh-core/pkg/fsmv2/deps/metrics.go
```diff
@@ -197,6 +197,74 @@ const (
 	GaugePendingMessages GaugeName = "pending_messages"
 )
 
+// CPU worker gauge names for the evidence behind a CPU health verdict.
+//
```
JeremyTheocharis:
```
//
// Three of these measure "how busy" on different scales — absolute cores, a
// 0..1 fraction, and a 0..1 ratio — so each of those three names says which.
// The remaining names state their range in the doc comment instead.
```

cut

umh-core/pkg/fsmv2/deps/metrics.go
```diff
@@ -197,6 +197,74 @@ const (
...
+// consumer cannot tell "not throttled" from "no throttle signal".
+const (
+	// GaugeCPUUsageRingActive qualifies cpu_avg_usage_cores and cpu_avg_usage_fraction.
+	GaugeCPUUsageRingActive GaugeName = "cpu_usage_ring_active"
+
+	// GaugeCPUHostBusyRingActive qualifies cpu_avg_host_busy_cores.
+	GaugeCPUHostBusyRingActive GaugeName = "cpu_host_busy_ring_active"
+
+	// GaugeCPUHostHeadroomAvailable reports whether this container sees the whole
+	// machine, reading 0 when it is pinned to a subset of CPUs. That is a scope
```
JeremyTheocharis: cut

```
That is a scope
	// check, NOT the readability of cpu_host_headroom_cores, so it is a weaker
	// companion than the two above: require cpu_host_busy_ring_active == 1
	// alongside it before trusting a headroom figure. Neither flag covers an
	// instance whose startup snapshot failed, which drops the host-capacity
	// signal for that instance's lifetime and pins the headroom figure at 0
	// (ENG-5752).
```

umh-core/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Improvements
+
+- Preview: `USE_FSMV2_CPU` now publishes CPU health metrics on `/metrics`
```
JeremyTheocharis: needs to be aligned with our changelog writing skills

umh-core/pkg/fsmv2/deps/metrics.go
```diff
@@ -197,6 +197,74 @@ const (
...
+	// check, NOT the readability of cpu_host_headroom_cores, so it is a weaker
+	// companion than the two above: require cpu_host_busy_ring_active == 1
+	// alongside it before trusting a headroom figure. Neither flag covers an
+	// instance whose startup snapshot failed, which drops the host-capacity
+	// signal for that instance's lifetime and pins the headroom figure at 0
+	// (ENG-5752).
+	GaugeCPUHostHeadroomAvailable GaugeName = "cpu_host_headroom_available"
```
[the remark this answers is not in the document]
JeremyTheocharis: You're right that `cpu_host_headroom_available` does not report whether `cpu_host_headroom_cores` is readable. It is dispatched on the sample's CPU scope (`buildDetails`: `CpuScope == ScopeHost`), so on a host-scoped box it reads 1 while the headroom figure sits at 0 — the ENG-5752 case, where a startup cpuset failure means the host-cpu-full signal is never declared.

One correction to the scope of it: headroom is not the only such gauge. `cpu_avg_usage_fraction`, `cpu_host_cpus` and `cpu_capacity_cores` have no flag either and can each publish a fabricated 0.

We did not add a readiness gauge, because the state that would populate it is discarded a layer down: `buildDetails` drops the `State` from `Reduced.Get`, and `Details` carries no field for it, so `recordGauges` has nothing to read. Adding one means a new `Details` field, and that struct documents a cap of two `*Available` members before the family has to be modelled as a set. That work belongs with ENG-5752, not here.

What changed instead: `metrics.md` and the constant block now state the rule — a flag exists only where one is declared, and a gauge without one cannot be told apart from a measured 0 — rather than an enumeration that rots. The two flag families are now documented separately, since `_ring_active` and `_signal_ready` call for opposite operator actions. And `cpu_last_sample_unix` was added, because on a tick that cannot measure the collector re-publishes the previous values and nothing else marked them stale.

umh-core/CHANGELOG.md
JeremyTheocharis: Rewritten against the changelog-writing skill. It now says what an operator gets rather than naming the mechanism, and it names both prerequisites — `USE_FSMV2_CPU=true` alone publishes nothing, since the supervisor carrying the worker metrics is only built when `UseFSMv2Transport` is set.

Kept under `### Improvements`: CPU health already shipped and `/metrics` already exposed other metrics, so this refines an existing preview. Worth knowing the skill says to write `### New`, which this repo's CI rejects outright — its validator accepts only Breaking Changes, New Features, Improvements, Fixes and `Preview: [Name]`.

umh-core/docs/production/metrics.md
```diff
@@ -6,4 +6,65 @@
...
 * Agent tick & FSM timings (each full reconcile loop < 100 ms by design)
 * Per-DFC counters: processed, error, latency, active / idle flag
-* Redpanda I/O and disk-utilisation stats
\ No newline at end of file
+* Redpanda I/O and disk-utilisation stats
+* CPU health evidence (preview), described below
+
+## CPU health evidence (preview)
+
+Every CPU gauge is scraped under the prefix `umh_fsmv2_worker_`, with a `hierarchy_path` label naming
```
JeremyTheocharis: cut

````
Every CPU gauge is scraped under the prefix `umh_fsmv2_worker_`, with a `hierarchy_path` label naming
the worker that published it. The gauge `cpu_avg_usage_cores` therefore arrives as the series
`umh_fsmv2_worker_cpu_avg_usage_cores{hierarchy_path="..."}`. A rule written against the bare gauge
name matches nothing, and Prometheus reports no error for it.
### Check the sample age first
`cpu_last_sample_unix` carries the unix seconds of the last tick that measured. A tick that fails to
measure republishes the previous values, so every CPU series keeps arriving at its last reading with
nothing to mark it stale. Alert on the age of this gauge before trusting anything beside it:
```promql
time() - umh_fsmv2_worker_cpu_last_sample_unix > 120
```
````

File: umh-core/docs/production/metrics.md [hunk 8]
```diff
@@ -6,4 +6,65 @@
...
+A flag gauge qualifies the measurement beside it. Each flag qualifies one measurement:
+
+| Flag | Qualifies |
+|---|---|
+| `cpu_usage_ring_active` | `cpu_avg_usage_cores` |
+| `cpu_host_busy_ring_active` | `cpu_avg_host_busy_cores` |
+| `cpu_throttle_signal_ready` | `cpu_throttle_ratio` |
+| `cpu_pressure_signal_ready` | `cpu_pressure_avg60_ratio` |
+
+The two families call for opposite actions. A `_ring_active` of 0 means the 60s window has not
```
JeremyTheocharis: cut

````
The two families call for opposite actions. A `_ring_active` of 0 means the 60s window has not
filled. A non-zero mean beside it is a real average over fewer samples, so keep it and read it as
covering less than 60s. A 0 beside it is ambiguous: the window may instead be empty because nothing
could be read, and the flag does not separate those two cases. A `_signal_ready` of 0 means the
signal could not be read at all, so the value beside it is a 0 nobody measured and you should
discard it.
A measurement that no flag qualifies gives you no way to separate an absent signal from a measured 0.
Alert on such a gauge only where both readings call for the same action.
`cpu_host_headroom_available` is not a flag. It reports whether this container sees the whole machine,
and reads 0 when the container is pinned to a subset of CPUs.
### Not published yet
The CPU verdict also judges steal time, and this preview publishes no gauge for it. On a virtual
machine with a noisy neighbour the verdict can turn degraded while every CPU gauge above reads normal.
jeremytheocharis
cut

Every CPU gauge is scraped under the prefix `umh_fsmv2_worker_`, with a `hierarchy_path` label naming
the worker that published it. The gauge `cpu_avg_usage_cores` therefore arrives as the series
`umh_fsmv2_worker_cpu_avg_usage_cores{hierarchy_path="..."}`. A rule written against the bare gauge
name matches nothing, and Prometheus reports no error for it.
### Check the sample age first
`cpu_last_sample_unix` carries the unix seconds of the last tick that measured. A tick that fails to
measure republishes the previous values, so every CPU series keeps arriving at its last reading with
nothing to mark it stale. Alert on the age of this gauge before trusting anything beside it:
```promql
time() - umh_fsmv2_worker_cpu_last_sample_unix > 120
```
````

File: umh-core/docs/production/metrics.md [hunk 9]
```diff
@@ -6,4 +6,32 @@
...
+| `cpu_pressure_avg60_ratio` | the kernel's PSI cpu-some avg60 as a 0..1 fraction |
+| `cpu_host_headroom_cores` | cores free on the host after the reserve; negative on a full machine |
+| `cpu_avg_host_busy_cores` | the whole machine's 60s mean busy time, in cores |
+| `cpu_capacity_cores` | the ceiling CPU health judged against, in cores |
+| `cpu_reserve_cores` | the part of that ceiling held back, in cores |
+| `cpu_host_cpus` | the machine's CPU count |
+
+### Flags
+
+Each flag qualifies one measurement:
```
Aaron99B: What does it mean for a measurement to be qualified?

umh-core/pkg/fsmv2/cpu/cpu.go
```diff
@@ -136,13 +137,51 @@ func Poll(ctx context.Context, d *CPUDeps, _ CPUConfig) (CPUStatus, error) {
 	env := cpuhealth.DeriveEnvironment(sample)
 	verdict, details := cpuhealth.Decide(d.engine, sample, env)
 
+	recordGauges(d.MetricsRecorder(), sample.Timestamp, details)
```
Aaron99B: Can we call it recordMetrics instead?

umh-core/pkg/fsmv2/cpu/cpu.go
```diff
@@ -136,13 +137,51 @@ func Poll(ctx context.Context, d *CPUDeps, _ CPUConfig) (CPUStatus, error) {
...
+	m.SetGauge(deps.GaugeCPUHostCpus, det.HostCpus)
+
+	m.SetGauge(deps.GaugeCPUUsageRingActive, gaugeBool(det.UsageRingActive))
+	m.SetGauge(deps.GaugeCPUHostBusyRingActive, gaugeBool(det.HostBusyRingActive))
+	m.SetGauge(deps.GaugeCPUHostHeadroomAvailable, gaugeBool(det.HostHeadroomAvailable))
+	m.SetGauge(deps.GaugeCPUThrottleSignalReady, gaugeBool(det.ThrottleSignalReady))
+	m.SetGauge(deps.GaugeCPUPressureSignalReady, gaugeBool(det.PressureSignalReady))
+}
+
+func gaugeBool(b bool) float64 {
```
Aaron99B: Let's rename the function to what it does. An bool to int conversion. Further instead of calling the function multiple time here, we could include it in metrics.go in the setGaugeFlag function

umh-core/pkg/fsmv2/deps/metrics.go
```diff
@@ -197,6 +197,64 @@ const (
...
+
+	// GaugeCPULastSampleUnix carries the unix seconds of the last tick that measured.
+	// Every other cpu_* gauge holds its last value until then, so its age reveals a freeze.
+	GaugeCPULastSampleUnix GaugeName = "cpu_last_sample_unix"
+)
+
+// CPU worker readability flags, 1 for true and 0 for false. *_ring_active: the 60s
+// window reduced to a trusted number. *_signal_ready: the signal could be read.
+const (
+	// GaugeCPUUsageRingActive qualifies cpu_avg_usage_cores.
```
Aaron99B: Why do all of these have different suffixes? Could we call all of them something like ... Measured?

umh-core/docs/production/metrics.md
JeremyTheocharis: Fair — the word was doing work it never defined, and it is gone.

The relation is now "applies to", and the flags table says what a flag means rather than assuming it: a flag reads 1 or 0, and a 0 means the number beside it is not worth acting on this tick. A figure averaged over too few samples and a 0 nobody measured both publish with the flag at 0, and nothing in the exposed metrics tells them apart.

Two gauges were also missing from the page entirely — `cpu_last_sample_unix` and `cpu_host_headroom_available` — and the series prefix was gone, so a rule written against a bare name from these tables would have matched nothing.

umh-core/pkg/fsmv2/cpu/cpu.go
JeremyTheocharis: Renamed to `recordMetrics`.

For the record on why I did not argue it: the function sets only gauges today, so `recordGauges` was the narrower and more accurate name. But "metrics" is the framework's own word for this path — the exporter is `WorkerMetricsExporter` and the recorder is `MetricsRecorder` — and the narrower name would have had to change the first time a counter joined it. Your name is the one that does not rot.

umh-core/pkg/fsmv2/deps/metrics.go
JeremyTheocharis: You were right to push on this, and my first answer was wrong.

I was going to argue that the two families call for opposite actions: keep the value beside a 0 `_ring_active`, discard the value beside a 0 `_signal_ready`. That does not survive the code.

`*SignalReady` is `Availability == Ready`, and `NoneReady` is explicitly the still-filling case (`pkg/diagnosis/engine.go`). `*RingActive` is `state == StateValue`, which excludes `StateAbsent` as well as the filling case, and an empty window reduces to v=0 (`sliding_window.go`). So both families read 0 while a window fills, and in both cases the number beside the flag may be a real average over too few samples or a 0 nobody measured — indistinguishable on the wire. The action is the same for both: do not act on it. That is now the first thing the flags section says, in the code and on the page.

I kept three suffixes rather than merging them, because what differs is the cause, and one of those causes never resolves:

* `_ring_active` — the 60-second window has not reduced to a trusted number.
* `_signal_ready` — the signal has no trusted reading. On a box with no instrument for it at all, such as a bare-metal host with no cgroup throttle counters, it never turns 1.
* `_available` — the sample could not supply the figure.

That is the const block's header comment now, so the question is answered where you asked it.

Two things fell out of chasing this down. `cpu_host_headroom_available` is back in the same block as the others: it is the scope gate `instrumentHostHeadroom` applies, so it does qualify a measurement, and I had moved it out on the wrong theory. And `cpu_host_headroom_cores` turned out to have no flag of its own while the documented rule named `cpu_host_busy_ring_active` — which does not gate it, because `measurementHostBusy` carries no scope guard, so a container pinned to a subset of CPUs publishes that flag at 1 beside an unmeasured 0. The three flags that do gate it are `cpu_host_headroom_available`, `cpu_host_busy_cores_available` and `cpu_host_busy_ring_active`, and the second of those was published nowhere, so no Prometheus rule could reproduce the conjunction `machineMeasured` already applies. It ships here.

umh-core/pkg/fsmv2/cpu/cpu.go
JeremyTheocharis: Both done. `gaugeBool` is gone, and the mapping is `SetGaugeFlag` on `MetricsRecorder`, next to `SetGauge`. It is exported because the callers live in `pkg/fsmv2/cpu`, and it routes through `SetGauge` so the mutex stays in one place.

One site I deliberately did not migrate: `pull.go` sets `GaugeBackpressureActive` to a literal 1 and 0 in two branches. Moving it onto `SetGaugeFlag` would make it write on every tick rather than on a transition, which is a behaviour change rather than a refactor.

umh-core/pkg/fsmv2/deps/metrics.go
```diff
@@ -197,6 +197,64 @@ const (
...
+
+	// GaugeCPUThrottleSignalReady qualifies cpu_throttle_ratio.
+	GaugeCPUThrottleSignalReady GaugeName = "cpu_throttle_signal_ready"
+
+	// GaugeCPUPressureSignalReady qualifies cpu_pressure_avg60_ratio.
+	GaugeCPUPressureSignalReady GaugeName = "cpu_pressure_signal_ready"
+)
+
 // =============================================================================
 // METRICS INFRASTRUCTURE
```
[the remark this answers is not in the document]
JeremyTheocharis: This is the finding you filed on 4 September and withdrew on 16 September, after agreeing the readiness state would need `Details` model work belonging to ENG-5752. Re-filing it under a new run does not make it new.

The substance is now addressed, by a different route than the headroom-specific readiness gauge you proposed.

The first-sample case you describe is real but was already visible: the headroom instrument extracts from `s.HostBusy` and returns `Unknown` when it is absent, and `cpu_host_busy_ring_active` is the state of that same host-busy measurement, so it reads 0 on the tick you name.

What was genuinely missing is different, and neither of your two filings found it. `machineMeasured` gates the host figures on `HostBusyRingActive && HostBusyCoresAvailable`, and the second half was published nowhere, so no Prometheus rule could reproduce the rule the product applies to the same number. `cpu_host_busy_cores_available` now ships. `cpu_host_headroom_cores` is documented as having no flag of its own, trustworthy only when `cpu_host_headroom_available`, `cpu_host_busy_cores_available` and `cpu_host_busy_ring_active` all read 1.

`GaugeCPUHostHeadroomAvailable` keeps its meaning and its name.

JeremyTheocharis:
## Review round: what changed and why

All four comments are addressed in `41bbed32c`, and I resolved the threads. Reopen any of them if the answer does not hold — one of them says I was wrong, so it is worth a look.

**Your suffix question changed the diff more than anything else in the round.** I was going to defend the three suffixes by arguing the families call for opposite operator actions: keep the value beside a `0` on `_ring_active`, discard it beside a `0` on `_signal_ready`. That is not what the code does. `*SignalReady` is `Availability == Ready`, and `NoneReady` is the still-filling case; `*RingActive` is `state == StateValue`, which excludes an empty window as well as a filling one, and an empty window reduces to `v=0`. Both families read `0` while a window fills, and in both cases the number beside the flag is either a real average over too few samples or a `0` nobody measured, with nothing on the wire to separate them. One action for both: do not act on it. That sentence now leads the flag documentation in the code and on the page.

I kept three suffixes because the *cause* differs and one of the causes never resolves: a window that has not filled, a signal with no trusted reading (permanently so on a box with no instrument for it), and a sample that could not supply the figure. That is the const block header now, so the answer sits where the question was asked.

**Chasing that down turned up two defects in the documentation, both mine.**

The rule for `cpu_host_headroom_cores` named `cpu_host_busy_ring_active`, which does not gate it. `instrumentHostHeadroom` measures only on a host-scoped sample; `measurementHostBusy` carries no scope guard. A container pinned to a subset of CPUs therefore publishes that flag at `1` beside an unmeasured `0` — the exact reading the sentence existed to prevent.

The three flags that do gate it are `cpu_host_headroom_available`, `cpu_host_busy_cores_available` and `cpu_host_busy_ring_active`. The second was published nowhere, so no Prometheus rule could reproduce the conjunction `machineMeasured` already applies to the same figure. It ships here, which moves one boolean out of the "not shipping" list in the description.

**How it was verified.** `cpu_host_busy_cores_available` reads the same three values as `cpu_pressure_signal_ready` over the ticks `flagTicks` staged, and that spec fails when two flags share a row, by design — that guard is what makes a crossed publish detectable. Rather than weaken it, the fixture gained a fourth tick where the host-busy read succeeds and the pressure read does not. Each of the six flags' value on that tick was derived from the code before the suite ran, and swapping two constants in `recordMetrics` fails the spec on exactly that tick.

**Two things I left alone**, both pre-existing and outside this change: the doc examples on `IncrementCounter` and `SetGauge` call a `deps.Metrics()` that does not exist, and `adapter/manager.go:41` still uses "qualifies" in the sense this round retired everywhere else.

Aaron99B (approved):

PR #2751, opened 2026-09-07
test(integration): prove the connection check rolls back an edit to an unreachable target

Aaron99B:
Top of the stack: #2708 → #2716 → **#2751**. Umbrella: ENG-5853.

## Problem

The connection check has unit coverage only. Nothing drives it through the real ManagementConsole backend, which is where the symptom was reported: an edit to an unreachable target answering `action-success`.

## What we are shipping

- A container harness that runs the ManagementConsole backend and router as local Go processes, the way ManagementConsole's own Playwright suite runs them, with the private source pulled through `gh` (`make pull-managementconsole`). What it makes real, and what a fake backend cannot: instance login against a database row, JWT-scoped user auth, and the v3 Redis message queue routing the reply back.
- One spec, run once per nmap backend: a bridge with a read flow edited to an unreachable target must reply `action-failure`, and the reply history must name the successful rollback. A terminal failure alone does not prove the restore, because `awaitRollout` reports the same state when the rollback itself fails and when it aborts on a render error without dialling anything.
- The edit back to the reachable target, as the positive control: without it a harness that fails every edit satisfies the spec above.
- `make connection-rollback-test`.

## What we are NOT shipping

- Any CI wiring. The `connection-edit-rollback` label is excluded from `make unit-test`, and `make integration-test` filters on `integration`, so no job runs these specs. They need `gh` access to the private ManagementConsole repo and a 30-minute timeout, so that may be deliberate. If it is not, this needs a workflow.
- Red-then-green in this PR's own CI. These specs assert the fixed behaviour and are green on this branch; the failing baseline lives in #2708's `pkg/fsmv2/nmap/manager_test.go` and `pkg/communicator/actions/edit-protocolconverter_awaitrollout_test.go`.

Part of ENG-5853

### DIFF
--- a/umh-core/.gitignore
+++ b/umh-core/.gitignore
@@ -2,6 +2,9 @@ data/
 
 integration/umh-config
 
+# Pulled from the private ManagementConsole repo by `make pull-managementconsole`
+integration/managementconsole
+
 *.test
 
 .docker-cache
--- a/umh-core/.gitignore
+++ b/umh-core/.gitignore
@@ -3,7 +3,7 @@ data/
 integration/umh-config
 
 # Pulled from the private ManagementConsole repo by `make pull-managementconsole`
-integration/managementconsole
+integration/.managementconsole
 
 *.test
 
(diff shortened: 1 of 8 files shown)


Aaron99B:
Will work on a dedicated e2e test suite instead of this https://linear.app/united-manufacturing-hub/issue/ENG-5900/rollout-e2e-test-suite

PR #2755, opened 2026-09-10
docs(eng-5132): CPU Health page for the preview CPU reporting

Aaron99B:
## Problem

The new CPU reporting shipped in umh-core without any documentation. A customer who sees "CPU taken by the server" or "Can't add another bridge: the machine is full" has nowhere to look up what it means, and the sizing guide is the only page that mentions CPU at all. The page existed on the closed #2613 and went nowhere when that PR closed.

## What we are shipping

- A **CPU Health** page under Production: what each status means and what to do about it, when UMH refuses a new bridge, and a glossary for throttling, CPU pressure, steal and host contention.
- The page is written for the preview it currently is. `USE_FSMV2_CPU` defaults to `false` (`container_monitor.go:95`), and with it off an instance is still marked degraded above 70% of its usable cores (`judgeLegacyCPUUsage`, `CPUHighThresholdPercent = 70.0`). The callout names both the flags needed and the default behaviour, so the page is not describing something no installation has.
- A one-line cross-link from the sizing guide.

Every status name, threshold and example message on the page was read off `pkg/cpuhealth` on `staging`, not off the closed PR.

## What we are NOT shipping

- **No sizing changes.** The 2 → 4 vCPU bump is ENG-5504 and its stated reason is the headroom reserve of a model that is not yet the default. It stays its own PR.
- **No rewrite of the sizing guide's CPU-blocking bullets.** They describe the default 70% behaviour, which is still what a default build does. They become wrong the day the flag flips, and that edit belongs with the flag flip.
- **No changelog entry.** Docs-only, and the preview itself is already announced in the Console changelog. Carries `skip-changelog-guard`.

Part of ENG-5132

### DIFF
--- a/umh-core/docs/SUMMARY.md
+++ b/umh-core/docs/SUMMARY.md
@@ -43,6 +43,7 @@
         * [nginx](production/deployment/docker-compose/additional-services/nginx.md)
       * [Updating](production/deployment/docker-compose/updating.md)
   * [Sizing Guide](production/sizing-guide.md)
+  * [CPU Health](production/cpu-health.md)
   * [High Availability](production/high-availability.md)
   * [Metrics](production/metrics.md)
   * [Migration from Classic](production/migration-from-classic.md)
--- a/umh-core/docs/production/cpu-health.md
+++ b/umh-core/docs/production/cpu-health.md
@@ -1,0 +1,64 @@
+# CPU Health
+
+{% hint style="info" %}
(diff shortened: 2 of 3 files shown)


Aaron99B:
Superseded by #2756. Same commit, re-opened on a branch named for ENG-5897 (the docs ticket) instead of the rollout ticket: the `eng-5132` branch name linked this PR to the rollout in Linear and moved it out of Reviewed.

PR #2756, opened 2026-09-10
docs(eng-5897): CPU Health page for the preview CPU reporting

Aaron99B:
## Problem

The CPU reporting shipped with no documentation. A customer who sees "CPU taken by the server", or has a bridge refused, has nowhere to look it up. The page was written in #2613 and lost when that PR closed.

## What we are shipping

A **CPU Health** page under Production: what each status means and what to do, the degrade and recover value per signal plus where the Console prints them live, and why a degraded CPU blocks a new bridge. Plus a cross-link from the sizing guide.

Written for the preview it is: `USE_FSMV2_CPU` defaults to `false`, and the callout says so. Every status, threshold and message came from `pkg/cpuhealth` on `staging`, not from #2613.

## What we are NOT shipping

- **No sizing changes** (ENG-5504), and no rewrite of the sizing guide's CPU-blocking bullets — those describe today's default and belong with the flag flip.
- **No changelog entry**: docs-only, carries `skip-changelog-guard`.

## For the reviewer

- `cpuhealth.BlockReason` has no caller outside its own tests. Its per-cause refusal lines were never wired up, so the page quotes no refusal wording. Worth a ticket.
- The `cpuset` limitation is documented here, though ENG-5132 calls that case out of scope. Say so and it comes out.

Thirteen review comments answered on their threads. Two changed what the page claims: a blocked bridge shows "System in degraded state", not the CPU message, because the container FSM state is checked first (`protocolconverter.go:1202`); and headroom recovers at a higher reading, not a lower one, so the thresholds paragraph no longer claims every signal recovers downwards (`signal_saturation.go:43-46`). Two suggestions were declined with reasoning — editing the quoted product output would have made the page wrong, and the legacy 70% rule in the callout really is exclusive (`cpu_legacy.go:70`).

The latest round is editorial: the page now opens on what the reporting does rather than on the limit question, the two asides about a slow machine are gone, the "CPU running near full" row says what it means without forward-referencing Technical Details, the section title carries its subject for a search result, and the glossary separates host contention from the CPU contention status. The early-access callout now says where the two flags go.

Fixes ENG-5897


umh-core/docs/production/cpu-health.md
```diff
@@ -0,0 +1,64 @@
...
+UMH measure starvation directly and turns on full CPU health monitoring.
+{% endhint %}
+
+## When UMH refuses a new bridge
+
+When CPU is in any of the "something's wrong" states above, UMH won't start an additional bridge on
+the instance, because a new bridge would only compete for CPU that's already short. The refusal
+message names the same cause and fix as the status, for example:
+
+```
```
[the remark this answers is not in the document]
Aaron99B: Done, tagged `text`. For the record it was optional: no markdownlint job runs in this repo (`.github/workflows/pull-request.yml` has none) and untagged fences are common across `umh-core/docs/`. Tagged anyway since the block is literal terminal output.

umh-core/docs/production/cpu-health.md
```diff
@@ -0,0 +1,64 @@
...
+won't flip the status, and a status won't clear until the reading comes back past a second,
+healthier threshold, so the status stays stable instead of flickering between healthy and degraded.
+
+## Glossary
+
+| Term | Meaning |
+|------|---------|
+| **Throttling** | The system caps each container to its CPU limit in short, repeating periods (about 100 ms). If the container needs more within a period, it's paused until the next one, so even a workload whose average looks fine can be paused during bursts. Frequent pausing means work waits. |
+| **CPU pressure** | How much time tasks spent waiting for a free CPU core (Linux Pressure Stall Information, PSI). High pressure means CPU is the bottleneck. |
+| **CPU steal** | Time the hypervisor gave this machine's CPU to other virtual machines (or, on burstable cloud instances, the point where your CPU credits run out). High steal means the server is oversubscribed. |
```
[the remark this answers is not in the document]
Aaron99B: Fair, they are two mechanisms. Reworded so the entry describes hypervisor scheduling and then says a burstable instance out of credits is capped by a different mechanism that the guest still counts as steal, rather than presenting them as the same thing.

umh-core/docs/production/sizing-guide.md
```diff
@@ -59,6 +59,8 @@ Since every bridge has different resource requirements (OPC UA with 10,000 tags
 - **Memory Usage**: Blocks if memory exceeds 80%
 - **Disk Usage**: Blocks if disk exceeds 85%
 
+For what each CPU status means and how CPU health is judged, see [CPU Health](./cpu-health.md).
```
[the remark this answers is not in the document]
Aaron99B: Addressed, but not with the proposed replacement. The 70% bullet is what a default build still does: `USE_FSMV2_CPU` defaults to `false` (`container_monitor.go:95`) and `judgeLegacyCPUUsage` degrades above `CPUHighThresholdPercent = 70.0`. Rewriting it as CPU-health behaviour would describe something no default installation has. The bullet now keeps its statement and adds what changes under the preview, with a link. The full rewrite belongs to the PR that flips the flag.

umh-core/docs/production/cpu-health.md
```diff
@@ -0,0 +1,96 @@
...
+
+Both the status and Technical Details carry the thresholds, so you do not have to look them up.
+A healthy instance states its remaining headroom, and each Technical Details line gives the
+reading next to the mark that would change it:
+
+```
+CPU healthy. This instance is using 0.3 of 2 cores (15% of its limit) and can use 1.5 more before it is marked degraded.
+
+Technical Details:
+Instance headroom 1.5 cores = 2 total - 0.3 used - 0.2 reserved (degrades below 0).
```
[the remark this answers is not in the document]
Aaron99B: The inconsistency was real, the suggested fix was not: that block quotes what the product prints, and `markSide` returns "below" for a non-inclusive mark (`message.go:269-278`), so `degrades at 0 or below` would have made the page wrong. Fixed the prose instead. `pkg/diagnosis/judge.go:130-134` compares `x > fx`, so the table now reads "above 5%" / "past 90%", "at 70%" for the one mark declared `Inclusive`, plus a sentence on the distinction.

umh-core/docs/production/cpu-health.md
```diff
@@ -0,0 +1,97 @@
...
+
+Steal uses the 95th percentile once 20 samples are in, and the mean before that, so a fresh
+instance is judgeable within seconds of starting. Bare metal reports no steal at all, so on a
+physical machine that signal reads "not possible" rather than 0%.
+
+## When UMH refuses a new bridge
+
+While CPU is degraded, UMH will not start an additional bridge on the instance, because it would
+compete for CPU that is already short. The bridge stays pending, and its status reason repeats the
+CPU status message, so the cause and the fix are the ones in the table above. Bridges already
```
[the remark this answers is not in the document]
Aaron99B: Correct, and this one contradicted the page. Traced it: a degraded CPU sets `OverallHealth = Degraded` (`container_monitor.go:178`), `areAllMetricsHealthy` reads that field (`actions.go:112`), and the FSM takes `metrics_not_ok` to degraded (`machine.go:55`) - so the state check at `protocolconverter.go:1202` fires first and the CPU branch is reached only before that transition. The page now says the reason names the resource gate, usually "System in degraded state", and points at the CPU status for the cause and fix.

umh-core/docs/production/cpu-health.md
```diff
@@ -0,0 +1,97 @@
...
+running are left alone.
+
+Setting `agent.enableResourceLimitBlocking: false` in `config.yaml` turns this off, and then a
```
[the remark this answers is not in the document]
Aaron99B: Right, the dotted key would fail strict parsing if copied. Now a nested YAML block.

umh-core/docs/production/cpu-health.md
```diff
@@ -0,0 +1,102 @@
+# CPU Health
+
+{% hint style="info" %}
+**Preview.** This reporting needs UMH Core v0.44.37 or later, started with `USE_FSMV2_CPU=true`
```
robertschwarz-umh: I'd adjust this to say "Early access", because that's what we use in other locations: [Example](https://docs.umh.app/usage/historian#:~:text=Early%20access.%20The%20Historian%20is%20not%20switched%20on%20by%20default.%20If%20you%20want%20to%20use%20it%2C%20get%20in%20touch%20with%20us%20on%20Discord%20or%20through%20your%20usual%20UMH%20contact%2C%20and%20we%20will%20show%20you%20how%20to%20get%20started.)

PS: where to I put `USE_FSMV2_CPU=true` and `USE_FSMV2_TRANSPORT=true`?

Aaron99B: True. Adapting that

umh-core/docs/production/cpu-health.md
```diff
@@ -0,0 +1,102 @@
...
+and `USE_FSMV2_TRANSPORT=true`. Both are read at startup, so changing them requires a container
+restart. Without them, an instance is marked degraded whenever its CPU usage stays above 70% of
+the cores it may use, whether or not work is being delayed.
+{% endhint %}
+
+What UMH measures against depends on whether the container has a CPU limit (a Docker `--cpus` or a Kubernetes CPU limit).
+
+With no limit, the machine is the ceiling. UMH averages how busy the machine is over 60 seconds and reports degraded when less than about one core is free, the point where everything on the box, UMH included, starts waiting for CPU. The view also shows the container's own usage, so you can see how much of the machine total is UMH.
```
robertschwarz-umh: > the point where everything on the box, UMH included, starts waiting for CPU

This sort information is unrelated "btw" context and would have a better place in a separate section of the documentation. At this portion of the doc, we're establishing "what is a limit" and "how does our CPU health feature behave with / without limits".

umh-core/docs/production/cpu-health.md
```diff
@@ -0,0 +1,102 @@
+# CPU Health
+
+{% hint style="info" %}
+**Preview.** This reporting needs UMH Core v0.44.37 or later, started with `USE_FSMV2_CPU=true`
+and `USE_FSMV2_TRANSPORT=true`. Both are read at startup, so changing them requires a container
+restart. Without them, an instance is marked degraded whenever its CPU usage stays above 70% of
+the cores it may use, whether or not work is being delayed.
+{% endhint %}
+
+What UMH measures against depends on whether the container has a CPU limit (a Docker `--cpus` or a Kubernetes CPU limit).
```
robertschwarz-umh: Introductory sentences to sections describe the functionality in 1-2 paragraphs, let's do the same here.

Example:
- [saving data to historian intro](https://docs.umh.app/usage/historian/save-to-historian#:~:text=A%20Historian%20bridge,than%20a%20device.)
- [payload shapes intro](https://docs.umh.app/usage/data-modeling/payload-shapes#:~:text=Payload%20shapes%20define%20the%20structure%20and%20validation%20rules%20for%20message%20payloads.%20While%20built%2Din%20shapes%20handle%20most%20time%2Dseries%20data%2C%20custom%20shapes%20enable%20relational%20data%20modeling%20for%20complex%20business%20records.)

Starting a sentence with 'What' leads the reader to believe that this is going to be a question, but it's actually an explanation.

umh-core/docs/production/cpu-health.md
```diff
@@ -0,0 +1,102 @@
...
+
+With no limit, the machine is the ceiling. UMH averages how busy the machine is over 60 seconds and reports degraded when less than about one core is free, the point where everything on the box, UMH included, starts waiting for CPU. The view also shows the container's own usage, so you can see how much of the machine total is UMH.
+
+With a limit, the limit is the ceiling. UMH measures headroom against those cores rather than the machine, and reports degraded once its usage passes 90% of the limit, the last 10% being held in reserve, or when the kernel throttles it. A full machine can still slow a container that sits below its limit, so where UMH can read the host, it warns on that too.
```
robertschwarz-umh: > a full machine can still slow [...]

Similar to this one: https://github.com/united-manufacturing-hub/united-manufacturing-hub/pull/2756/changes#r3987741915

umh-core/docs/production/cpu-health.md
```diff
@@ -0,0 +1,102 @@
...
+| **CPU contention** | Tasks inside the instance spent time waiting for a free CPU core, for example 23% of the last minute. | Reduce the load, or give the instance more CPU. Workloads sharing the server may be competing for it. |
+| **CPU taken by the server** | Other virtual machines on the same physical server took CPU this instance needed. | On your virtualization platform, give this VM more guaranteed CPU, or move the other VMs off the server. |
+| **CPU running near full** | The host or the CPU limit is exhausted, so there is no room for the next burst of work. With host statistics readable, this fires when less than one core is free; without them, at a coarser proxy of sustained usage above 70% of the machine's cores. Technical Details names which case it is. | Add CPU capacity, reduce load, or raise the CPU limit. If the host is full and this container is not the cause, reserve or pin cores for UMH, or reduce what else runs on the machine: a CPU limit caps UMH rather than protecting it. |
```
robertschwarz-umh: > The host or the CPU limit is exhausted, so there is no room for the next burst of work. With host statistics readable, this fires when less than one core is free; without them, at a coarser proxy of sustained usage above 70% of the machine's cores. Technical Details names which case it is.

I've read through this multiple times and I don't really understand what this is trying to tell me. Also, "Technical Details" is established _after_ this table in the `text` example of `## Where to see it`. When I read through this the first time, I didn't know what Technical Details means, so it further confused me.

In contrast, the "What to do" portion of this row is very clear and easy to understand.

umh-core/docs/production/cpu-health.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Where to see it
+
+The Management Console shows CPU on the instance's detail page: the status, the usage row, and a
+Technical Details section listing every signal UMH can measure on that machine.
+
+Both the status and Technical Details carry the thresholds, so you do not have to look them up.
```
robertschwarz-umh: > so you do not have to look them up

This becomes a bit redundant, because as a reader, I wouldn't know or care about what to look up where in that situation.

umh-core/docs/production/cpu-health.md
```diff
@@ -0,0 +1,102 @@
...
+healthy verdict. Set a CPU limit on such a container to get a correct reading.
+
+## Glossary
+
+| Term | Meaning |
+|------|---------|
+| **Throttling** | The kernel caps a container to its CPU limit in short repeating periods, about 100 ms each. A container that needs more within a period is paused until the next one, so a workload whose average looks fine can still be paused during bursts. |
+| **CPU pressure** | How much time tasks spent waiting for a free CPU core, from Linux Pressure Stall Information (PSI). High pressure means CPU is the bottleneck. |
+| **CPU steal** | Time the hypervisor scheduled this machine's CPU elsewhere. High steal usually means the physical server is oversubscribed. A burstable cloud instance whose CPU credits have run out is capped by a different mechanism, but the guest counts that capped time as steal too. |
+| **Host contention** | CPU used by software outside UMH on the same machine. UMH cannot see which processes those are, only that they are using CPU it needs. |
```
robertschwarz-umh: What's the difference between CPU Contention in `## What each CPU status means` and Host Contention?

umh-core/docs/production/cpu-health.md
```diff
@@ -0,0 +1,102 @@
...
+| **CPU contention** | Tasks inside the instance spent time waiting for a free CPU core, for example 23% of the last minute. | Reduce the load, or give the instance more CPU. Workloads sharing the server may be competing for it. |
+| **CPU taken by the server** | Other virtual machines on the same physical server took CPU this instance needed. | On your virtualization platform, give this VM more guaranteed CPU, or move the other VMs off the server. |
+| **CPU running near full** | The host or the CPU limit is exhausted, so there is no room for the next burst of work. With host statistics readable, this fires when less than one core is free; without them, at a coarser proxy of sustained usage above 70% of the machine's cores. Technical Details names which case it is. | Add CPU capacity, reduce load, or raise the CPU limit. If the host is full and this container is not the cause, reserve or pin cores for UMH, or reduce what else runs on the machine: a CPU limit caps UMH rather than protecting it. |
+
+## Where to see it
```
robertschwarz-umh: When writing titles, assume that a user drops in from a google search, or is still in the google search index. The title will be displayed to the user there, just saying: "Where to see it".

Instead, call it: "Where to see the CPU health" / "How to see the CPU health" (or something along these lines)

robertschwarz-umh (commented):
I think there's 3 things that we need to tackle in a sweep:
- Avoid [center embedding](https://en.wikipedia.org/wiki/Center_embedding), because it makes technical documentation difficult to comprehend
- Lower the complexity of sentences with the [Dale-Chall readability formula.](https://en.wikipedia.org/wiki/Dale%E2%80%93Chall_readability_formula) 
- Flip the order of information to flow from 'known context' to 'something new here'. 

Examples:
**Explaining 'degraded' state**

Before the 3 rules:
> UMH averages how busy the machine is over 60 seconds and reports degraded  when less than about one core is free, the point where everything on the box, UMH  included, starts waiting for CPU.

After:
> Your Instance's health reports degraded when less than one core is free. Below that point, UMH and every other process on the machine have to wait for available CPU. The degraded state is defined by the average CPU load over 60 seconds.


**Explaining CPU Exhaustion**
Before:
> The host or the CPU limit is exhausted, so there is no room for the next burst of work. With host statistics readable, this fires when less than one core is > free; without them, at a coarser proxy of sustained usage above 70% of the machine's cores. Technical Details names which case it is.

After:
> The CPU is exhausted. The next burst of work has to wait. 
>
> The trigger depends on what UMH can read. When host statistics are readable, the alert fires when less than one core is free. When they are not, the alert fires when sustained usage exceeds 70% of the machine's cores. The `Technical Details section` names which case applies. See `[Where to see CPU health](#link-to-heading)` to learn about 'Technical Details'

umh-core/docs/production/cpu-health.md
Aaron99B: cut it

umh-core/docs/production/cpu-health.md
Aaron99B: rewrote it

umh-core/docs/production/cpu-health.md
Aaron99B: cut it

umh-core/docs/production/cpu-health.md
Aaron99B: rewrote it

umh-core/docs/production/cpu-health.md
Aaron99B: cut

umh-core/docs/production/cpu-health.md
Aaron99B: They are genuinely different, and the page never said so. Glossary entry now states it: host contention is other software on the contention?machine; the CPU contention status is measured inside the instance, as UMH's own tasks waiting for a core.

umh-core/docs/production/cpu-health.md
Aaron99B: changed to: Where to see the CPU health

robertschwarz-umh (approved):
Very nice! It's much easier to read and follow the text now, even with low previous context on the subject matter & feature. 

Thank you!

PR #2758, opened 2026-09-11
fix(cpuhealth): an absent cpu.stat stops degrading the instance

Aaron99B:
## Problem

RHEL 8 mounts cgroup v1 by default, so there is no `/sys/fs/cgroup/cpu.stat` on those hosts. The CPU worker treated that file as primary and failed the whole sample without it, so every poll failed, the instance reported `CPU degraded: poll error: read /sys/fs/cgroup/cpu.stat: no such file or directory`, and every bridge on it was blocked over a file the box was never going to have.

## What we are shipping

- An unreadable `cpu.stat` reads as absent no-signal, the way every other file the sampler cannot read already does. A `cpu.stat` that exists and does not parse still fails the sample, and so does a cancelled tick.
- On such a host the poll succeeds carrying no capacity, which the message layer already renders as CPU monitoring being unavailable. No false degrade, no blocked bridges.
- The startup snapshot reports the condition to Sentry once. The failed poll used to be how these hosts were visible to us, and it is gone.

## What we are NOT shipping

- No cgroup v1 numbers. This makes the CPU panel say "unavailable" on RHEL 8 rather than show figures; reading the v1 files is the PR stacked on this one.
- `cgroup_memory.go` reads `memory.max` and `memory.current` and makes the same cgroup v2 assumption, so it will misreport on the same hosts. Untouched here, wants its own ticket.

Fixes ENG-5906


umh-core/pkg/cpuhealth/read.go
```diff
@@ -104,12 +104,22 @@ func (s *linuxSampler) Read(ctx context.Context) (Sample, error) {
...
+	smp.record(OpCPUStat, statResult)
+	if statResult == ReadUnparsable {
+		// A cpu.stat that opens and does not parse is corrupt, and every number
+		// derived from it would be a guess. A cpu.stat that will not open is a
+		// different thing: the three readings below stay absent and the sample
+		// carries on, so a host keeping its CPU accounting elsewhere is not
+		// degraded over a file it was never going to have.
+		return smp, fmt.Errorf("parse %s/cpu.stat: %w", s.cgroup.base, statErr)
 	}
+	// A cancelled tick fails every read, which is the same shape as a host with
```
JeremyTheocharis, line 117: i dont understand why we are checking on ctx error?

umh-core/pkg/cpuhealth/read.go
```diff
@@ -104,12 +104,22 @@ func (s *linuxSampler) Read(ctx context.Context) (Sample, error) {
...
 	smp.CPUStatRaw = stat.Raw
-	smp.record(OpCPUStat, statOutcome(stat, statErr))
-	if statErr != nil {
-		// cpu.stat is primary: a read failure there fails the WHOLE sample,
-		// never a silent drop of the throttle counters as absent no-signal.
-		return smp, fmt.Errorf("read %s/cpu.stat: %w", s.cgroup.base, statErr)
+	statResult := statOutcome(stat, statErr)
+	smp.record(OpCPUStat, statResult)
+	if statResult == ReadUnparsable {
+		// A cpu.stat that opens and does not parse is corrupt, and every number
```
JeremyTheocharis, line 110: why? if it is unparsable, does that mean it is in our business logic here cgroupv1?

umh-core/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Instances on hosts whose CPU accounting sits outside the container's own cgroup are no longer reported as degraded, which previously blocked every bridge on them. CPU monitoring reads as unavailable on such a host instead
```
JeremyTheocharis, line 7: shouldn't it be that if the cgroup is absent, we should fallback to host metrics instead? so that the whole thing is not unavailable. we should say "we know we run in a docker container, but cgroupv1 is not supported yet. Please upgrade to cgroupv2. Using host metrics instead."<br><br>i remember that the the decision on whether to use docker or host signals is done during setup basedfo n the environment, so maybe there we need to add an additional cgroupv1 check there, that if it is cgroupv1 we switch to host metrics with that aditional cevat added to all messages?

JeremyTheocharis (changes requested):

umh-core/pkg/cpuhealth/read.go
Aaron99B: v1 does not land here, as for v1 the file does not even exist. We just reach this readUnparsable branch in a case where the cpu.stat file exists, but is not readable (which could happen in theory (for future cgroup versions, if someone someone mounted something else to that path or if another component than the kernel is actually serving the file

umh-core/pkg/cpuhealth/read.go
Aaron99B: Because a cancelled tick and a host with no cgroup files now produce the same sample. We include this check to differentiate the cases

umh-core/pkg/cpuhealth/read.go
```diff
@@ -104,12 +104,22 @@ func (s *linuxSampler) Read(ctx context.Context) (Sample, error) {
...
+		// derived from it would be a guess. A cpu.stat that will not open is a
+		// different thing: the three readings below stay absent and the sample
+		// carries on, so a host keeping its CPU accounting elsewhere is not
+		// degraded over a file it was never going to have.
+		return smp, fmt.Errorf("parse %s/cpu.stat: %w", s.cgroup.base, statErr)
 	}
+	// A cancelled tick fails every read, which is the same shape as a host with
+	// none of these files. Without this the sample reports the second.
+	if ctxErr := ctx.Err(); ctxErr != nil {
+		return smp, ctxErr
```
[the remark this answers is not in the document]
Aaron99B: closed the gap

umh-core/pkg/cpuhealth/cgroup_source.go
```diff
@@ -183,7 +183,7 @@ func parseCounter(data []byte, key string) (diagnosis.Reading, error) {
 		if len(fields) >= 2 && fields[0] == key {
 			v, err := strconv.ParseFloat(fields[1], 64)
 			if err != nil {
-				return diagnosis.Unknown(), fmt.Errorf("unparsable %s value %q: %w", key, fields[1], err)
+				return diagnosis.Unknown(), fmt.Errorf("%w: %s value %q: %w", errUnparsableRead, key, fields[1], err)
```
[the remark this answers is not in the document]
Aaron99B: Fixed

umh-core/CHANGELOG.md
Aaron99B: Actually this fallback is already included. If cpu.stat is gone but cpuset is readable we use the host cpu measure. I adapted the changelog accordingly.

umh-core/pkg/cpuhealth/read.go
JeremyTheocharis: → maybe put this into the code comment then 🙂 the relation to cgroupv1 and v2, etc.

umh-core/pkg/cpuhealth/read.go
JeremyTheocharis: i dont understand this sentence :(

Aaron99B: With a canceled tick I mean that the reading process is cancelled

umh-core/pkg/cpuhealth/read.go
Aaron99B: Included it

umh-core/pkg/cpuhealth/read.go
JeremyTheocharis: Can we put this into the comment? i find the current comment confusing.<br><br>e.g., "checking whether the reading process was canceled, and if yes, returbning the cancelation error" (therei s no need to mention tick)

umh-core/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Instances on hosts whose CPU accounting sits outside the container's own cgroup are no longer reported as degraded, which previously blocked every bridge on them. Such a host keeps the CPU signals it can still read, machine headroom and steal, and reports CPU monitoring as unavailable only where it can read none of them
```
JeremyTheocharis: try to get @m4jus to understand this. i think in its current form it is rather complex.

JeremyTheocharis (approved):
OK except the changelog

umh-core/pkg/cpuhealth/read.go
Aaron99B: Yes, added it

umh-core/CHANGELOG.md
Aaron99B: adapted it

PR #2761, opened 2026-09-11
feat(telemetry): declare every Sentry event umh-core can report

Aaron99B:
## Problem

A umh-core Sentry alert names an event and nothing else. `event_name` arrives as
`persistent_push_failure` or `stuck_action_detected`, and whoever is triaging has
to grep the repo to learn what it means, who owns it, and whether it is worth
waking someone. Nothing declared the set: 149 names existed as string literals at
call sites, eight of them prose rather than identifiers, and an event's severity
was decided per call by whether the author reached for `SentryWarn` or
`SentryError` — three names were reported at both levels from different places.

ManagementConsole solved this for itself in `shared/error_telemetry/`. This is the
same shape for umh-core.

## What we are shipping

- **`pkg/telemetry/telemetry.yaml` is where an event comes to exist.** 156 entries,
  each with the sentence a triager reads and the level it reports at. Editing that
  file and running `make generate-telemetry` is the whole workflow.
- **An unregistered event cannot be emitted.** `FSMLogger.Sentry` takes a generated
  identifier instead of a string, so a typo is a compile error, and the generator
  refuses ten classes of entry it could not turn into compiling Go.
- **The brief reaches Sentry.** A caller with no error of its own gets one
  synthesized from the brief, so the tag is the issue title and the sentence sits
  under it. A warn-level event used to arrive with no exception at all.
- **Names are hierarchical**, so `config watch: failed to read config` becomes
  `config::watch::read_failed`, and three names that carried several meanings are
  now several events rather than one with two severities.

## What we are NOT shipping

- **No shared module with ManagementConsole.** Its taxonomy is already vendored for
  `CRYPTOLIB=true` builds, but the default umh-core build has no MC sources at all,
  and umh-core is public where MC is private. Sharing would make core logging work
  only in the private-source build. Two YAML files, aligned by convention, as the
  three `UX_STANDARDS.md` files already are.
- **Nothing user-visible changes**, which is why this carries
  `skip-changelog-guard` rather than a changelog entry. No UI, no config, no
  product behaviour — only which names umh-core uses when reporting to our Sentry.
- **Every existing alert rule and saved search keyed on `event_name` needs
  updating.** All 149 names change, and each existing issue's history is orphaned
  under its old name. That is the cost of the rename and it is not avoidable
  halfway.
- **Three call sites change level**, because the names they shared carried
  different conditions. A restart on a stopped collector is now a warning distinct
  from a restart that failed after exhausting its attempts.
- **No batching and no TypeScript output.** umh-core already has two suppressors
  (the hook's fingerprint debouncer and per-instance report gates), and nothing
  outside umh-core reads these identifiers.

## Reading order

The diff is 2673 lines, and about 1150 of them are the taxonomy: 156 registry
entries and the file generated from them. That part scales with how many events
umh-core reports, so it does not get smaller. This order gets you through the
parts that carry a decision.

1. **`pkg/telemetry/generate/generator.go`** — the only intricate code here. It
   builds a tree from the `::` segments and emits it in sorted order, because the
   regeneration test compares bytes.
2. **`pkg/fsmv2/deps/logger_impl.go`** — one method, about 30 lines. The level
   comes from the identifier, a nil cause becomes the brief, and a zero-value
   identifier reports as a defect rather than a blank event.
3. **Ten entries of `pkg/telemetry/telemetry.yaml`**, whichever ten you like. If
   the briefs read well and the severities match what you would page on, the other
   146 follow the same shape.
4. **One converted call site**, then skim the rest. The message argument became
   the identifier and moved to the front, the error moved to fourth, and
   `deps.Err(err)` passed as a field became the cause. The compiler checked the
   other 164.

`umh-core/pkg/telemetry/identifiers.gen.go` is generated and marked as such, so
review collapses it. Nobody edits it; `make generate-telemetry` writes it, and a
test fails if it is stale.

The spec and the plan this work followed are deliberately not in the repo.

## Known non-failures

`pkg/service/redpanda` and `pkg/communicator/api/v2/http` fail on `staging` as
well, verified with this work stashed.


Aaron99B:
Superseded by the stack #2762 (registry and the new method) and #2763 (moving all 165 call sites). Same commits, split at the point where judgement stops and mechanical substitution starts, and with the gofmt churn and the process docs removed.

PR #2765, opened 2026-09-11
Report TimescaleDB metrics from the historian monitor

Aaron99B:
## Problem

umh-core knows whether the historian database answers, but nothing about its
state. Disk used, whether compression runs, whether a retention job has been
failing for a month — answering any of it means a psql prompt on the customer's
machine.

## What we are shipping

A `get-historian-metrics` action reads the database on request: versions, size
and growth, compression totals, per-table detail (size, rows, policies, first and
last write), and the background jobs. Tables the historian did not create are
counted and summed rather than listed, so the reported size adds up.

The connection monitor additionally reports two figures continuously: the table
names, and the count of failing jobs. Nothing else was added to the status
message — the rest is read only when the console asks.

## Notes for review

- First action in the package that performs external I/O, hence its own error
  code, a real timeout, and Sentry reporting.
- Rows come from the catalog, not a count: compressed chunks record their exact
  pre-compression count, uncompressed ones contribute the planner's estimate.
- Write times are RFC 3339 instants, so a console renders them without using its
  own clock.
- The summary read is bounded at 100ms and its failure discarded: a table listing
  must not drive the worker degraded for a database that is answering.
- `failedJobs` counts jobs that exist and are failing. A historian with no
  policies has none to fail, so zero is not proof that data is compressed or
  expired; the policy counts are read on request.

## Not included

No free-disk gauge (Postgres reports bytes used, never free). Freshness is
reported and never judged — no staleness threshold is ours to set. Failing jobs
are reported, not alarmed. No per-column compression algorithm: it is chosen per
batch and lives in the stored bytes. No Management Console rendering — that is
united-manufacturing-hub/ManagementConsole#3567.

Fixes ENG-5374

### DIFF
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### New Features
+
+- The instance now reports the state of its historian database alongside the connection check it already ran: the Postgres and TimescaleDB versions, how much disk the database occupies, how many hypertables and chunks it holds, how much the compressed chunks save, and how many of its background compression and retention jobs are failing. The figures refresh once a minute rather than every second, since none of them move faster than that, and a database that answers but refuses the reads reports the reason instead of going unhealthy
+
 ## [0.44.38]
 
 ### Improvements
--- a/umh-core/CHANGELOG.md
+++ b/umh-core/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### New Features
 
-- The instance now reports the state of its historian database alongside the connection check it already ran: the Postgres and TimescaleDB versions, how much disk the database occupies, how many hypertables and chunks it holds, how much the compressed chunks save, and how many of its background compression and retention jobs are failing. The figures refresh once a minute rather than every second, since none of them move faster than that, and a database that answers but refuses the reads reports the reason instead of going unhealthy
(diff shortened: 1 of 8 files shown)


umh-core/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### New Features
+
+- The instance can now report the state of its historian database when asked: how much disk it uses and how fast it is growing, what compression saves, the per-table detail behind those figures, and the background jobs that maintain it. The reads happen only when the Management Console asks, so an instance with a large historian sends no more than before. The table names and the count of failing background jobs are reported continuously, alongside the connection check
```
JeremyTheocharis, line 7: behind FF 🙂

also you can write something like "from tjis version on when the FF is enabled, the instance is able to provide the frontend with information on the state of its historian database"/

umh-core/pkg/fsmv2/historian/timescale.go
```diff
@@ -312,3 +329,52 @@ func init() {
...
 }
+
+// Neither figure moves faster than this: a table appears when a contract is
+// deployed, and a failing job is not urgent to the second.
+const summaryInterval = 60 * time.Second
+
+// Sized against the poll cycle, not the observation deadline: a read that blocks
+// delays the connection check this worker exists to produce. Exceeding it costs
+// nothing beyond a stale summary, since the previous value is kept.
+const summaryBudget = 100 * time.Millisecond
```
JeremyTheocharis, line 340: do we need this bduget here? do we need the budget at all? tight budgets like this made the original fsmv1 implementation unstable under load.theoretically the Poll of the historian worker has i think 20 seconds or so time before it is marked as stale.

PR #2770, opened 2026-09-14
feat: sentry reporting for ads plugin errors

FalconTube:
## Problem

Need Sentry error reporting in umh-core for rollout of Beckhoff ads plugin from Benthos.
Check here for an example sentry log: https://united-manufacturing-hub.sentry.io/issues/146941889/

## What we are shipping

- Generic sentry reporter for ALL benthos plugins -> But only `ads` enabled for now -> Can think about, if we want more tracing in future
- Fires via goroutine, so that it does not block anything, because it runs in the reconcile loop

## What we are NOT shipping

- NOT enabling this for all protocols, only `ads` here


umh-core/pkg/fsm/benthos/sentry_plugin_errors.go
```diff
@@ -0,0 +1,169 @@
...
+	"go.uber.org/zap"
+
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/sentry"
+	s6service "github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/service/s6"
+)
+
+// pluginErrorReportingEnabled lists the benthos-umh plugin IDs (as returned
+// by dataflowcomponentserviceconfig.BenthosPluginID) reported to Sentry.
+var pluginErrorReportingEnabled = map[string]bool{
+	"ads": true,
```
Aaron99B: What does this ads mean?

umh-core/pkg/fsm/benthos/sentry_plugin_errors.go
```diff
@@ -0,0 +1,169 @@
...
+
+	"go.uber.org/zap"
+
+	"github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/sentry"
+	s6service "github.com/united-manufacturing-hub/united-manufacturing-hub/umh-core/pkg/service/s6"
+)
+
+// pluginErrorReportingEnabled lists the benthos-umh plugin IDs (as returned
+// by dataflowcomponentserviceconfig.BenthosPluginID) reported to Sentry.
+var pluginErrorReportingEnabled = map[string]bool{
```
Aaron99B, line 31: Shouldn't we check this then in the shouldReportPluginError function?

Aaron99B (changes requested):

umh-core/pkg/fsm/benthos/sentry_plugin_errors.go
FalconTube: The check in `shouldReportPluginError` is not strictly needed, because it is caught in `reportPluginErrorAsync` (which actually sends the request to Sentry).
But I added the additional check now in the function here as well, can't hurt :)

Aaron99B (approved):

umh-core/pkg/fsm/benthos/sentry_plugin_errors.go
```diff
@@ -0,0 +1,169 @@
...
+	"ads": true,
+}
+
+const pluginErrorDebounceWindow = 1 * time.Hour
```
JeremyTheocharis, line 35: dont we have for fsmv2 already debouncing? maybe we can import the fsmv2 logger here and re-use it? it would start centralizing

FalconTube: Ah, I wasn't aware of the sentry debouncer, thought we only had the logs debouncer which I added.
But I added it now, good catch!

umh-core/pkg/fsm/benthos/sentry_plugin_errors.go
```diff
@@ -0,0 +1,169 @@
...
+func reportPluginErrorAsync(protocol, instanceID string, logs []s6service.LogEntry, currentTime time.Time, logWindow time.Duration, triggerMessage string, logger *zap.SugaredLogger) {
+	if !pluginErrorReportingEnabled[protocol] {
+		return
+	}
+
+	event := pluginErrorEvent{
+		protocol:       protocol,
+		instanceID:     instanceID,
+		triggerMessage: triggerMessage,
+		recentLogs:     recentLogsWithinWindow(logs, currentTime, logWindow, pluginErrorLogWindowMaxLines),
```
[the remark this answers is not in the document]
FalconTube: solved in a different way, by just gating behind the debounce window

PR #2772, opened 2026-09-15
Eng 5775 reject over long data model names in umh core not only at

wilsonedgar-umh:
## Problem
When a user created a data model with an extremely long name the error would only be visible downstream in benthos

## What we are shipping
Validation at data model creation level. If the name is too long the user will not be allowed to create it and will receive an appropriate error message

## What we are NOT shipping
The character limit is not synced with the benthos repo.

Fixes ENG-5775


umh-core/pkg/communicator/actions/add-datamodel.go
```diff
@@ -124,6 +130,10 @@ func (a *AddDataModelAction) Validate() error {
 		return errors.New("missing required field Name")
 	}
 
+	if len(a.payload.Name) > maxContractLen {
+		return fmt.Errorf("data model name %q is %d characters; use %d or fewer, because PostgreSQL truncates the table name at 63 bytes", a.payload.Name, len(a.payload.Name), maxContractLen)
```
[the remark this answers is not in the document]
wilsonedgar-umh: While then might be UTF-8 the frontend only accepts ASCII so each character will take one byte

FalconTube (changes requested):
This is not a proper review, just a "thinking hint": What happens when the user does not interact with the `add-datamodel` action, but rather edits the `config.yaml` directly?<br><br>Unfortunately, this is something we always have to consider.
Makes implementations more comples.<br><br>You can try to solve this on your own.
Otherwise, just ping me again for support on this 🙂

umh-core/pkg/communicator/actions/add-datamodel.go
```diff
@@ -48,6 +48,12 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+// maxContractLen is the maximum length of a data model name. The data contract
```
JeremyTheocharis, line 51: maybe add there the source of this? a link to this weird number? and maybe also linking to the database table definitions? so that a person can understand where this is coming from (and also when does it need to be changed)

.gitignore
```diff
@@ -22,3 +22,4 @@ finalization/
 .worktrees/
 .self-review/
 vsdd/
+/umh-core/compose.yaml
```
JeremyTheocharis, line 25: is this indetended?

umh-core/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
```
JeremyTheocharis, line 5: does this change also require a change  in the management console?

umh-core/pkg/communicator/actions/add-datamodel.go
JeremyTheocharis: maybe could even be exposed from the historian package and then re-used here. could be morem aintainable there?

umh-core/CHANGELOG.md
JeremyTheocharis: so to have a UI check there as well?

.gitignore
wilsonedgar-umh: Yes this my local compose file

umh-core/CHANGELOG.md
wilsonedgar-umh: No there are no changes in the management console. The error message comes from umh-core

JeremyTheocharis: shouldnt we then add a check there as well? so if the user wants to add it there, he immediately gets a validation wanring even before he presses save?

wilsonedgar-umh: Validations, should preferably live in one place, as to avoid avoid duplication of work and the scattering of business logic. With that said I would leave all here.  And unlike some validations that happens in the frontend (check for required field,  max length, etc…) this validation is regarding a database rule so it should really live in the backend

JeremyTheocharis: generally agree and this is where we want to move towards to. however, we already do it like this in a lot of places to prevent user frustrations. in the future we can move towards it (e.g., autogenerate frontend schemas based on validation rules in the backend). feel free to check with Roger for example how we do it across the code base.

umh-core/pkg/communicator/actions/add-datamodel.go
wilsonedgar-umh: Added the same comment as we currently have in benthos point to Postgres limit 
https://www.postgresql.org/docs/current/limits.html

umh-core/CHANGELOG.md
wilsonedgar-umh: Added validation to the Management console [PR-3350](https://github.com/united-manufacturing-hub/ManagementConsole/pull/3550)

wilsonedgar-umh:
For the time being we have validation for name maxlength at these locations:

* frontend (management console)
* benthos-umh
* umh-core
* database (postgres - the point of constraint)

File validation will be left for a future when we create a more integrated validation framework

JeremyTheocharis:
@Linear add this to the config /cli rollout ticket that hsoudl be in state TODO as background.

JeremyTheocharis (approved):

FalconTube:
@linear here we made the decision, to only gate the data model length on the `add` action. We do not validate this in the `config.yaml`, so if a long datacontract name gets there via direct modification of the file, then we are not safe against it.
Add this as a decision to the original ticket

FalconTube (approved):

PR #2777, opened 2026-09-17
docs: how to enable and verify CPU pressure stats (PSI)

JeremyTheocharis:
Closes ENG-5949.

## Problem

UMH tells a user to "boot with `psi=1`" when it cannot read CPU pressure, without saying which operating systems need it, how to set a kernel parameter, or how to check it worked; and on the Red Hat family's version 8, where this is most common, `psi=1` alone changes nothing, because the container's `cpu.pressure` also needs cgroup v2.

## What We're Shipping

A section on the CPU Health page that says which of the two conditions is missing before the reboot rather than after it, links each operating system's own instructions for setting a kernel parameter instead of carrying commands we would have to maintain, and checks the fix by reading the file UMH reads, plus one line under System requirements and the page's URL in the status message that asks for the fix.

## What We're NOT Shipping

No table of which distribution ships what, because that is a fact about someone else's kernel build that changes without telling us; no change to the other two messages mentioning `psi=1`, which are asides; no clickable link, because the Console renders the alert as plain text.

## Corrections this turned up

The ticket body and my first draft were both wrong in places. Each of these is checked against the implementation or a vendor kernel config:

- Debian needs nothing. `CONFIG_PSI=y` with `CONFIG_PSI_DEFAULT_DISABLED` unset in Debian's kernel config on bullseye, bookworm and trixie, and cgroup v2 since Debian 11. The ticket lists Debian as needing `psi=1`.
- The ticket's check reads the wrong thing. `grubby --info | grep args` shows the kernel took the flag, not that the container can read the file. `cgroup_source.go` `readPSI` opens `/sys/fs/cgroup/cpu.pressure`.
- `update-grub` is a Debian and Ubuntu wrapper, absent on Fedora and on other GRUB systems, which use `grub-mkconfig`.
- Switching a host to cgroup v2 needs Docker 20.10, or the machine reboots into a state where UMH Core does not start.
- A Compose install names the container `<project>-umh-1`, not `umh-core`.
- `Details.LimitedVisibility` is set only where the quota is absent, so a container with a CPU limit never reports limited visibility however absent PSI is.

## Tests

`message_test.go` asserted the old status message text exactly, so the string change went red before it went green.

## Not checked on a machine

Nothing here ran on a RHEL-family 8 host. The claim that `psi=1` alone leaves `cpu.pressure` absent under cgroup v1 rests on the kernel's documented behaviour and on our read path, not on a reproduction.


umh-core/docs/production/cpu-health.md
```diff
@@ -68,6 +68,68 @@ Steal uses the 95th percentile once 20 samples are in, and the mean before that,
 instance is judgeable within seconds of starting. Bare metal reports no steal at all, so on a
 physical machine that signal reads "not possible" rather than 0%.
 
+## Enabling CPU pressure stats
+
+UMH reads CPU pressure from the container's own `/sys/fs/cgroup/cpu.pressure`. That file exists only
+if the kernel booted with Pressure Stall Information (PSI) on and the machine runs cgroup v2, the
+Linux facility that meters a container's CPU, memory and disk use. Most distributions ship both on.
```
JeremyTheocharis:
```
the
Linux facility that meters a container's CPU, memory and disk use. M
```

cut this

umh-core/docs/production/cpu-health.md
```diff
@@ -68,6 +68,68 @@ Steal uses the 95th percentile once 20 samples are in, and the mean before that,
...
+The Red Hat family is the exception: RHEL, Rocky, AlmaLinux and Oracle Linux ship PSI off, and
+version 8 of each also boots cgroup v1.
+
+Check your own machine:
+
+```bash
+docker exec umh-core cat /sys/fs/cgroup/cpu.pressure
+```
+
+A Docker Compose install names the container `<project>-umh-1` rather than `umh-core`. `docker ps`
```
JeremyTheocharis: cut this line

umh-core/docs/production/cpu-health.md
```diff
@@ -68,6 +68,68 @@ Steal uses the 95th percentile once 20 samples are in, and the mean before that,
...
+
+A line beginning `some avg10=` means there is nothing to do. `No such file or directory` means one or
+both are missing. Find out which, on the host:
+
+```bash
+stat -fc %T /sys/fs/cgroup
+```
+
+`cgroup2fs` means only PSI is missing: set `psi=1`. `tmpfs` means the machine runs cgroup v1 and both
+are missing: set `psi=1` and, on a systemd distribution, `systemd.unified_cgroup_hierarchy=1`.
```
JeremyTheocharis: cut

```
and, on a systemd distribution, `systemd.unified_cgroup_hierarchy=1`.
Setting both at once costs one reboot instead of two. Version 8 of RHEL, Rocky and AlmaLinux boots
cgroup v1; version 9 and later boots cgroup v2.
```

umh-core/docs/production/cpu-health.md
```diff
@@ -68,6 +68,68 @@ Steal uses the 95th percentile once 20 samples are in, and the mean before that,
...
+
+### Set the kernel parameters
+
+Each operating system sets kernel parameters differently. Follow yours, then reboot:
+
+| Operating system | Instructions |
+|------------------|--------------|
+| RHEL, Rocky, AlmaLinux, Oracle Linux, Fedora | [Configuring kernel command-line parameters](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/managing_monitoring_and_updating_the_kernel/configuring-kernel-command-line-parameters_managing-monitoring-and-updating-the-kernel): `sudo grubby --update-kernel=ALL --args="psi=1"` |
+| Debian and Ubuntu | [GRUB 2 setup](https://help.ubuntu.com/community/Grub2/Setup): add `psi=1` to `GRUB_CMDLINE_LINUX` in `/etc/default/grub`, then run `sudo update-grub` |
+| Void Linux | [Void Handbook: kernel](https://docs.voidlinux.org/config/kernel.html) |
```
JeremyTheocharis: why void linux? remove it

umh-core/docs/production/cpu-health.md
JeremyTheocharis: Cut in 2096523df. The sentence now stops at "the machine runs cgroup v2".

umh-core/docs/production/cpu-health.md
JeremyTheocharis: Cut in 2096523df.

umh-core/docs/production/cpu-health.md
JeremyTheocharis: Cut in 2096523df. The branch now ends at "`tmpfs` means the machine runs cgroup v1, where `cpu.pressure` does not exist at all", so the page no longer tells anyone to switch a host to cgroup v2. Two knock-ons: a RHEL 8 reader's only way out is the CPU-limit alternative at the end of the section, and the Docker 20.10 warning lost the switch it was warning about, so I cut that too in 37bd5818b.

umh-core/docs/production/cpu-health.md
JeremyTheocharis: Removed in 2096523df. It came from the ticket's distribution list; nothing in our docs or support history points at Void, so it was carrying a row for a user we do not have.

umh-core/docs/production/cpu-health.md
```diff
@@ -68,6 +68,68 @@ Steal uses the 95th percentile once 20 samples are in, and the mean before that,
...
+loses CPU pressure, throttling and the container's own CPU limit.
+
+### Set the kernel parameters
+
+Each operating system sets kernel parameters differently. Follow yours, then reboot:
+
+| Operating system | Instructions |
+|------------------|--------------|
+| RHEL, Rocky, AlmaLinux, Oracle Linux, Fedora | [Configuring kernel command-line parameters](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/managing_monitoring_and_updating_the_kernel/configuring-kernel-command-line-parameters_managing-monitoring-and-updating-the-kernel): `sudo grubby --update-kernel=ALL --args="psi=1"` |
+| Debian and Ubuntu | [GRUB 2 setup](https://help.ubuntu.com/community/Grub2/Setup): add `psi=1` to `GRUB_CMDLINE_LINUX` in `/etc/default/grub`, then run `sudo update-grub` |
```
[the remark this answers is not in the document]
JeremyTheocharis: Not applying this. The split you describe needs the cgroup v2 parameter in the table, and that instruction was removed from the page in 2096523df: nothing here tells a reader to move a host from cgroup v1 to cgroup v2 any more. RHEL 8 now reads as a dead end for CPU pressure, with the CPU-limit alternative at the end of the section as the way out. The grouped row is therefore correct as it stands, since psi=1 is the only parameter the page asks for.

Aaron99B (approved):

PR #2779, opened 2026-09-21
Removed stale documentation regarding usage of refModel

wilsonedgar-umh:
## Problem
Removed section of documentation that are no longer valid. We no longer support ref to other data models

## What we are shipping

Updated data models documentation

## What we are NOT shipping

Fixes ENG-6186


rpaland (approved):
LGTM.

I just saw we also have this in some tests and references. Should be on the map to remove entirely from the codebase.

